PR #1, opened 2016-09-21
typo

DarkSide666:
You're always mixing Symantic with Semantic. Correct is S**e**mantic UI.


romaninsh:
:) ok.

romaninsh:
at least it's not Symantec :)

PR #2, opened 2016-09-23
Keep requirement to develop for now

gowrav-vishwakarma:


romaninsh:
fine until 1.1.0 is released. Will override merge error.

PR #16, opened 2016-11-27
Feature/implement basic views

romaninsh:
Initial implementation


src/Tree.php
```diff
@@ -0,0 +1,59 @@
...
+            $item['Content'] = $sub_output;
+
+            $output .= $item->render();
+        }
+        return $output;
+    }
+
+    function render()
+    {
+        $this->output = $this->renderSubTree($this->model);
```
DarkSide666: I guess it should be `return $this->renderSubTree($this->model);` here

romaninsh: tree needs to be cleaned up, i'll do that now.

PR #20, opened 2016-12-27
Feature/js integration

romaninsh:
Following #18 , this is a second step of the integration. Chains now are integrated into the View class and js() and on() methods are implemented.

Test suite has been added and I've also added our first Selenium test!

We're actually running IE7 to test our JavaScript engine.

Travis tests are currently ignored because of http://stackoverflow.com/questions/41337139/allow-phpunit-test-to-access-actual-environment-variable?noredirect=1#comment69877502_41337139


docs/js.rst
```diff
@@ -30,11 +30,19 @@ with:
 
 .. code-block:: js
 
+<<<<<<< HEAD
```
DarkSide666: something is wrong here with merging my docs fixes from develop branch. <<HEAD and >>develop left here.

src/View.php
```diff
@@ -349,19 +377,249 @@ public function render()
...
+     * $view->js();                   // technically does nothing
+     * $a = $view->js()->hide();      // creates chain for hiding $view but does not
+     *                                // bind to event yet.
+     *
+     * 2. Binding existing chains
+     * $img->on('mouseenter', $a);    // binds previously defined chain to event on
+     *                                // event of $img.
+     *
+     * Produced code: $('#img_id').on('mouseenter', function(ev){ ev.preventDefault();
+     *    $('view1').hide(); });
```
DarkSide666: proper selector naming $('#view_id').hide()

src/View.php
```diff
@@ -349,19 +377,249 @@ public function render()
...
+     *
+     * Documentation:
+     *
+     * @link http://agile-ui.readthedocs.io/en/latest/js.html
+     *
+     * @param string|bool|null          $when     Event when chain will be executed
+     * @param array|jQuery_Chain|string $code     JavaScript chain(s) or code
+     * @param string                    $instance Obsolete
+     *
+     * @return jQuery_Chain
```
DarkSide666: should be `@return jQuery` (not Chain)

src/View.php
```diff
@@ -349,19 +377,249 @@ public function render()
...
+     *   if (!$data['clickable']) {
+     *      return new jsExpression('alert([])', ['This record is not clickable'])
+     *   }
+     *   return $js->parent()->hide();
+     * });
+     *
+     * For more information on how this works, see documentation:
+     *
+     * @link http://agile-ui.readthedocs.io/en/latest/js.html
+     */
```
DarkSide666: no @param, @return descriptions

src/jsFunction.php
```diff
@@ -11,41 +11,43 @@ class jsFunction implements jsExpressionable
...
-    public $fx_statements;
+    public $fx_statements = [];
+
+    /** @var bool add preventDefault(event) */
+    public $preventDefault = false;
+
+    /** @var bool add stopPropagation(event) */
+    public $stopPropagation = false;
+
+    /** @var string Array of statements */
```
DarkSide666: wrong comment

romaninsh: not sure why wrong.

src/View.php
```diff
@@ -64,6 +64,8 @@ class View implements jsExpressionable
      * Path to template. If you don't specify full path
      * by starting with '/' then will be prepended by
      * default template path.
+     *
+     * @var Template
```
DarkSide666: It's not Template, it's a string!

src/View.php
```diff
@@ -197,20 +205,26 @@ public function init()
...
-        $this->app = new \atk4\ui\App(['skin'=>$this->skin]);
+        $this->app = new App(['skin'=>$this->skin]);
         $this->app->init();
     }
 
     /**
      * In addition to adding a child object, set up it's template
      * and associate it's output with the region in our template.
+     *
+     * @param $object View New object to add
```
DarkSide666: should be `@param View $objec tNew object to add`

src/View.php
```diff
@@ -294,8 +306,6 @@ public function addClass($class)
      * Remove one or several CSS classes from the element.
      *
      * @param string|array $class CSS class name or array of class names
```
DarkSide666: only string is treated correctly (see explode). Or it should be like this in code:
```
if (is_string($remove_class)) {
    $remove_class = explode(' ', $remove_class);
}
```

src/View.php
```diff
@@ -440,10 +450,10 @@ public function getHTML()
      * @link http://agile-ui.readthedocs.io/en/latest/js.html
      *
      * @param string|bool|null          $when     Event when chain will be executed
-     * @param array|jQuery_Chain|string $code     JavaScript chain(s) or code
+     * @param array|jQuery|string $code     JavaScript chain(s) or code
```
DarkSide666: $when can not be array as far as I see in code. Only string|boolean|jQuery

src/View.php
```diff
@@ -516,7 +526,7 @@ public function on($event, $selector = null, $action = null)
             // if callable $action is passed, then execute ajaxec()
             //
             throw new Exception('VirtualPage is not yet implemented');
-            $url = '.virtualpage->getURL..';
+            /*$url = '.virtualpage->getURL..';
```
DarkSide666: no respective closing */

src/View.php
romaninsh: it's actually string first but is replaced with a template. In a way, it combines function of 'defaultTemplate'. If I specify 'string|Template' then code hinting is not working. Something need to be renamed here.

src/View.php
romaninsh: i was planning to make array possible, but perhaps I shouldn't bother.

src/View.php
romaninsh: yah, it's just blocking unfinished code, so it's intended. That area is not reachable anyway.

src/View.php
romaninsh: fixed in f0ae37d

src/View.php
romaninsh: Dark, where did i say that "when" could be an array? Give me link an i'll fix it there.

src/View.php
romaninsh: Fixed in 096a361

src/View.php
romaninsh: Fixed in 096a361

PR #28, opened 2016-12-28
implement server-side on() events

romaninsh:
This pull requests aims to make this code work:

``` php
$b = new Button();
$b->on('click', function($js){ $js->text(rand(1,20)); });
```

Clicking a button here would execute request to the server that would generate the new label for a button.


romaninsh:
I am OK with the remaining 4 code-climate issues. I don't want to disable them because we will need to address them at some point.

docs/js.rst
```diff
@@ -72,75 +193,149 @@ will output:
...
-    Some of the clases that implement jsExpressionable are:
-     
-    - jsExpression
-    - jsChain
-    - View
+    $b1 = $buttons->add(new Button('One'));
+    $b2 = $buttons->add(new Button('Two'));
+    $b3 = $buttons->add(new Button('Three'));
 
+    $buttons->on('click', '.button', $b3->js()->hide());
```
DarkSide666: Wrong example!

src/View.php
```diff
@@ -478,7 +495,7 @@ public function getHTML()
      *
```
DarkSide666: no description what $extra parameter is for

src/View.php
```diff
@@ -555,8 +576,51 @@ public function on($event, $selector = null, $action = null)
...
+                    } else {
+                        throw new Exception(['Incorrect callback. Must be string or action.', 'r'=>$r]);
+                    }
+                }
+
+                $ajaxec = implode(";\n", array_map(function (jQuery $r) {
+                    return $r->jsRender();
+                }, $actions));
+
+                echo json_encode(['success'=>true, 'message'=>'Hello World', 'eval'=>$ajaxec]);
```
DarkSide666: Hello world???

src/View.php
```diff
@@ -555,8 +576,51 @@ public function on($event, $selector = null, $action = null)
...
+                exit;
+            });
+
+            $thisAction->api(['on'=>'now', 'url'=>$cb->getURL(), 'obj'=>new jsExpression('this')]);
```
DarkSide666: $thisAction->api ? what is that?

src/jsExpression.php
```diff
@@ -73,4 +73,94 @@ function ($matches) use (&$namelessCount) {
 
         return trim($res);
     }
+
+    /**
+     * Provides replacement for json_encode that will respect jsExpressionable objects
+     * and call jsRender() for them instead of escaping.
+     *
+     * @param mixed @arg anything
```
DarkSide666: should be $arg instead of @arg

src/jsAPI.php
```diff
@@ -0,0 +1,19 @@
+<?php
```
DarkSide666: what is this class for? where is it used?

romaninsh: removing it, it's confusing.

src/jsExpression.php
romaninsh: fixed

src/View.php
romaninsh: that's Semantic UI ;) 

http://semantic-ui.com/behaviors/api.html

src/View.php
romaninsh: This is changed in the next PR, so i'll leave it to avoid conflicts.

src/View.php
romaninsh: it's changed in the next PR, so leaving.

docs/js.rst
romaninsh: There is major overhaul on js.rst in the next PR.

PR #29, opened 2017-01-03
Feature/form

romaninsh:
(includes #28), on-going work for implementing Form. Because of how extensive the functionality of Form is, there might be multiple phases for Form implementation, i'll try to keep it structured though.

 - [x] Implement R/O standalone fields
 - [x] Implement field labels 
 - [x] Association with Form 

Further work is done in feature/form2 branch:

 - [ ] Integrate form submit action
 - [ ] Fetch value from model and display
 - [ ] Take value from submitted data and put back into model


romaninsh:
Added form implementation:

screen shot 2017-01-04 at 06 36 46

(see demos/input.php)

romaninsh:
Added more field types: 
screen shot 2017-01-04 at 15 58 32

src/Button.php
```diff
@@ -25,6 +27,18 @@ public function renderView()
             $this->addClass('icon');
         }
 
+        if ($this->rightIcon && !is_object($this->rightIcon)) {
+            $this->rightIcon = $this->add(new Icon($this->rightIcon), 'Content');
+
+            if ($this->content) {
```
DarkSide666: what is $this->icon or $this->rightIcon is already an object? Shouldn't we execute this part of code in such case too? (add class labeled, remove content etc.)

romaninsh: that's currently incosistent in the project but i've addressed that in the next PR, but i'm not fully sure.

DarkSide666: OK, then merging this in.

PR #33, opened 2017-01-24
Feature/layouts

romaninsh:
We need a couple of usable full-screen layouts coming as a standard.

This pull request adds:

 - Centered - layout that puts content in the center of the page.
 - Admin - layout with main menu/dropdowns on top, user icon on the right and wide content

Also demos are not neatly organised. Simply clone and open "demos" folder in the browser.

screen shot 2017-02-07 at 10 49 21


src/View.php
```diff
@@ -134,11 +134,11 @@ public function __construct($defaults = [])
...
      * @return Model
      */
-    public function setModel(\atk4\data\Model $model)
+    public function setModel(\atk4\data\Model $m)
     {
-        $this->model = $model;
+        $this->model = $m;
 
-        return $model;
+        return $m;
```
romaninsh: why not leave $model ?

DarkSide666: ok we can leave $model, but then in method comments it should be $model too :)

romaninsh:
Dark, I have added some basic documentation on the "Layouts". It's not deep yet, but I'm not sure in which directions Layouts will go. They are a new concept for ATK so I'd like to give it some consideration.

Do verify the rest and let's switch to Grids next.

src/Button.php
```diff
@@ -43,6 +43,26 @@ public function renderView()
...
 
     /**
+     * Makes button into a "<a>" element with a link.
+     */
+    public function link($url)
+    {
+        $this->element = 'a';
+        if (is_string($url)) {
+            $sch = substr($url, 0, 3);
+            if ($sch === 'htt' || $sch = '://') {
```
romaninsh: I'm actually going to do like this: all strings are inserted as-is. Array routed through url()

src/Button.php
```diff
@@ -43,6 +43,26 @@ public function renderView()
...
+                $this->setAttr('target', '_blank');
```
romaninsh: should i drop target setting or add as argument to link() ?

DarkSide666: i think - drop that line :)

src/Button.php
DarkSide666: i agree

PR #63, opened 2017-03-08
Feature/minor fixes

romaninsh:
Adding missing button documentation, cleaning up Button logic, etc.

Also changing button's "rightIcon" property into "iconRight" (to help with auto-complete, start typing icon, see options).

Remove hack that would put "Button" inside a newly added button even if label is not specified. Now will appear as empty button, but much less of a hack for us to implement.


docs/button.rst
```diff
@@ -0,0 +1,77 @@
...
+
+    $button = $view->add(new \atk4\ui\Button('Click me'));
+
+Button will typically inherit all same properties of a :php:class:`View`. Functionality
+of View alone yields in many various usage patterns such as::
+
+    $b1 = new Button(['Load', 'primary']);
+
+    $button = new Button('Hello there');
+    $button->addClass(['size big'=>true]);
```
DarkSide666: do we really need to write like this?
will `$button->addClass('size big')` not work?

PR #64, opened 2017-03-11
Feature/grid table rename and advanced grid

romaninsh:
I have decided to split this change up into multiple smaller chunks. 

The first one renames "Grid" into "Table" and also "Column" namespace into "TableColumn". This is needed because I'm going to add class "Column" which would otherwise clash with table columns. This is not backwards-compatible change, as we are still early on in the versions and creating backwards compatibility would be too messy.


src/TableColumn/Actions.php
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Formatting money.
+ */
+class Actions extends Generic
+{
+    public $actions = [];
+
+    public function addAction($icon, $callback)
+    {
+        $cb = $this->add(new jsCallback(), $action);
```
DarkSide666: There is no variable $action and not used parameter $icon. I guess parameter should be named - $action.

src/TableColumn/Link.php
```diff
@@ -1,14 +1,25 @@
...
 class Link extends Generic
 {
+    /**
+     * TODO: Refactor so that this goes away!!
+     */
     public $page = null;
 
+    /**
+     * Use
+     *   new Link('http://yahoo.com?id={$id}');
```
DarkSide666: This syntax is not supported as I see below in constructor

src/TableColumn/Link.php
```diff
@@ -36,6 +47,14 @@ public function getCellTemplate(\atk4\data\Field $f)
         return $output;
     }
 
+    /**
+     * Auto-fill tag for the row-id.
+     *
+     * @param [type] $row   [description]
```
DarkSide666: need to set correct parameter types

src/TableColumn/Actions.php
romaninsh: oh actually Action is incomplete., i should have dropped it.

romaninsh: addressed (removing Actions for now)

PR #66, opened 2017-03-12
Feature/js reloading

romaninsh:
New action that reloads a specific view:

``` php
$button->js('click', new \atk4\ui\jsReload($table));
```


docs/field.rst
```diff
@@ -122,3 +122,15 @@ Integration with Form
 ---------------------
 
 This section explains how Field interracts with the form.
+
+JavaScript on Input
+-------------------
+
+.. php:meth:: jsInput()
```
DarkSide666: No description of jsInput() arguments
`jsInput($when = null, $action = null)`

src/FormField/Input.php
```diff
@@ -41,6 +41,17 @@ class Input extends Generic
     public $actionLeft = null;
 
     /**
+     * Method similar to View::js() however will adjust selector
+     * to target the "input" element.
+     *
+     * $field->jsInput(true)->val(123);
+     */
+    public function jsInput($when = null, $action = null)
```
DarkSide666: Why not add this in FormField/Generic?
Currently it will not work for FormField/Checkbox for example.

src/jsCallback.php
```diff
@@ -53,9 +53,7 @@ public function set($callback, $args = [])
                 return $r->jsRender();
             }, $actions));
 
-            echo json_encode(['success'=>true, 'message'=>'Hello World', 'eval'=>$ajaxec]);
-            $this->app->run_called = true; // prevent shutdown function from triggering.
-            exit;
+            $this->app->terminate(json_encode(['success'=>true, 'message'=>'Hello World', 'eval'=>$ajaxec]));
```
DarkSide666: I still don't get why we need that `message=>'Hello World'` here

docs/field.rst
romaninsh: added arguments.

src/FormField/Input.php
romaninsh: because it's jsInput, so only relevant for "Input" type fields. Fields that are not based on "input" will have something else.

src/jsCallback.php
romaninsh: addressed.

PR #73, opened 2017-03-17
Feature/js library

ob6160:
I've started thinking about the best way to construct a library which provides functionality that fits in with the patterns being used by atk4/ui. This is an initial implementation which explores some some ideas regarding both build tools and file structure/layout. Currently only implements two plugins:

/js/src/plugins/reloadView
/js/src/plugins/spinner


js/LICENSE
```diff
@@ -0,0 +1,22 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 Krasimir Tsonev
```
romaninsh: i don't think it's needed.

ob6160: whoops - left over from the boilerplate code :)

js/src/plugin.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * class MyPlugin {
+ *     constructor(element, options) {
+ *         // ...
+ *     }
+ * }
+ *
+ * MyPlugin.DEFAULTS = {};
+ *
+ * plugin('myPlugin', MyPlugin');
```
romaninsh: a typo?

PR #74, opened 2017-03-17
Add "File Finder" component demo

romaninsh:
screen shot 2017-03-17 at 17 07 53

This PR adds a sample-component implementing functionality of MacOSX Finder. 

The data is stored in a recursive model. Also including a data importer for populating database.

IMPORTANT: Merge #66, #71 first. May conflict with #73.

Issue #92, opened 2017-03-29
Multiple formatters for single column in Table


romaninsh:
As per #67 We need to support multiple formatters for column and columns without field in Table component.

romaninsh:
Added initial implementation which should be OK but can continue to look to expand in 1.1.x

Issue #67, opened 2017-03-13
Do we need double-formatter support in Table/Grid?


romaninsh:
Currently when you add column to grid, you can use:

``` php
$table->add('name', new \atk4\ui\TableColumn\Link('details.php?id={$id}'));
```

Columns contribute to formatting of the CELL but NOT DATA, so your "money" type will be formatted in 123,345.10 regardless. With existence of a Template (http://agile-ui.readthedocs.io/en/latest/grid.html#template) you already have a great deal of control over cell formatting. Finally you can extend TableColumn class and do more formatting yourself.

There may be cases, however, when multiple CELL formatters are necessary. For instance if you want to present "status" but also put a link over it. From the Agile UI perspective it complicates things, but if there are extensive use cases, then something like this could be implemented:

``` php
$grid->addColumn('rating', [
   new \atk4\ui\TableColumn\Template('{$rating} out of 10'),
   new \atk4\ui\TableColumn\Link('rate.php?id={$id}')
]);
```

I think this change can be done along with addition of un-named columns (e.g. radio boxes, actions, expanders etc).

DarkSide666:
+1 for this. Definitely needed functionality.

Of course we can always extend Generic formatter and do our fancy stuff there, but that would need additional coding and create a lot of NxN classes in developers project. For example, money+link, status+link, link+money+positive_negative_coloring etc.

romaninsh:
Thanks. I'd like to hear more actual examples.

PR #93, opened 2017-03-30
Finally implementing Grid

romaninsh:
Merge #74 first.

This PR implements a new Grid class with a bunch of features:

 - menu / button support on top of the grid
 - quick-search input field
 - support for checkboxes
 - support for actions (buttons)

screen shot 2017-03-30 at 11 05 35


romaninsh:
I must also add that this PR fixes HTML-escaping in Table.

src/Button.php
```diff
@@ -70,24 +70,4 @@ public function renderView()
...
-        if (is_string($url)) {
-            $this->setAttr('href', $url);
-
-            return $this;
-        }
-        $this->setAttr('href', $this->app->url($url));
-
-        return $this;
-    }
 }
```
romaninsh: moved to View

romaninsh:
Actually - docs will come in the final implementation of Grid, which I'm yet to PR.

src/Table.php
```diff
@@ -384,9 +405,19 @@ public function getRowTemplate()
     {
         $output = [];
         foreach ($this->columns as $name => $column) {
-            $field = $this->model->getElement($name);
 
-            $output[] = $column->getCellTemplate($field);
+            // If multiple formatters are defined, use the first for the header cell
```
DarkSide666: If data cell has multiple formatters, then all of them should be used here. This is not header cell here !

src/TableColumn/Checkbox.php
```diff
@@ -0,0 +1,45 @@
...
+    public $class = null;
+
+    /**
+     * Return action which will calculate and return array of all checkbox IDs, e.g.
+     *
+     * [3, 5, 20]
+     */
+    public function jsChecked()
+    {
+        return new \atk4\ui\jsExpression(' $('.$this->table->jsRender().").find('.checked.".$this->class."').closest('tr').map(function(){ ".
```
DarkSide666: can't we use simpler JS syntax here? something like:
```
$(...).find('.checked.xxx').closest('tr').get(0).data('id');
```

src/Table.php
romaninsh: I'm unclear how to implement that. I need more info in #92 and #67. The mechanism is there, so I propose we just look for bugs and various combinations then address them.

src/TableColumn/Checkbox.php
romaninsh: I'll refactor those later with JSLib. Will create a separate issue.

PR #94, opened 2017-03-30
Feature/paginator

romaninsh:
Merge #93 before (although - not a dependency)

This is implementation of Paginator component:

screen shot 2017-03-30 at 19 06 51

The component can work on it's own, it does not need to be attached anywhere. You can manually specify number of total pages to the paginator. The above output:

``` php
$layout->add(['Paginator', 'total'=>40]);
```

Multiple paginators will not affect each-other.

Paginator can reload other view which can synergise by pulling current page from paginator.

``` php
$seg = $layout->add(['View', 'ui'=>'blue segment']);
$label = $seg->add(['Label']);
$bb = $seg->add(['Paginator', 'total'=>50, 'range'=>2, 'reload'=>$seg]);
$label->addClass('blue ribbon');
$label->set('Current page: '.$bb->page);
```

screen shot 2017-03-30 at 19 11 03


src/Paginator.php
```diff
@@ -0,0 +1,192 @@
...
+     * 1, ..., 5, 6, *7*, 8, 9, ..., 34
+     *
+     * @var int
+     */
+    public $range = 4;
+
+    /**
+     * If specified, must be instance of a view which will be reloaded on selection.
+     */
+    public $reload = null;
```
DarkSide666: I don't see this implemented

romaninsh: in renderView() method.

romaninsh: https://github.com/atk4/ui/pull/94/files/bad925285feb1864eeb7f755823caec34b4e8041#diff-b7dae88f55a82631386bada74d3d1e9fR186

DarkSide666: sorry, missed these lines :)

PR #95, opened 2017-03-31
Feature/grid part4

romaninsh:
Hopefully last PR in Grid implementation.

 - added paginator
 - added documentation
 - cleaned up code
 - added comments

screen shot 2017-03-31 at 14 49 49


src/Grid.php
```diff
@@ -51,8 +92,27 @@ public function addQuickSearch($fields = [])
...
+        $this->quickSearch = $form->add(new \atk4\ui\FormField\Input(['placeholder'=>'Search', 'short_name'=>$this->name.'_q', 'icon'=>'search']))
+            ->addClass('transparent');
+
+        if (isset($_GET[$this->name.'_q'])) {
+            $q = $_GET[$this->name.'_q'];
+            $this->quickSearch->set($q);
+
+            $cond = [];
+            foreach ($fields as $field) {
+                $cond[] = [$field, 'like', '%'.$q.'%'];
```
DarkSide666: Is this safe for SQL injections?

src/Paginator.php
```diff
@@ -57,6 +47,20 @@ public function init()
     }
 
     /**
+     * Set total number of pages.
+     */
+    public function setTotal($total)
+    {
+        $this->total = (int) $total;
+
+        if ($this->page < 1) {
```
DarkSide666: `$this->page < 1` is not needed here. Only `> $this->total` is needed.

src/Grid.php
romaninsh: yep. 3rd argument is escaped.

src/Paginator.php
romaninsh: if someone would manually put page = 0, we don't any weird stuff.

Issue #50, opened 2017-02-23
Add mechanism to include CSS / JS through a call to App


romaninsh:
Some components may require additional CSS / JS includes to function. The "Application" class needs to some methods for this.

Example:

``` php
$this->app->includeJS('https://code.jquery.com/jquery-3.1.1.js');
```

Additional note here - ability to install and use bundled resources.

romaninsh:
please lets use requireJS / requireCSS

PR #120, opened 2017-04-06
Implement requireJS and requireCSS

DarkSide666:
* implements requireJS and requireCSS methods for App
* adds documentation of new metods
* App now have new property $defaultTemplate = 'html.html' for better configurability

### DIFF
--- a/docs/app.rst
+++ b/docs/app.rst
@@ -30,7 +30,7 @@ Will be true if application is currently rendering recursively through the Rende
 Links
 =====
 
-.. php:method:: url(page, extension)
+.. php:method:: url(page)
 
 Method to generate links between pages. Specified with associative array::
 
@@ -39,8 +39,6 @@ Method to generate links between pages. Specified with associative array::
 this method must respond with a properly formatted url such as::
 
     contact.php?from=John+Smith
-
-If your pages use extension other than .php, then you should pass that extension too.
 
 You may redefine this metod if you are using beautiful URLs and advanced
 routing::
--- a/docs/app.rst
+++ b/docs/app.rst
@@ -30,7 +30,7 @@ Will be true if application is currently rendering recursively through the Rende
 Links
 =====
 
-.. php:method:: url(page)
+.. php:method:: url(page, extension)
 
 Method to generate links between pages. Specified with associative array::
 
@@ -40,6 +40,8 @@ this method must respond with a properly formatted url such as::
 
     contact.php?from=John+Smith
 
+If your pages use extension other than .php, then you should pass that extension too.
+
 You may redefine this metod if you are using beautiful URLs and advanced
 routing::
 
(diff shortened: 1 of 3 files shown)


src/App.php
```diff
@@ -294,6 +303,34 @@ public function url($args = [])
...
     /**
+     * Adds additional JS script include in aplication template.
+     *
+     * @param string $url
+     *
+     * @return $this
+     */
+    public function requireJS($url)
+    {
+        $this->html->template->appendHTML('HEAD', '<script src="'.$this->url($url, null).'"></script>');
```
romaninsh: this should't call url, should simply use the $url , for simplicity.

src/App.php
```diff
@@ -282,7 +291,7 @@ public function url($args = [])
         $page = $args[0];
         unset($args[0]);
 
-        $url = $page ? ($page.'.php') : '';
+        $url = $page ? ($page.($extension ? '.'.$extension : '')) : '';
```
romaninsh: i want to encourage integrators (people who integrate Agile UI into a framework) to rewrite url() method with the goal to properly create links between PAGES. Passing extension to a page is not right.

romaninsh: I think we should rename url() into page() in the future. The full name could be getPageURL().

src/App.php
```diff
@@ -157,7 +165,7 @@ public function initLayout($layout, $options = [])
         }
         $layout->app = $this;
 
-        $this->html = new View(['defaultTemplate'=>'html.html']);
+        $this->html = new View(['defaultTemplate' => $this->defaultTemplate]);
```
romaninsh: application doesn't really have a template, as it's not extending View. Having defaultTemplate would be confusing here. If possible leave it 'html.html' for now as I don't want people to tamper with this too much at this point.

docs/app.rst
```diff
@@ -40,13 +40,25 @@ this method must respond with a properly formatted url such as::
...
+Includes
+========
+
+.. php:method:: requireJS($url)
+
+Method to include additional JavaScript file in page.
+
+.. php:method:: requireCSS($url)
+
+Method to include additional CSS stylesheet in page.
```
romaninsh: example would be nice here. Advise people to use CDN.

src/App.php
romaninsh: at best we can make it possible for someone to inject "html" property, so 

if (!$this->html) {
 ...
}

DarkSide666:
all change requests addressed

src/App.php
```diff
@@ -311,7 +305,7 @@ public function url($args = [], $extension = 'php')
      */
     public function requireJS($url)
     {
-        $this->html->template->appendHTML('HEAD', '<script src="'.$this->url($url, null).'"></script>');
+        $this->html->template->appendHTML('HEAD', '<script src="'.$url.'"></script>');
```
romaninsh: wonder if we can use getTag here...

romaninsh:
implements #50

PR #143, opened 2017-04-26
Implements password field

### LINKED ISSUE #141: Add field type support password
Would be nice to have pasword field. Currently when I set 'type'=>'password' i can still see it in CRUD and Form.

DarkSide666:
Implements password field:
* new TableColumn - in Table will show *** instead of actual value
* new FormField - in Form will use input type=password
* fixes #141

### DIFF
--- a/src/TableColumn/Password.php
+++ b/src/TableColumn/Password.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace atk4\ui\TableColumn;
+
+/**
+ * Implements Column helper for grid.
+ */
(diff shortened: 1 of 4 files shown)


src/TableColumn/Password.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace atk4\ui\TableColumn;
+
+/**
+ * Implements Column helper for grid.
```
romaninsh: test

PR #148, opened 2017-04-27
Remove protocol from links. 

DarkSide666:
This way they will work correctly on http and https sites.

fixes #147

Issue #147, opened 2017-04-27
Error while loading external files on HTTPS


romaninsh:
Looks like centered layout does not work when installed on HTTPS.

Layout requires file "css" from HTTP site

``` html
    <link rel="stylesheet" type="text/css" href="http://semantic-ui.com/dist/semantic.css">
```

<img width="336" alt="screen shot 2017-04-27 at 10 29 59" src="https://cloud.githubusercontent.com/assets/453929/25477166/802baaca-2b34-11e7-8313-b8e1bddc2909.png">

romaninsh:
Temporary workaround - "use http://" instead of "https://"

DarkSide666:
Can you try setting it this way?
`href="//semantic-ui.com/dist/semantic.css"`

without protocol. It should work and automatically pick correct protocol.

romaninsh:
i thought we switched to CDN.. weird.

romaninsh:
The issue is still there, because jslib is loaded from ui.agiletoolkit.org. We should modify our build so that it uploads complied JS file somewhere.

mvorisek:
Is this issue still valid in 2020?

DarkSide666:
I think no because now we use CDN with https:
```
    public $cdn = [
        'atk'              => 'https://cdn.jsdelivr.net/gh/atk4/ui@2.0.4/public',
        'jquery'           => 'https://cdnjs.cloudflare.com/ajax/libs/jquery/3.3.1',
        'serialize-object' => 'https://cdnjs.cloudflare.com/ajax/libs/jquery-serialize-object/2.5.0',
        'semantic-ui'      => 'https://cdn.jsdelivr.net/npm/fomantic-ui@2.7.2/dist',
    ];
```

PR #150, opened 2017-04-28
fix dependencies to use proper CDN /https for #147

romaninsh:
A better implementation than https://github.com/atk4/ui/pull/148

### DIFF
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -5,7 +5,22 @@
 Overview of Agile UI
 ====================
 
-Agile UI is a component framework that makes you type less HTML in your
+Agile UI is a PHP component framework.
+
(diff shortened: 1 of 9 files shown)


docs/overview.rst
```diff
@@ -5,7 +5,22 @@
...
+Agile UI is a PHP component framework.
+
+Before I start diving in, I'd like to make sure you understand cleary what
+Agile UI is and what does it do.
+
+ 1. Agile UI is an open-source software library/framework, that makes certain
+     tasks easier to be done for developers.
+
+ 2. In particular Agile UI focuses on 
+
```
DarkSide666: focuses on what?

romaninsh: uh that shouldn't even be there. I commited the file by accident, I am rewriting it anyway.

Issue #99, opened 2017-03-31
Overhaul documentation


romaninsh:
Make sure documentation is readable. When someone opens it at the first page and begins to read, it must flow all the way till the end.

PR #154, opened 2017-05-08
Started documentation overhaul with the Getting Started section

romaninsh:
READ DOC HERE: http://agile-ui.readthedocs.io/en/feature-documentation-restructure/

I have started re-writing documentation. Previously we were just adding patches, but now I'm starting it from the very beginning and making sure it's consistent, no references are lacking and that examples execute properly.

This PR may also depend on some other features in UI, for example:

> Currently Array persistence does not work with Pagination. The UI layer should automatically detect capabilities of Persistance and disable search/pagination if persistence does not support these features.


DarkSide666:
Great!
It's important to somehow mark these places in docs where you describe some not already implemented functionality. That way we can find that easier and fix docs when needed.

romaninsh:
#99

romaninsh:
I have narrowed it down to the following top topics:

- Overview of Agile UI
- Quickstart
- Core Concepts
- Components
- JavaScript Mapping
- Advanced Topics

PR #81, opened 2017-03-21
Feature/move jsmodal to jslib

romaninsh:
#80 implementation.,

PR #156, opened 2017-05-11
Feature/grid crud improvements

romaninsh:
Grid is now much nicer!!

 - right-aligned
 - buttons easier to click
 - animation if deleting

Known issues:
 - Grid no longer appear compact, because icon take extra border
 - clicking DELETE, then cancel, then DELETE again won't work. (see #81)

delete-crud

### DIFF
--- a/src/CRUD.php
+++ b/src/CRUD.php
@@ -104,9 +104,10 @@ class CRUD extends Grid
         }
 
         if ($this->can('d')) {
-            $this->addAction(['icon'=>'red trash'], function($j, $id) { 
+            $this->addAction(['icon'=>'red trash'], function ($j, $id) {
                 $this->model->load($id)->delete();
-                return $j->closest('tr')->transition('fade left'); 
+
+                return $j->closest('tr')->transition('fade left');
             });
         }
 
--- a/src/CRUD.php
+++ b/src/CRUD.php
@@ -87,7 +87,7 @@ class CRUD extends Grid
         $m = parent::setModel($m, $this->fieldsGrid ?: $this->fieldsDefault);
 
         if ($this->can('u')) {
-            $this->addAction(['icon'=>'edit'], new jsModal('Edit', $this->pageEdit, [$this->name=>$this->table->jsRow()->data('id')]));
+            $this->addAction(['icon'=>'edit'], new jsModal('Edit', $this->pageEdit, [$this->name=>$this->jsRow()->data('id')]));
 
             $this->pageEdit->set(function () {
                 $this->model->load($this->app->stickyGet($this->name));
--- a/src/CRUD.php
+++ b/src/CRUD.php
@@ -87,7 +87,7 @@ class CRUD extends Grid
         $m = parent::setModel($m, $this->fieldsGrid ?: $this->fieldsDefault);
 
         if ($this->can('u')) {
-            $this->addAction(new Icon('pencil'), new jsModal('Edit', $this->pageEdit, [$this->name=>$this->table->jsRow()->data('id')]));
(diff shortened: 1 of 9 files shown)


romaninsh:
Docs will go through #154

src/jsCallback.php
```diff
@@ -19,11 +23,44 @@ public function flatternArray($response)
...
+
     public function set($callback, $args = [])
     {
+        $this->args = [];
+
+        foreach ($args as $key=>$val) {
+            if (is_numeric($key)) {
+                $key = 'c'.$key;
+            }
+            $this->args[$key] = $val;
```
DarkSide666: this line should be moved into `if` statement above

src/jsCallback.php
```diff
@@ -19,11 +23,44 @@ public function flatternArray($response)
...
+        }
+
         return parent::set(function () use ($callback) {
             $chain = new jQuery(new jsExpression('this'));
-            $response = call_user_func($callback, $chain);
+
+            $values = [];
+            foreach ($this->args as $key=>$value) {
```
DarkSide666: to avoid `Avoid unused local variables such as '$value'` warning this line should be:
`foreach(array_keys($this->args) as $key) {`

src/CRUD.php
```diff
@@ -87,7 +87,7 @@ public function setModel(\atk4\data\Model $m, $defaultFields = null)
         $m = parent::setModel($m, $this->fieldsGrid ?: $this->fieldsDefault);
 
         if ($this->can('u')) {
-            $this->addAction(new Icon('pencil'), new jsModal('Edit', $this->pageEdit, [$this->name=>$this->table->jsRow()->data('id')]));
+            $this->addAction(['icon'=>'edit'], new jsModal('Edit', $this->pageEdit, [$this->name=>$this->table->jsRow()->data('id')]));
```
romaninsh: yes, correct. i'll update.

src/TableColumn/Actions.php
```diff
@@ -21,7 +27,7 @@ public function addAction($button, $callback)
         $this->actions[$name] = $button;
         $button->addClass('b_'.$name);
         $button->addClass('compact');
-        $this->table->on('click', '.b_'.$name, $callback);
+        $this->table->on('click', '.b_'.$name, $callback, [$this->table->jsRow()->data('id'), 'confirm'=>'Are you sure?']);
```
romaninsh: agree will fix.

src/View.php
```diff
@@ -886,7 +909,7 @@ public function on($event, $selector = null, $action = null, $defaults = null)
             $this->js(true)->on($event, $action);
         }
 
-        return $thisAction;
+        return new jQuery();
```
romaninsh: that's what $thisAction was :) not that i think i might have forgotten to add it into $actions[].

src/jsCallback.php
romaninsh: that's more cpu/memory intensive. I prefer to ignore the warning.

DarkSide666: ok

src/TableColumn/Actions.php
romaninsh: hmm this change no longer there......

romaninsh: nevermind.

src/jsCallback.php
romaninsh: no it's designed like this. We keep key as-is if it's symbolic.

romaninsh: i probably need some tests for that.

DarkSide666: yes you are right, sorry. I thought there is `foreach($this->args as ...)` :)

Issue #157, opened 2017-05-11
Add Grid column sorting


romaninsh:
It should be possible to change sort order through grid by clicking on column headers.

PR #163, opened 2017-05-11
Implement Grid Sorting

romaninsh:
Table will now have controls for displaying sorting (non-interractive) and Grid takes it to the next level by enabling dynamic sorting.

grid-sorting

Too bad mouse is invisible...

### DIFF
--- a/docs/grid.rst
+++ b/docs/grid.rst
@@ -98,6 +98,20 @@ method to reference value of selected checkboxes inside any :ref:`js_action`::
         'alert("Selected: "+[])', [$sel->jsChecked()]
     ));
 
+Sorting
+=======
+
+.. php:attr: $sortable
+
+When grid is associated with a model that supports order, it will automatically make itself sortable. You can
+override this behaviour by setting $sortable property to `true` or `false`. 
+
+Additionally you may set list of sortable fields to a sortable property if you wish that your grid would be
+sortable only for those columns.
+
+See also :php:attr:`Table::$sortable`.
+
+
 Advanced Usage
 ==============
 
--- a/docs/table.rst
+++ b/docs/table.rst
@@ -13,7 +13,7 @@ using a model). :ref:`no_data`
 
 
 Using Table
-==========
+===========
 
 The simplest way to create a table::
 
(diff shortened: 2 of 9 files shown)


romaninsh:
Resolve #157

romaninsh:
I'm not sure why coverage drop...

docs/table.rst
```diff
@@ -220,6 +220,35 @@ As a final note in this section - you can re-use column objects multiple times::
...
+.. php:attr:: sort_order
+
+Table does not support an interractive sorting on it's own, (but :php:class:`Grid` does), however
+you can designade columns to display headers as if table were sorted::
+
+    $table->sortable = true;
+    $table->sort_by = 'name';
+    $table->sort_order = 'ascending';
+
+This will highlight the
```
DarkSide666: something missing in text

src/Grid.php
```diff
@@ -55,6 +55,12 @@ class Grid extends View
     public $selection = null;
 
     /**
+     * Table can be sorted by clicking on column headers. This will be automatically enabled
```
DarkSide666: "**Grid** can be sorted ..." not Table

src/Grid.php
```diff
@@ -129,9 +135,42 @@ public function addAction($label, $action)
...
+
+        $this->table->on('click', 'thead>tr>th', new jsReload($this, [$this->name.'_sort'=>(new jQuery())->data('column')]));
+    }
+
     public function setModel(\atk4\data\Model $model, $columns = null)
     {
-        return $this->model = $this->table->setModel($model, $columns);
+        $this->model = $this->table->setModel($model, $columns);
+
+        if ($this->sortable === null) {
```
DarkSide666: why not simply setting $this->sortable=true as default when we define Grid class?

src/TableColumn/Generic.php
```diff
@@ -83,10 +83,8 @@ public function setAttr($attr, $value, $position = 'body')
      *
```
DarkSide666: no description of new $attr parameter

docs/grid.rst
```diff
@@ -98,6 +98,20 @@ method to reference value of selected checkboxes inside any :ref:`js_action`::
...
 
+Sorting
+=======
+
+.. php:attr: $sortable
+
+When grid is associated with a model that supports order, it will automatically make itself sortable. You can
+override this behaviour by setting $sortable property to `true` or `false`. 
+
+Additionally you may set list of sortable fields to a sortable property if you wish that your grid would be
```
DarkSide666: I don't see how this is implemented - "set list of sortable fields ..."

src/Grid.php
romaninsh: we will try and detect capability of a model to set order. E.g. if Array persistence does not support, then this will be disabled. Similar with pagination.

docs/grid.rst
romaninsh: you are right it's not, but we will need to have PR into agile data to make it work like this. A to-do before release.

PR #165, opened 2017-05-13
Form Improvements

romaninsh:
Implemented various improvements for the form:

## Currency
screen shot 2017-05-13 at 20 32 40

Now shows currency in a box, not as a text.

## DatePicker
screen shot 2017-05-13 at 20 34 17

## TimePicker
screen shot 2017-05-13 at 20 33 35

## And ...drumroll... hasOne field now shows up as drop-down:
screen shot 2017-05-13 at 20 37 02

Knows visual glitches:
 - [ ] date / time popup shows in double frame (couldn't figure out why)
 - [ ] dropdown does not show up same width as other fields (again - not sure about CSS)

Functionally all works well.

### DIFF
--- a/src/Persistence/UI.php
+++ b/src/Persistence/UI.php
@@ -17,11 +17,12 @@ use atk4\ui\Exception;
  */
 class UI extends \atk4\data\Persistence
 {
-    public $date_format = 'd/m/Y';
+    public $date_format = 'M d, Y';
 
-    public $time_format = 'h:i:s';
+    public $time_format = 'H:i';
 
-    public $datetime_format = 'D, d M Y H:i:s O';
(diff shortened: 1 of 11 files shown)


src/Persistence/UI.php
```diff
@@ -17,11 +17,12 @@
  */
 class UI extends \atk4\data\Persistence
 {
-    public $date_format = 'd/m/Y';
+    public $date_format = 'M d, Y';
```
DarkSide666: how is this format better that d/m/Y ?

romaninsh: it's not confusing for Americans and is working with the calendar widget.

Issue #173, opened 2017-06-04
Submitting form with "Enter" creates two records


romaninsh:
After much testing, I noticed that the problem ONLY OCCURS if there is a single field in a form.

``` php
require 'vendor/autoload.php';

class Brand extends \atk4\data\Model {
    public $table = 'brand';

    function init()
    {
        parent::init();

        $this->addField('name');
        $this->addField('producer');   // comment this out and it stops working!
    }
}

$db = new \atk4\data\Persistence_SQL('mysql:host=127.0.0.1;dbname=warehouse;charset=utf8', 'root', 'root');

$g = $layout->add(['CRUD']);
$g->setModel(new Brand($db));
```

Two requests are sent to the backend:

![screen shot 2017-06-24 at 16 32 56](https://user-images.githubusercontent.com/453929/27508997-e8010cd2-58fa-11e7-926c-10b48be07a2e.png)

and

![screen shot 2017-06-24 at 16 33 21](https://user-images.githubusercontent.com/453929/27508996-e2d6a6d6-58fa-11e7-8d89-a5619309206d.png)

With two fields only one request is sent:

![screen shot 2017-06-24 at 16 34 30](https://user-images.githubusercontent.com/453929/27509004-13c0b98a-58fb-11e7-9c95-62656218bf8a.png)

MAYBE it's because of error 500 ?

romaninsh:
Unable to reproduce on develop...

DarkSide666:
It's because of how browser treats 1 field forms. Some browsers send submit when you press Enter in 1-field form, but some doesn't.

Some reading:
https://stackoverflow.com/a/1370314/1466341
https://stackoverflow.com/a/39603896/1466341
https://stackoverflow.com/a/17797230/1466341

romaninsh:
this native submit should be disabled.

acicovic:
Hasn't been this solved?
If yes let's close the issue.

DarkSide666:
This should be solved now by #379.
If problem will reappear then we can always create new github issue.

Issue #155, opened 2017-05-10
Form not loading boolean value in a checkbox


skondakov:
$model = new \atk4\data\Model();
$model->addField('test', ['type' => 'boolean']);
$model['test'] = true;

$f = $page->add(new \atk4\ui\Form(['segment']));
$f->setModel($model);

// When this is done, the checkbox field "test" is not checked

romaninsh:
May be related to #130

romaninsh:
Wasn't this solved?

DarkSide666:
No it wasn't. Issue is in setting field value. I am on it now.

DarkSide666:
Most likely reason for this issue is here: https://github.com/atk4/ui/issues/281

PhilippGrashoff:
Seems fixed by #291, can we close?

PhilippGrashoff:
Tested, works fine now, closing.

Issue #130, opened 2017-04-12
Form - unable to 'un-tick' check-boxes


romaninsh:
Observed on #125 

Un-ticking checkboxes enhanced with SemanticUI `$.checkbox` still things that the checkbox is TICKED. (test on demos/form3.php) http://ui.agiletoolkit.org/demos/form3.php

It seems to be caused by serializeForm.

![checkbox-problem](https://cloud.githubusercontent.com/assets/453929/24977069/c8e26348-1fc3-11e7-81cd-60d30b07798d.gif)

romaninsh:
fixed some time ago.

Issue #127, opened 2017-04-08
Form integration demo - "Check Duplicate" does not seem to be implemented


romaninsh:
Looks like this one is still not implemented fully.

We might need to improve callback API's to make this functionality easier.

For now - lets work-around absence of https://github.com/atk4/ui/issues/77, but after it's implemented we might refactor.

PR #177, opened 2017-06-16
Reimplemented Form Validation

romaninsh:
New validation now is in line with the Agile Data 1.2 release. Define validations at the model level for multiple fields!

Several other validation ways are depreciated and some major changes which is why we'll pack this into UI 1.2.

 - form->setModel() will automatically save() data when submitted, because ... why not?
 - form->onSubmit() can execute error as before.
 - form->validate hook is no longer there, so we must review code to make sure it's not used in some field.

See if we can address the following issues with this PR:

 - [x] // moved to #179 - Create consistent addField(data_field_spec, form_field_spec) + test scripts
 - [ ] #173 Enter - two records 
 - [ ] #155 boolean not working
 - [ ] #130 unable to un-check box
 - [ ] #127 implement "Check Duplicate"
 - [ ] #126 "disable" does not work.
 - [ ] #106 captions respected , should probably be solved
 - [ ] #102 success to use Message template

### DIFF
--- a/demos/database.php
+++ b/demos/database.php
@@ -49,6 +49,7 @@ class Country extends \atk4\data\Model
 
         // look if name is unique
         $c = clone $this;
+        $c->unload();
         $c->tryLoadBy('name', $this['name']);
         if ($c->loaded() && $c->id != $this->id) {
             $errors['name'] = 'Country name must be unique';
--- a/demos/database.php
+++ b/demos/database.php
@@ -28,7 +28,7 @@ class Country extends \atk4\data\Model
         $this->addField('numcode', ['caption'=>'ISO Numeric Code', 'type'=>'number', 'required'=>true]);
         $this->addField('phonecode', ['caption'=>'Phone Prefix', 'type'=>'number']);
 
-        $this->addHook('beforeSave', function($m) {
+        $this->addHook('beforeSave', function ($m) {
             if (!$m['sys_name']) {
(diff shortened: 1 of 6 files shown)


docs/form.rst
```diff
@@ -53,26 +54,59 @@ To create a form you need the following code::
...
 
-    $form->addField('is_accept_terms', ['type'=>'boolean']);
+.. important:: Always use `'type'=>` because this also takes care of
+    `type-casting <http://agile-data.readthedocs.io/en/develop/typecasting.html>`_
+    e.g. converting data formats.
+
+Integrating Form with a Model
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+As you work on your application, in most cases you will be linknig Form with
```
DarkSide666: typo - linking

romaninsh:
I'm leaving other fixes for later.

demos/database.php
```diff
@@ -21,11 +21,40 @@ public function init()
...
+            $errors['iso'] = 'Must be exactly 2 characters';
+        }
+
+        if (strlen($this['iso3']) !== 3) {
+            $errors['iso3'] = 'Must be exactly 3 characters';
+        }
+
+        // look if name is unique
+        $c = clone $this;
+        $c->tryLoadBy('name', $this['name']);
```
DarkSide666: this still is not 100% safe way to check name uniqueness.
what is name already is not unique in database?

PR #182, opened 2017-06-19
Delegate view javascript rendering to hosting app if exist.

ibelar:
When integrating with 3rd party frameworks, Agile UI may need to output JavaScript in a different way (not as echo '<script>' tag)

This PR will check if hosting app has a method call getViewJs and delegate js rendering to it.


romaninsh:
I think that this particular approach is too "hacky" - checking for a non-existant hard-coded method name in the App. I'll think how this can be done differently.

Issue #183, opened 2017-06-21
clicking delete-cancel-delete will not work in CRUD


romaninsh:
Click "Delete" on any row in a CRUD, then cancel the confirmation. Following click on Delete won't work.

Test here:
http://ui.agiletoolkit.org/demos/crud.php

romaninsh:
fixed, can be seen here: http://ui.agiletoolkit.org/demos/crud.php

PR #184, opened 2017-06-22
Issue #183/Fixed

ibelar:
When user press cancel on confirm dialog after clicking a table delete
button, the ajaxec plugin is returning an initialized but empty data
object. Therefore, simply testing for an initialized object will
prevent calling ajaxec . We now also check that condition in order to
be able to call ajaxec plugin on an empty object as well.


romaninsh:
Linking to #183

PR #185, opened 2017-06-22
Fix sticky args in URL

DarkSide666:
* sticky args should be added even if we use current page in URL
* adds comments and slightly shorten code


romaninsh:
Not sure what's "current page" refers to. you mean index?

DarkSide666:
well, in this case yes - index page is kind of "current" page.
It's when you call url() without specifying another page.

PR #189, opened 2017-06-29
Feature/js enhance

ibelar:
Here are the changes:
## View::renderJSON()
Create a new method in View::renderJSON() that will handle js output in json format. 
- basically use it as $this->app->terminate($view->renderJSON())
  - update method in Table/Column/Delete, jsReload and VirtualPage classes to use the new renderJSON in app->terminate.
   * in regards to virtual page, since cb->triggered 'cut' is use for normal rendering and js rendering, I had to put a test using a query param to differentiate between both. May be in future you could have a new triggered value, say 'cutjs' for example for js rendering.
- renderJSON() will also send Exception in json format.
## jsModal
Change jsModal and delegate creation of the modal dialog to a new js plugin createModal.js 
  - the main advantage is that you can now easily pass semantic-ui modal option;
  - when more than one modal is need on the page, you are not duplicating html code as before within the script tag.
## reloadView.js
Change was made to this plugin and now use the semantic api method. The main advantage is that all the json response are sent via the api and catch by api.settings.succesTest function. Also allow to eval script code without duplicating it in DOM.
## Lib Atk4JS.js
- Refactor addParam to its own file;
- add createModal.js plugin;
- refactor jquery var declaration so it does not get include in final bundle. I really add trouble with webpack configuration for this.
## html.html
Refactor $.fn.api.successTest function to handle new json format.

# Testing
I have test this new code against all the demos file and did not find any issue.


romaninsh:
Great PR.

romaninsh:
Merge devel again and Travis will pass.

ibelar:
Merge is done. Thanks!

PR #195, opened 2017-07-09
Feature/misc improvements

romaninsh:
Not so big improvements.

### DIFF
--- a/src/TableColumn/Checkbox.php
+++ b/src/TableColumn/Checkbox.php
@@ -31,7 +31,7 @@ class Checkbox extends Generic
     public function getHeaderCellHTML(\atk4\data\Field $f = null, $value = null)
     {
         if (isset($f)) {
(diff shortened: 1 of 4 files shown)


src/TableColumn/Checkbox.php
```diff
@@ -31,7 +31,7 @@ public function init()
     public function getHeaderCellHTML(\atk4\data\Field $f = null, $value = null)
     {
         if (isset($f)) {
-            throw new Exception(['Checkbox must be placed in an empty column. Don\'t specify any field.', 'field'=>$f]);
+            throw new \atk4\ui\Exception(['Checkbox must be placed in an empty column. Don\'t specify any field.', 'field'=>$f]);
```
DarkSide666: If we do like this here, then probably (for consistency) we need to do like this everywhere in UI?
For example here too:
- TableColumn/Money
- TableColumn/Status
and maybe in other parts of code too?

Besides that all looks great!

romaninsh: if it's producing error there, sure.

PR #196, opened 2017-07-10
Enhance readme

romaninsh:
I have rewritten README file for the upcoming release of Agile UI 1.2. The focus is now on our premium components - CRUD and Form as well as highlighting "warehouse" application.

This is not the final / ideal form of README, but it's still better than the one we currently have.


README.md
```diff
@@ -6,44 +6,167 @@
...
+```
+
+![Login](./docs/images/login-demo.png)
+
+### Extending Layouts
+
+An add-on can add new menu items. It does not happen automatically but through a very simple line of code:
+
+``` php
+$auth->addChangePassword($app->layout->menu_user);
```
DarkSide666: why not vice-versa? `$app->layout->menu_user->addItem($auth->getChangePasswordMenuItem())`

PR #197, opened 2017-07-11
Feature/implement tabs

DarkSide666:
Basic Tab functionality is implemented.

Not implemented callbacks.
No Documentation.


romaninsh:
Thanks for submitting, looks great, just need to add dynamically loadable tabs now and we're good.

Imants - i'll also add documentation on Virtual Pages since you said it's unclear how to use them.

DarkSide666:
Yeah, VirtualPage magic should be well described. It's not that intuitive, but with it you can do many fancy things :)

romaninsh:
Working on it!

Issue #198, opened 2017-07-12
Label details='0' is invisible


romaninsh:
The following code:

```
$layout->add(['Label', 'Books available to rent: ', 'detail'=>'0']);
```

Thinks that detail is not specified at all:

<img width="154" alt="screen shot 2017-07-12 at 11 02 42" src="https://user-images.githubusercontent.com/453929/28112632-ac8857aa-66f1-11e7-9b7f-9b701f09daa0.png">

Unless 'detail' is null, it should be displayed.

DarkSide666:
will be fixed in https://github.com/atk4/ui/pull/199

PR #199, opened 2017-07-13
Fix/#198 label detail 0

ibelar:
Change how $detail value is being test to allow 0 value to pass in Label.php
Change how $content value is being test to allow 0 value but still block false value in View::recursieRender().


romaninsh:
TODO: add test-script.

DarkSide666:
adding link here to #198

PR #202, opened 2017-07-20
Fix/double spinner

ibelar:
# Changes

## Server response property change
  Change response object property that hold javascript from eval to atkjs .
Using eval as property name was conflicting when in Wordpress environment causing the javascript
code to be run twice. Once prior to ApiService::onSucces() method and once in ApiService::onSuccess().

# JS fix

## Deleting CRUD record
Now properly eval code, even in a Wordpress environment.

## Double spinner
Remove spinner from reloadView plugin since, semantic api call already load one.

## Add timeout to spinner
Add a timeout function to spinner plugin.

### DIFF
--- a/js/src/plugins/reloadView.js
+++ b/js/src/plugins/reloadView.js
@@ -3,12 +3,12 @@ import atkPlugin from 'plugins/atkPlugin';
 export default class reloadView extends atkPlugin {
 
     main() {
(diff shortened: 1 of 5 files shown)


romaninsh:
looks ok. Is "eval" now "atkjs" ?

PR #205, opened 2017-07-27
Feature/notify

ibelar:
# Notification

## notify.js as an AtkPlugin

Add js plugin for creating nice notification.

- Notification is attach to body using $.atkNotify(options)
- Can be attach to any dom element using $('selector').atkNotify(options)
- Can be personalize with:
  - type: success, error, info or warning,
  - icon: any of semantic-ui icon,
  - size: any of semantic-ui size (height) like mini, small etc. ,
  - transition for opening or closing: any of semantic-ui transition,
  - width in percentage,
  - position: topLeft, topCenter, topRight, bottomLeft, bottomCenter, bottomRight or center,
  - content: text for notification,
  - duration: time in ms.

## notify.php

For demonstration purpose in demos.


romaninsh:
Few notes after I reviewed.

First it looks like a visual glitch is introduced:

screen shot 2017-08-25 at 14 14 01

Secondly unlike many other components there are no "simple usage". Perhaps something like this:

``` php
$button->onClick(function() { 
   return new jsNotify('All went well');
});
```

We need to decide what is the best positioning for this. Also a few question about the content - can anything be in there?

My final note is that the HTML escaping is not working, we should by default escape everything.

romaninsh:
That UI layouting problem is not only in your branch, I'll have to check why it's like that.

ibelar:
Humm, don't have the layout problem on my side.

romaninsh:
weird, ok, it's off the list anyway. Will have to investigate.

romaninsh:
will do some integration testing but on course for 1.2

PR #212, opened 2017-08-22
Feature/implement tabs rev

ibelar:
# Ui Semantic Tab

When mixing static tab content with dynamic content on the same page, the tab() function need to be called on each item element. $('.item').tab() In other word, the path option for dynamic content needs to be set in order for the tab to load it's content. As oppose to static content where only the connector need to be set for showing the content.

In order to achieve mix and match of dynamic and static content, I have move the call to tab() into the tab item specifically (Tab.php) instead of Tabs.php. Thus, each tab item is responsible to setup it's content individually.

## Note

The others demo file was updated to satisfy Style CI.  Probably because I did a copy of the original pull request.


romaninsh:
Once again - Awesome work, @ibelar!

DarkSide666:
exactly - awesome!

sorry for not responding earlier.

Issue #188, opened 2017-06-25
table addColumn() argument refactor


romaninsh:
This is ticket to refactor addColumn considering the following:

 - https://github.com/atk4/core/issues/50 - refactor of seeds
 - need to improve use-cases where Table is used with static setSource()
 - need to separate addColumn() into addDecorator()

The current implementation of addColumn allows only 2 arguments which are unambiguous. 

The poposal is to clarify arguments to addColumn to the following:

 - first argument always name of the column
 - second specifies a column decorator.
 - third argument specifies model field.

The process of adding column would be:

1. If no model is associated - create a blank one.
2. If field is null, Column is added without field (useful for actions for example)
3. If field name is specified, we look for existing field.
4. If field does not exist, we create new one. Third argument may be used to describe new field.
5. Once field is established, we'll look into second argument to create 'Column'.
6. Column is always created and will be the return value of addColumn().

### First Argument

Can be either string or null. Other values will cause exception. If model has a field with specified name, specifying third argument will cause exception.

### Second Argument

Here are possible values:

 - Object (e.g. TableColumn\Link), will be associated with field and added.
 - Seed (e.g. ['Link', 'details']), [will create object](https://github.com/atk4/core/issues/50), associate with field and add.
 - String (e.g. 'Your Age') - will create Column based on field type, then set label to specified string
 - null - will create Column based on field type.

### Third Argument

Here are possible values:

 - Seed (e.g. ['type'=>'integer']) will be used to create field object. 
 - Object (e.g. new Image()) can specify Model Field object
 - String (e.g. 'integer') will be used as a type
 - null, add generic field or use existing

Use cases with only first two arguments:

``` php
// regular generic column
$t->addColumn('one');

// column with custom caption
$t->addColumn('two', 'Caption');

// column with custom decorator
$t->addColumn('three', ['Password']);

// will not add or associate with field. Yet template may reference model values.
$t->addColumn(null, ['Template', 'hello, {$two} {$one}']); 

// specifies arguments to Link column
$t->addColumn('five', ['Link', ['page2', 'static'=>1], ['id', 'type'=>'record_type']]);

// identical to above
$t->addColumn('five', new TableColumn\Link(['page2', 'static'=>1], ['id', 'type'=>'record_type']);
```

Sometimes you want to specify a data-type to a column:

``` php
// add a monetary field, will deduct the column type
$t->addColumn('one', null, ['type'=>'money']); 

// Will use Money column, but specify caption. Type can be just string.
[...]

PR #179, opened 2017-06-19
Big addColumn, addField and Seed refactor (PART1)

romaninsh:
Refactor to have a consistent addField / addColumn mechanics implementing #188  and #187 as well as incorporating https://github.com/atk4/core/issues/50.

 - addColumn, addField and addColumnDecorator arguments should be consistent
 - new Seed should be respected
 - addColumn may not be called twice on same column
 - adding addDecorator (instead of Formatter) because formatting is done in UI Persistence
 - add consistent arguments (<field name>, <form field/table column> = null, <data field> = null)
 - significantly improve docs

Things to do:

 - [ ] create a good testSutie for all addColumn addDecorator and addField scenarios
 - [ ] reflect in the code
 - [ ] reflect in the docs

PR #223, opened 2017-09-10
Big addColumn, addField and Seed refactor (PART2)

romaninsh:
In this part I've implemented Table::addColumn() as per requirements in #188.

demo/table.php works well now, but other pages still fail. Some additional documentation needed here, but other UI elements will be resolved in PART3.

Approve / Merge #179 before this.


src/Table.php
```diff
@@ -160,50 +201,58 @@ public function addColumn($name, $columnDef = null)
...
         }
+
+        return $this->_add($this->factory($seed, $defaults, 'TableColumn'));
     }
 
+    protected $typeToDecorator = [
+        'password' => 'Password',
+        'text'     => 'Text',
+        'boolean'  => ['Status', 'positive'=>[true], 'negative'=>['false']],
+    ];
```
DarkSide666: don't we have more types ?

romaninsh: did we?

PR #242, opened 2017-10-17
Feature/js notify

ibelar:
## New jsNotify Class

Notifier can now be set using 

```
new jsNotify($options, $viewToAttach);
```
You may also set options one by one using one of the various setOptionName function.

## notify.php demo file

Improve readability of notify.php demo file by setting each option individually.

## Color

Notifier now use one of the semantic ui define color for background, which make them more visible.

#### Note: Since notify.js plugin has change, atk4JS.min.js need to be rebuild.


romaninsh:
the jiggle effect on the demo does look awful, but looks like it's the best one we got now.

romaninsh:
tested locally, works well.

PR #246, opened 2017-10-25
Feature/loader

ibelar:
# Loader

Add Loader class. This loader will display spinner icon when performing time consuming action.

## Use

```
$l2 = $layout->add(new \atk4\ui\Loader(['loader'=> new atk4\ui\View(['ui'=> 'segment'])]));
$l2->set(function ($loader) {
    //set your time expensive function here.
    sleep(2);
    $loader->add(new \atk4\ui\Message(['text' => 'Load using button']));
});

//Start via user action.
$b = $layout->add(['Button', 'Start Loader']);
$b->on('click', $l2->jsStartLoader());
```

## Loader view
Any view can be attached to it when supply at Loader creation. If none is supply, one is created by default.

## Demo file
You can try this new Loader via demo loader.php


src/Loader.php
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+
+        return $this;
+    }
+
+    /**
+     * Return loader callback url when set.
+     *
+     * @return null
```
DarkSide666: string|null

src/Loader.php
```diff
@@ -0,0 +1,98 @@
...
+     * @param array|string $fx
+     * @param null         $args
+     *
+     * @throws Exception
+     *
+     * @return $this
+     */
+    public function set($fx, $args = null)
+    {
+        if (!is_object($fx) && !($fx instanceof Closure)) {
```
DarkSide666: is_object why? or maybe use !is_callable($fx)

src/Loader.php
```diff
@@ -0,0 +1,98 @@
...
+     *
+     * The loader view is pass as an argument to the loader set function.
+     * This allow to easily update the loader view content.
+     *  $l1 = $layout->add('Loader');
+     *  $l1->set(function ($loader_view) {
+     *    do_long_processing_action();
+     *    $loader_view->set('new content');
+     *  });
+     *
+     * @param array|string $fx
```
DarkSide666: incorrect param type

src/Loader.php
```diff
@@ -0,0 +1,98 @@
...
+     * @param null         $args
```
DarkSide666: array

src/Loader.php
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        $this->loaderCallback = $this->loader->add('CallbackLater');
+
+        if ($this->loaderCallback->triggered() && $fx) {
+            call_user_func($fx, $this->loader);
+            $this->app->terminate($this->renderJSON());
+        } else {
+            if ($this->needOnPageLoad) {
+                $this->loader->js(true)->atkReloadView([
```
DarkSide666: can't this use jsStartLoader() somehow to avoid code duplication?

DarkSide666:
Just for reference, this is current Loader class from smbo project:

```
class Loader extends \atk4\ui\View
{
    /**
     * Initialization.
     */
    public function init()
    {
        parent::init();

        $r = new \atk4\ui\jsReload($this, [$this->name=>'loaded']);

        if (!$this->app->stickyGET($this->name)) {
            $l = $this->add(['ui'=>'active loader']);
            $l->js(true, $r);
        }
    }

    /**
     * Sets loader callback.
     *
     * @param callable $callback
     *
     * @return $this
     */
    public function set($callback)
    {
        if ($this->app->stickyGET($this->name)) {
            call_user_func($callback, $this);
        }

        return $this;
    }
}
```

src/Loader.php
ibelar: Since Closure function are Object of type Closure, I think this is ok.

src/Loader.php
ibelar: Good point. Will fix that.

docs/virtualpage.rst
```diff
@@ -0,0 +1,293 @@
...
+    $loader = $app->add('Loader');
+    $loader->add(['Message', 'Please wait until we load LoremIpsum...', 'red']);
+    $loader->set(function($p) {
+
+        // Simulate slow-loading component
+        sleep(2);
+        $p->add('LoremIpsum');
+
+    });
+
```
ibelar: Do you think it is necessary to be able to remove the loading message once loader is finish? May be we can add this implementation inside Loader. 
or simply use code like this:
```
$loader = $app->add('Loader');
$m = $loader->add(['Message', 'Please wait until we load LoremIpsum...', 'red']);
$loader->needOnPageLoad = true;

$loader->set(function ($loader) use ($m) {
    //set your time expensive function here.
    sleep(2);
    $loader->add(new \atk4\ui\LoremIpsum(['size'=>1]));
    $m->js(true)->hide();
});
```

docs/virtualpage.rst
```diff
@@ -0,0 +1,293 @@
...
+
+Triggering Loader
+^^^^^^^^^^^^^^^^^
+
+By default, Loader will display a spinner and will start loading it's contents as soon as DOM Ready() event fires.
+Sometimes you want to control the event. 
+
```
ibelar: Loader in does not start on page load by default. You need to set it with:

```
$loader->needOnPageLoad = true;
```
But we can change the behaviour if you want.

docs/virtualpage.rst
```diff
@@ -0,0 +1,293 @@
...
+.. php:method:: jsLoad()
+
```
ibelar: name of method is jsStartLoader()

docs/virtualpage.rst
```diff
@@ -0,0 +1,293 @@
...
+    $loader = $app->add(['Loader', 'loadEvent' => 'kaboom']);
+
+    $loader->set(function($p){
+        $p->add('LoremIpsum');
+    });
+
+
+    $app->add(['Button', 'Load data'])->on('click', $loader->js()->trigger('kaboom'));
+
```
ibelar: Not sure what you mean here, this property loadEvent does not exist....

docs/virtualpage.rst
DarkSide666: I personally think that it should be `true` by default, but of course that depends on use-case.

docs/virtualpage.rst
romaninsh: I was thinking that the loading message would be appearing inside loader div and will be overwritten anyway with the desired content when loading is complete. 

As for the loader, if I understand correctly that's just a class which will be added/removed by JS, so we don't have to worry.

docs/virtualpage.rst
romaninsh: i have read your implementation very througfully, the wrote documentation which diverges slightly from your implementation. Those differences are there due to

 - better argument naming
 - more usable
 - better out of the box experience.

I suggest to adopt documented spec 100%. In this particular case, the documentation explains that the loader will load itself by default automatically as Imants have also suggested, but if you do invoke jsLoad(), that will disable that. My suggestion is to add check inside render() and track call to isLoad through a protected property.

docs/virtualpage.rst
romaninsh: Loader::jsLoad is my recommendation for method name.

docs/virtualpage.rst
romaninsh: yes, i recommend to add it. The property may be "true" by default and it can be passed as 1st argument to js() function inside render().

romaninsh:
PR low due to lack of proper ui-testsuite with coverage reflection. Can't be helped now, but I'm planning to address this in next PR.

PR #248, opened 2017-10-25
Feature/layout to app in docs

### LINKED ISSUE #247: In docs - replace $layout with $app.
$app->add() is now the suggested behaviour over $layout->add(). 

The reason is that it's simpler for beginners and cleaner in the code. Even though calling this will end up $layout->add() anyways, this have been more successful with people who are new to programming and haven't learned about layouts yet.

Someone should look through the docs and replace `$layout->add` with `$app->add`

DarkSide666:
fixes #247


DarkSide666:
there will be some conflicts with https://github.com/atk4/ui/pull/246 to solve

romaninsh:
mass replace? :D i wonder just if there are any examples that demonstrate $layout features explicitly, e.g. $layout->addMenu, can you double-check before  merging.. Manual review will be too complex anyway.

romaninsh:
OK to merge, i'll fix the #246 if needed.

DarkSide666:
It was not mass replace, but close to that. I manually checked each replace. Of course there is small possibility that i made mistake somewhere, but i also checked demo scripts and they worked fine.
If i remember correctly there was few places which demonstrate specific usage of layout (like addMenu etc.) and these are not changed.

PR #250, opened 2017-10-28
Feature/loader romans

romaninsh:
Some refactoring, better test, cleanups:@

 - Added LoaderShim class. Not sure 100% we want it like that though. Will probably change it back.
 - Added testing of reloadability inside and outside of loader.
 - Added test for nested loaders
 - Loader itself now have 'ui segment', which you can change if you wish.
 - Loader itself is reloaded, not a nested element.
 - Loader will reload itself unless jsLoad is called()

 - added ViewTester - really nice component for testing reload-ability.


romaninsh:
I'm experiencing this weird behaviour:

Watch the last counter reset to 100ms with one extra reload.

kapture 2017-10-28 at 16 19 33

src/Loader.php
```diff
@@ -39,8 +39,9 @@ public function init()
         parent::init();
 
         if (!$this->shim) {
-            $this->shim = ['View', 'class' => ['padded segment'], 'style' => ['min-height' => '7em']];
+            $this->shim = ['View', 'class' => ['ui active centered inline loader']];
```
romaninsh: sometimes this shows double-loader. Also if we disable auto-load this shows loading indicator even though it's not loading.

romaninsh: I'm ok leaving this for now like this.

DarkSide666: rolled back

PR #257, opened 2017-11-01
Fix/issue254 255 modal

ibelar:
# Fix issue 254

- Apply loading indicator to modal create either via jsModal or Modal.php 
- allow to supply loading indicator label, default to "Loading...", from jsModal or Modal

# Enhancement issue 255

- Allow to pass url argument to Modal::show($arg) method.

```
$app->add(['Button', 'Show'])->on('click', $modal->show(['color'=>'blue']));
```

## Apply some refactoring on passing data to modal

- data is now pass via the $.data() function either for Modal.php or jsModal, thus removing the html data attribute in modal.html template.


romaninsh:
Is this ready for review?

romaninsh:
I tried added test, but I'm getting a failure. Maybe I'm not doing it correctly?

screen shot 2017-11-02 at 11 55 14

ibelar:
Change the way how arguments are save to modal when using Modal::show($args).

Also remove view tester on second modal since tester will reset Message color on it's last callback because $_Get['color'] is not available when view tester reload.

romaninsh:
works perfectly now. Will merge now.

PR #258, opened 2017-11-02
Feature/sse

ibelar:
Implementation of `Server Side Event`  that allows developer to execute background tasks in PHP.

``` php
$bar = $app->add(['View', 'template' => new \atk4\ui\Template('<div id="{$_id}" class="ui teal progress">
  <div class="bar"></div>
  <div class="label">Testing SSE</div>
</div>')]);
$bar->js(true)->progress();

$button = $app->add(['Button', 'Turn On']);
// non-SSE way
//$button->on('click', $bar->js()->progress(['percent'=> 40]));

$button->on('click', $app->add('jsSSE')->set(function($sse) use ($bar) {

    $sse->send($bar->js()->progress(['percent'=> 33]));
    sleep(2);
    $sse->send($bar->js()->progress(['percent'=> 66]));
    sleep(2);
    
    // non-SSE way
    return $bar->js()->progress(['percent'=> 100]);

}));
```

## Goal of this PR

Create a drop-in replacement for jsExpression (which is used quite widely in Agile UI) that is backward-compatible, but would provide interface for sending status updates on task completion.


romaninsh:
Hey Alain. Pretty through implementation, but needs to be slightly refactored. There is wrong implementation fundamentally. Let me explain.

SSE is basically allows you to establish connection from the browser to the server. Should server close the connection, browser will re-connect, but connection should remain open for as long as possible. While connection is maintained, it can be used to deliver events. so $send should perform "echo" and "flush" but it should keep running. 

Normally SSE could be used for a social site to alway have one connections for all users and notify browser about posts and such. This could be pretty neat with Chat applications too. But PHP is not very well suited for those uses, PHP has limit on a run-time, its connections are expensive etc so we don't really want to use SSE for THOES purposes.

For us the SSE have a different use pattern. It is to establish connection to perform a complex task which, lets say, takes 10 seconds to complete.

``` php
$sse = $button->add('SSE')->set(function ($sse) use ($button) {

    for ($i = 1; $i<10; $i++) {
        sleep(1);
        $sse->send($button->js()->set(rand(1, 100));
    }
});
```

In this particular case, once the connection is terminated, it should not be reconnected. The send() would emit javascript and the browser should execute it immediately, so you would see button change it's label 10 times with just 1 SSE connection, then it would stop.

I'd also like to see SSE to extend 'jsCallback' class instead of rendering itself. Technically we can emit Action "jsReload" which will send re-render of any object we want, since SSE is just a vehicle.

http://agile-ui.readthedocs.io/en/latest/callbacks.html?highlight=jscallback#jscallback

With that our new jsSSE class could be a drop-in replacement for Callback:

``` php
$label = $app->add('Label');
$cb = $label->add('jsSSE');

$cb->set(function($sse, $arg1){

    return 'width is '.$arg1;

}, [new \atk4\ui\jsExpression( '$(window).width()' )]);

$label->detail = $cb->getURL();
$label->js('click', $cb);
```

The difference is that we can use interface of $sse to send more expressions from inside callback:

``` php
$label = $app->add('Label');
$cb = $label->add('jsSSE');

$cb->set(function($sse, $arg1){
    $sse->send(new jsNotify('Process started'));
    sleep(2);
    $sse->send(new jsNotify('Almost there'));
    sleep(2);
    $sse->send(new jsNotify('DONE'));

    return 'width is '.$arg1;

}, [new \atk4\ui\jsExpression( '$(window).width()' )]);

$label->detail = $cb->getURL();
$label->js('click', $cb);
```

I'll try to create a very basic reference implementation of jsSSE class.

src/jsCallback.php
```diff
@@ -117,8 +79,62 @@ public function set($callback, $args = [])
                 $m = new Message($e->getMessage());
                 $m->addClass('error');
 
-                $this->app->terminate(json_encode(['success' => false, 'message' => $m->getHTML()]));
+                $this->terminate($m->getHTML(), false);
+                // TODO, may have a bug here? passing HTML as ajaxec?
```
romaninsh: @ibelar - can you check?

PR #259, opened 2017-11-04
Fix/js sse on non supporting browser

ibelar:
# Fix jsSse to run callback function on non supporting browser.

On non supporting browser, like IE, the set callback function is never run.
Using this fix, the atkServerEvent.js plugin will fallback with using atkAjaxec.js when this occur, in order for the callback function define in set to be able to run and at least complete.

### DIFF
--- a/demos/sse.php
+++ b/demos/sse.php
@@ -2,31 +2,17 @@
 
 require 'init.php';
 
+$btn1 = $app->add('Button')->set('start');
+$btn2 = $app->add('Button')->set('start');
 
-//$vp = $app->add('VirtualPage');
-//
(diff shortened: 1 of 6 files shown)


romaninsh:
Should this be merged aftter #258  ?

romaninsh:
@DarkSide666 you'll have to test / review this because I don't have IE.

romaninsh:
@ibelar i had to fix a little, it was broken. can you check if it still OK as a fix?

romaninsh:
OK to ignore codecov/patch coverage, because zombiejs cant trigger SSE.

PR #277, opened 2017-11-30
Improvements for Tables, CRUDS and few other areas.

romaninsh:
Tables are good for outputting data from a model, but sometimes you need to add that extra row, some sub-header or subtotal. 

This adds example on how you can do that and adds two new hooks into table: `beforeRow` and `afterRow`

``` php
$table->template->appendHTML('SubHead', '<tr class="center aligned"><th colspan=2>This is sub-header, goes inside "thead" tag</th></tr>');
```

screen shot 2017-11-30 at 21 52 16

### DIFF
--- a/demos/table2.php
+++ b/demos/table2.php
@@ -3,10 +3,10 @@
 date_default_timezone_set('UTC');
 include 'init.php';
 
-$data = [ 
-    ['id'=>1, 'action'=>'Salary', 'amount'=>200], 
-    ['id'=>2, 'action'=>'Purchase goods', 'amount'=>-120], 
-    ['id'=>3, 'action'=>'Tax', 'amount'=>-40], 
+$data = [
+    ['id'=>1, 'action'=>'Salary', 'amount'=>200],
+    ['id'=> 2, 'action'=>'Purchase goods', 'amount'=>-120],
+    ['id'=> 3, 'action'=>'Tax', 'amount'=>-40],
 ];
 
 $p = new \atk4\data\Persistence_Array($data);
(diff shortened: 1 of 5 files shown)


demos/table2.php
```diff
@@ -0,0 +1,35 @@
...
+
+$table->template->appendHTML('SubHead', '<tr class="center aligned"><th colspan=2>This is sub-header, goes inside "thead" tag</th></tr>');
+$table->template->appendHTML('Body', '<tr class="center aligned"><td colspan=2>This is part of body, goes before other rows</td></tr>');
+
+$table->addHook('beforeRow', function ($table) {
+    if ($table->current_id == 2) {
+        $table->template->appendHTML('Body', '<tr class="center aligned"><td colspan=2>This goes above row with ID=2 ('.$table->current_row['action'].')</th></tr>');
+    } elseif ($table->current_id == 3) {
+        $table->renderRow();
+        $table->model->set(['action'=>'manually injected row after Tax', 'amount'=>0]);
```
DarkSide666: this is quite hackish way to do this, but it works of course :)

src/Table.php
```diff
@@ -426,6 +359,43 @@ public function renderView()
         return View::renderView();
     }
 
+    public function renderRow()
```
DarkSide666: No comments added for this method.

src/Table.php
```diff
@@ -426,6 +359,43 @@ public function renderView()
...
+                }
+            }
+
+            foreach ($this->columns as $name => $columns) {
+                if (!is_array($columns)) {
+                    $columns = [$columns];
+                }
+                $field = $this->model->hasElement($name);
+                foreach ($columns as $column) {
+                    if (!method_exists($column, 'getHTMLTags')) {
```
DarkSide666: please simplify this to:
```
if (method_exists($column, 'getHTMLTags')) {
    $html_tags = array_merge($column->getHTMLTags($this->model, $field), $html_tags);
}
```

No need for negation and `continue;` here.

src/Table.php
```diff
@@ -372,44 +336,13 @@ public function renderView()
         $rows = 0;
         foreach ($this->model as $this->current_id => $tmp) {
             $this->current_row = $this->model->get();
+            $this->hook('beforeRow');
```
DarkSide666: I think hook should be after updateTotals(), because if you change model values (like in example where you add row after current one), then totals will be messed up too.
On the other hand - sometimes it can be useful to add some additional conditions on updateTotals().

Also i think it would be great if we add ability to return `false` from beforeRow hook and in that case row will not be rendered (and not added to totals) at all. In that case it is good to have hook before updateTotals().

src/Table.php
romaninsh: i dont understand, please propose through a separate PR.

Issue #275, opened 2017-11-23
[Epic] - Console


romaninsh:
Console allows a background PHP process to output text which would be visible to the user in **real time**. The initial implementation would be simple, although we may introduce ability to automatically follow progress later.

``` php
$console = $app->add('Console');
$console->run(function($console) {
    $console->output('hello there');
    $console->runCommand('du -ms /usr/*');
});
```

We can also use Console to capture output from the code:

``` php
$console = $app->add('Console');
$console->setModel(new User($db), 'generateFullReport');
```

In this scenario Console will execute method of the model and show progress. Any "echo" output will be presented to the user immediately, but we can also use DebugTrait for a more sophisticated debugging.

PR #280, opened 2017-12-04
Implementing Console

romaninsh:
Implementation of `Console` component (see ticket #275). Simple use:

``` php
$app->add('Console')->set(function($console) {
    $console->output('Executing test process...');
    sleep(1);
    $console->output('Now trying something dangerous..');
    sleep(1);
    throw new \atk4\data\Exception('BOOM');
});

$app->add('Console')->setModel($model, 'generateReport');
```

screen shot 2017-12-04 at 12 08 54

 - Documentation: [docs/file.rst](docs/file.rst) TBC
 - Demo: [demos/console.php](demos/console.php)


src/Console.php
```diff
@@ -0,0 +1,249 @@
...
+    /**
+     * Output a single line to the console.
+     *
+     * @param $text string
+     *
+     * @return $this
+     */
+    public function output($text)
+    {
+        $this->_output_bypass = true;
```
DarkSide666: how about simply using `return $this->outputHTML(htmlspecialchars($text));` instead in this method to avoid code duplication?

src/Console.php
```diff
@@ -0,0 +1,249 @@
...
+     * @param $text string
+     *
+     * @return $this
+     */
+    public function outputHTML($text)
+    {
+        $this->_output_bypass = true;
+        $this->sse->send($this->js()->append($text.'<br/>'));
+        $this->_output_bypass = false;
+    }
```
DarkSide666: `return $this;` missing

src/Console.php
```diff
@@ -0,0 +1,249 @@
...
+        // temporarily override app logging
+        if (isset($model->app)) {
+            $old_logger = $model->app->logger;
+            $model->app->logger = $this;
+        }
+
+        $this->output('--[ Executing '.get_class($model).'->'.$method.' ]--------------');
+        $model->debug = true;
+        $result = call_user_func_array([$model, $method], $args);
+        $this->output('--[ Result: '.json_encode($result).' ]------------');
```
DarkSide666: why json_encode here? goal is to convert anything callback method returns to string, right?

src/jsSSE.php
```diff
@@ -15,6 +15,8 @@ class jsSSE extends jsCallback
     public $browserSupport = false;
     public $showLoader = false;
 
+    public $echoFunction = null;
```
DarkSide666: please add @var here

src/jsSSE.php
```diff
@@ -88,7 +91,11 @@ public function flush()
      */
     private function output($content)
     {
-        echo $content;
+        if ($this->echoFunction) {
```
DarkSide666: i think using `if (is_callable($this->echoFunction))` here would be better / more correct

docs/components.rst
```diff
@@ -37,17 +42,52 @@ control over even the small elements.
...
+=======================
+
+.. figure:: images/interactive.png
+   :scale: 30%
+   :align: right
+
+Interactive components rely on :ref:`callback`, :ref:`virtualpage` or :ref:`sse` to communicate with themselves in the PHP realm. You add them
+just as you would add any other component, yet they will send additional requests, like loading additional data or executing
+other code.
+
```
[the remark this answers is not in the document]
romaninsh: it actually is visible on the image, but the preview is unavailable.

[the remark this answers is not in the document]
romaninsh: anyway - addressed.

src/Console.php
```diff
@@ -0,0 +1,249 @@
...
+
+                    $output = '';
+                    $this->sse->echoFunction = function ($str) use (&$output) {
+                        $output .= $str;
+                    };
+                    $this->output($content);
+                    $this->sse->echoFunction = false;
+
+                    return $output;
+                }, 2);
```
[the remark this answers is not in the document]
romaninsh: because PHP (at least some verson) correct 1 byte to 4K for some reason.

src/Console.php
romaninsh: done.

src/Console.php
romaninsh: well, something better may be there, but not sure what. either that or export, and i dont like var_export.

src/jsSSE.php
romaninsh: this has hidden danger. If you specify a non-existent callback, nothing will happen and you will bang your head against the table just because of the typo or private method.

PR #283, opened 2017-12-05
Fix renderRow in case of hierarchical template

DarkSide666:
Fix renderRow in case of hierarchical Template.

We need to get fresh row template on each iteration! That's especially important in case of hierarchical row template.
This solution is also needed in case we use custom getHTMLTags which not always return same set of tags.

For example, row template:
```
{link}<a href="{$url}">{$text}</a>{/link}
```

And getHTMLTags:
```
    public function getHTMLTags($row, $field)
    {
        $v = $field->get();

        // If no value, then don't show link at all
        if (!$v) {
            return ['link' => ''];
        }

        // otherwise show link and fill url and text
        return ['url' => $v, 'text' => 'Click here'];
    }
```

Without this fix the first row without value will delete {link} tag from template at all and following rows which have value will not render <a>, because it's not in template anymore.

### DIFF
--- a/src/Table.php
+++ b/src/Table.php
@@ -437,45 +437,51 @@ class Table extends Lister
      */
     public function updateTotals()
     {
-        foreach ($this->totals_plan as $key => $val) {
+        foreach ($this->totals_plan as $key => $f) {
 
-            // if value is array, then we treat it as built-in or callable aggregate method
-            if (is_array($val)) {
-                $f = $val[0]; // shortcut
+            // backward-compatibility
+            if (is_array($f) && isset($f[0]) && !is_object($f[0])) {
+                $f = $f[0];
+            }
 
-                // initial value is always 0
-                if (!isset($this->totals[$key])) {
-                    $this->totals[$key] = 0;
+            // initial value is always 0
+            if (!isset($this->totals[$key])) {
+                $this->totals[$key] = 0;
+            }
+
+            // built-in methods
+            if (is_string($f)) {
+                switch ($f) {
+                    case 'sum':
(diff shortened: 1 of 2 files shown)


src/Table.php
```diff
@@ -364,10 +364,15 @@ public function renderView()
...
      * Render individual row. Override this method if you want to do more
      * decoration.
+     *
+     * @param array|\atk4\data\Model $m
      */
-    public function renderRow()
+    public function renderRow($m = null)
     {
-        $this->t_row->set($this->model);
+        $this->t_row = new Template($this->t_row_master);
```
romaninsh: i don't like that... specifically tried to avoid creating any objects during table iteration. That will slow down your big tables  A LOT.

DarkSide666: hmm... you are right about creating object on each iteration, that should slow down.
will cloning on each iteration be faster?
like $t_row = clone $t_row_master ?

romaninsh:
I think it's very rare to have complex templates inside table. You probably better of creating new component or use Lister. This PR would sacrifice a lot of performance and I don't see any benefit except for some "edge" case.

DarkSide666:
Is there any possibility to implement `reset` method in Template class?
There is `del` method which clears all tags, but if there would be `reset` method which sets all tags back to their original content, then it would solve this issue because we could use template->reset instead of template->del in renderRow. Also that way we don't need to create new template object on each iteration.

romaninsh:
Given that we need to refactor Template() anyway - we can implement some reset function there, why not. The rest of the PR should be re-formatted and re-submitted.

PR #285, opened 2017-12-07
initial commit

ibelar:
Set Semantic api cache to off by default.
Allow to set cache using 'cache' property to true.

Example when adding an autocomplete field with cache 'on':
```
$form->addField('country3', [
    'AutoComplete',
    'model'       => new Country($db),
    'placeholder' => 'Search for country by code, LV or UK',
    'search'      => ['name', 'iso', 'iso3'],
    'cache'       => true,  
]);
```

### DIFF
--- a/src/FormField/AutoComplete.php
+++ b/src/FormField/AutoComplete.php
@@ -16,6 +16,7 @@ class AutoComplete extends Input
     public $search;
 
     public $plus = false; // set this to cerate right-aligned button for adding a new a new record
+    public $cache = false; // set this to cache initial result in session storage.
 
     public function init()
     {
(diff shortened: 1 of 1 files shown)


src/FormField/AutoComplete.php
```diff
@@ -16,6 +16,7 @@ class AutoComplete extends Input
     public $search;
 
     public $plus = false; // set this to cerate right-aligned button for adding a new a new record
+    public $cache = false; // set this to cache initial result in session storage.
```
romaninsh: just a note, put comments before variable:

```
/**
  * comment
  */
  public $pluc
```

this way they are compliant with phpdoc.

romaninsh:
@ibelar , i have made some changes, can you review if it still fully works?

ibelar:
Yes, still work, as long as you set cache to 'local' cause if it is set to true only, it will not work locally. May be updating the var comment to reflect this would be good so people are aware of it. 
BTW this is not the dropdown configuration but rather the apiSettings configuration, which is one of the many  configuration options of the dropdown. Perhaps we should rename $config to $apiConfig or $apiSetting to avoid confusion. 

Taking into account that the variable name is apiConfig instead of just config, then in order to have cache to work you would have to set it this way:
```
$m->hasOne('country2', [new Country(), 'ui' => ['form' => [
    'AutoComplete',
    'plus' => true,
    'apiConfig' => ['cache'=> 'local'],
]]]);
```

romaninsh:
- ok, agree to rename $config into $apiConfig (will commit) 
- comment explains about 'local' so we're good.
- your comment seems reasonable.

I'll fix this and will merge. Also I need to add some doc.

PR #286, opened 2017-12-09
Feature/refactor totals

DarkSide666:
Total support in Grid is currently quite basic, implementing ability to add up numeric values, define per-column strategies and format the footer.

This PR offers many enhancements:

## Ability to add multiple Total lines

Sometimes you would want to define multiple total lines, for instance - SubTotals and Totals. They will be increased simultaneously, but will output in 2 separate lines. They will also have separate strategy.

However SubTotals need ability to be "inserted" multiple times too, which we can implement using hook beforeRow (#277).

``` php
$table->addTotals([
   'name'=>'Totals for {$category}',
   'category'=>false,  // will not output and will use colspan
   'amount'=>'sum'
], 'sub');

$last_category = null;
$table->beforeRow(function($row) use ($table) {
    if ($last_category == null) { 
      $last_category = $row['category'];
    } elseif ($last_category != $row['category']) {
      $table->renderTotals('sub', ['category'=>$last_category]);
    }
});
```

## Advanced calculation strategies

Here for the data that contains various dates, we find and display date range in the bottom: "1 Jul - 20 Aug"

``` php
$table->addTotals([
  'date'=>[
    'default'=> [null, null],
    'row'=>function($v, $prev) {
        return [min($prev[0], $v), max($prev[1], $v]
    },
    'total'=>function($v) {
        return $v[0]->format('M d').' - '.$v[1]->format('M d');
    },
    'format' => false, // to not use date column formatter
]);
```


DarkSide666:
Need to add tests for all different kinds of totals.

romaninsh:
wow relally good PR but do I have to write all the docs myself again ? 👍

DarkSide666:
Tomorrow I will add one more feature there. I have it now (idea) written down on paper but to sleepy to implement it now :)
Also I will write docs with examples tomorrow.

romaninsh:
something that came to my mind was ability to specify 'name' for the total rows, e.g. addTotals('row2'), then it is safer for you to modify it.

DarkSide666:
Also it would be better to move all that totals functionality outside Table in separate class Totals.
That's better because:
1. make Table class simpler
2. easier to extend just Totals class and make some changes there
3. Totals class should be added to Table invisibly for user, so user still will call Table->addTotals(...) and that will do $table->totals_plan[] = new Totals(...);

demos/table.php
```diff
@@ -35,11 +35,17 @@
     'name'   => 'Total {$_row_count} rows:',
     'surname'=> [
         // longest surname
-        function ($total, $value, $model) {
+        'compare'=>function ($total, $value, $model) {
             return strlen($value) > strlen($total) ? $value : $total;
         },
+        'title'=>function ($total, $model) {
+            return "Shortes is: ".$total;
```
DarkSide666: longest not shortest

docs/table.rst
```diff
@@ -495,20 +504,118 @@ is set. The above example can therefore be shortened::
...
 
 `init` value is typically a number, but can be any value, especially if you are going to control it's
 increment / formatting.
 
-`update` can be string - either "sum" or "increment". Other values are not supported. You can also pass
-a callable which gives you an option to perform update yourself.
+`update` can be string. Supported built-ins are:
+
+ - min
+ - inc
```
DarkSide666: inc or incr ?

docs/table.rst
```diff
@@ -495,20 +504,118 @@ is set. The above example can therefore be shortened::
...
+ - max
+ - sum
+ - avg
+
+You can use a callable which gives you an option to perform update yourself. Also, make note that `inc`
+will ONLY increment when value of the column it's associated with is not empty. If you need to get
```
DarkSide666: please be more specific what "empty" means here. Is 0 empty? Is empty string empty? Or is it just null. I think that 0 is not _empty_, but empty string is _empty_.

docs/table.rst
```diff
@@ -495,20 +504,118 @@ is set. The above example can therefore be shortened::
...
+total number of rows, use a special value `{$_row_count}`
+
+Update can also be set to `false` if you don't want update to take place. If not set, or set to `null`
+then default action (based on column type) will be invoked.
 
 `format` uses a template in a format suitable for :php:class:`Template`. Within the template you can
 reference other fields too. A benefit for using template is that type formatting is done automatically
 for you.
 
+If `format` set to `null` or omitted then default action will be used which is to display totals only
```
DarkSide666: I didn't understand this sentence.

docs/table.rst
```diff
@@ -495,20 +504,118 @@ is set. The above example can therefore be shortened::
...
+    }
+
+`update` defines callback like this::
+
+    function($total, $value, $model) {
+
+        // $total is previous value 
+        // $value is value of the column
+        // $model will be loaded to current column
+    }
```
DarkSide666: please add `return` statement here to make things clearer

romaninsh (changes requested):

mvorisek:
I have rebased https://github.com/atk4/ui/commit/56da26eb40026aa8aaaed1a0ea6f1367f5925a4d 1:1 and simplified the history as much as possible (by squashing CS fixer for example) into https://github.com/atk4/ui/commit/d34d398dd6a3a3ad2c6f93da737961697c85fccf. The diff shows zero changes.

Then I have reverted the changed file paths (files) on the latest develop and rebased the changes again (now with paths matching the current develop): all original changed: https://github.com/atk4/ui/compare/c75b7ae082a8b1755e2b573818af3926d633bed6...d34d398dd6a3a3ad2c6f93da737961697c85fccf

mvorisek:
Hi @DarkSide666, in my post above there is a link to view the real diff - what features/fixes are missing and can you rebase them on the latest develop?

Issue #290, opened 2017-12-13
form2 demo page is bad


DarkSide666:
First form contains additional fields in country form and these fields are not set in data Models and not found in database. So when we try to save values we get DSQL exception that such fields (first_name, last_name etc.) are not in database.

Similar situation with second (person) form below. We don't have table "person" in our test data at all so when saving form throws ugly dsql exception.

romaninsh:
Also need to add examples for checkbox and valuelist.

Issue #281, opened 2017-12-04
set() function of class Input does not set value


PhilippGrashoff:
$input->set('somevalue') should set its value, but it does not.

Current implementation is:
```
    public function set($value = null, $junk = null)
    {
        $this->content = $value;
    }
```

This might be the way, it works:
  ```
  public function set($value = null, $junk = null)
    {
        $this->field->set($value);
    }
```

**Comment by romaninsh** [comment]
content should work for input fields which you create without the form:

```
$app->add('Field\Input')->set('blah');
```

but set() should also work for when field is part of the form. (see also https://github.com/atk4/data/issues/262)

PhilippGrashoff:
closing as it was fixed by #291 or some later commit.

Issue #264, opened 2017-11-06
Behat/Mink: Replace Zombie driver with PhantomJS driver


romaninsh:
see "tools/test.sh". We currently use http://mink.behat.org/en/latest/drivers/zombie.html, which is not maintained and is limited and is unable to verify various features.

Instead we should replace it with PhantomJS: https://github.com/jcalderonzumba/MinkPhantomJSDriver, which should be able to handle all the tests.

PR #291, opened 2017-12-13
Enhancements of our Form

romaninsh:
It's time we focus on cleaning up Form a bit more, specifically with some Fields. This PR aims to solve number of issues:

 - #281 $field->set() should work
 - #290 improve demo
 - #155 checkbox saving

We will also add more UI tests (behat) and will try to switch ZombieJS with PhantomJS (#264).

### DIFF
--- a/src/FormField/CheckBox.php
+++ b/src/FormField/CheckBox.php
@@ -13,7 +13,20 @@ class CheckBox extends Generic
 
     public $defaultTemplate = 'formfield/checkbox.html';
 
-    public $label;
+    /**
+     * Label appears to the right of the checkbox. If label is not set specifically
+     * then the $caption property will be displayed as a label instead.
+     */
+    public $label = null;
+
+
+    public function __construct($label = null, $class = null) {
+        $this->label = $label;
+
+        if ($class) {
+            $this->addClass($class);
+        }
+    }
 
     public function init()
     {
@@ -31,8 +44,29 @@ class CheckBox extends Generic
         }
     }
 
+    public function set($value = null, $junk = null)
+    {
+        if (!is_bool($value)) {
+            throw new Exception(['Field\CheckBox::set() needs value to be a boolean', 'value'=>$value]);
+        }
+
+        parent::set($value);
+    }
+
     public function renderView()
     {
+        $this->template['label'] = $this->label ?: $this->caption;
+
+        if ($this->field ? $this->field->get() : $this->content) {
+            $this->template->set('checked', 'checked');
+        }
+
+        /**
+         * We don't want this displayed, because it can only affect "checked" status anyway
+         */
+        $this->content = null;
+
+
         $this->js(true)->checkbox();
 
         return parent::renderView();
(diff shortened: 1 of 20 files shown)


DarkSide666:
Some thoughts (ticked if fixed):
 - [ ] I don't think we need FormField/Line. It's the same as Input and I don't see any need for this class. Currently Password field extends it, but it could simply extend Input class too. Only advantage of having Input/Line as two classes I see is if we move part of decoration functionality from Input to Line and some other field classes like Money and TextArea (which have similar decoration capabilities) extends Line not Input. So Input will become more abstract class and Line will be nicely decorated Input.
 - [ ] Probably Radio don't work too because its implementation is quite similar to Checkbox
 - [ ] Methods jsInput(), getValue(), getInput(), set() and part of renderView() should be moved from Input to Generic
 - [ ] Maybe we need setValue() method too which will correctly typecast field value?

Implemented part of all that, but I don't like that messy code. Especially treating $this->content as field value and then there is template "Content" etc. That's a mess. Need to sort this out and make much cleaner and nicer codebase :)

src/FormField/CheckBox.php
```diff
@@ -13,13 +13,26 @@ class CheckBox extends Generic
...
     public $defaultTemplate = 'formfield/checkbox.html';
 
-    public $label;
+    /**
+     * Label appears to the right of the checkbox. If label is not set specifically
+     * then the $caption property will be displayed as a label instead.
+     */
+    public $label = null;
+
+    public function __construct($label = null, $class = null)
```
romaninsh: with constructors it's normal not to call parent. The idea of a constructor is to simply set some default values. It should not be complex. We do not support array argument on purpose.

src/FormField/CheckBox.php
```diff
@@ -31,10 +44,31 @@ public function init()
...
+        }
+
+        /*
+         * We don't want this displayed, because it can only affect "checked" status anyway
+         */
+        $this->content = null;
+
         $this->js(true)->checkbox();
 
+        $this->content = null; // no content again
```
romaninsh: Due to the way how template defines $Content we can still add a more sophisticated label (such as link or custom message) inside a label.

src/FormField/Input.php
```diff
@@ -177,9 +180,4 @@ public function addAction($defaults = [])
 
         return $this->action;
     }
-
-    public function set($value = null, $junk = null)
```
romaninsh: this has gone to Generic.

romaninsh:
- Input field is essential. You will appreciate it more as we add more non-input fields in the future (e.g. capcha or multi-input fields such as "day/month/year)
 - disagree on moving stuff to Generic. Removed some of them, because not all fields will support it or it may be too confusing.
 - setValue() - not needed. set() will call model->set which will typecast, once https://github.com/atk4/data/issues/262 is fixed.
 - content is value. I've tried moving it back and fourth but this is probably the best. I have added comments for Caption and Label to avoid confusion.

PR #292, opened 2017-12-16
Fix implementation of Radio field

romaninsh:
Implementation of radio selection field. Depends on #294. Simple use:

``` php
$form->addField('gender', ['Radio', 'values'=>['F'=>'Female', 'M'=>'Male']]);
```

Works with `enum` and `values`.

screen shot 2017-12-16 at 17 23 58

 - Documentation: [docs/file.rst](docs/file.rst) TODO
 - Demo: [demos/file.php](demos/file.php) TODO

### DIFF
--- a/src/FormField/CheckBox.php
+++ b/src/FormField/CheckBox.php
@@ -13,7 +13,20 @@ class CheckBox extends Generic
 
     public $defaultTemplate = 'formfield/checkbox.html';
 
-    public $label;
+    /**
+     * Label appears to the right of the checkbox. If label is not set specifically
+     * then the $caption property will be displayed as a label instead.
+     */
+    public $label = null;
+
+
+    public function __construct($label = null, $class = null) {
+        $this->label = $label;
+
+        if ($class) {
+            $this->addClass($class);
+        }
+    }
 
     public function init()
     {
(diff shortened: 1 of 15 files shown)


src/FormField/CheckBox.php
```diff
@@ -31,10 +44,31 @@ public function init()
...
+        }
+
+        /*
+         * We don't want this displayed, because it can only affect "checked" status anyway
+         */
+        $this->content = null;
+
         $this->js(true)->checkbox();
 
+        $this->content = null; // no content again
```
DarkSide666: why we need to set content=null again here? can js()->checkbox() change it?

src/FormField/Radio.php
```diff
@@ -2,38 +2,45 @@
...
     {
-        $this->js(true)->checkbox();
+        if (!$this->model) {
+            $p = new \atk4\data\Persistence_Static($this->values);
+            $m = new \atk4\data\Model($p);
+        }
+
+        $value = $this->field ? $this->field->get() : $this->content;
+
+        $this->lister->setModel($m);
```
DarkSide666: Probably we need this line like this:
```
$this->lister->setModel($this->model ?: $m);
```

src/FormField/CheckBox.php
romaninsh: because we output it differently. Normally for inputs the content is a valid spot, but for Radio it is not used. We keep the tag still because we can add other views in.

PR #294, opened 2017-12-16
Move cloneRegion into lister's init

romaninsh:
Implementation of #292 needs to specify "name" for radio buttons from inside Radio class. Unfortunately currently Radio does not have access to Lister's template.

This PR requires #293 which makes Lister's template available inside init() so we can clone regions right away. This enables the following code:

``` php
$myview->add('Lister', 'MyRows');
$myview->t_row['_name'] = 'abc';
```

 - Documentation: n/a
 - Demo: n/a


romaninsh:
depends on https://github.com/atk4/data/pull/266 - then must re-run test-suite.

Issue #23, opened 2016-12-27
Basic Lister


romaninsh:
Implement lister

romaninsh:
Lister is implemented, but still need documentation.

Issue #11, opened 2016-09-23
0.3 Lister


romaninsh:
Implement Lister and integration with Model

PR #298, opened 2017-12-21
Added implementation of Lister

romaninsh:
Implementation of Lister see #11, #23 and #27. Simple use:

``` php
$app->add(['Lister', 'defaultTemplate'=>'lister.html'])->setSource([
    ['icon'=>'map marker', 'title'=>'Krolewskie Jadlo', 'descr'=>'An excellent polish restaurant, quick delivery and hearty, filling meals'],
    ['icon'=>'map marker', 'title'=>'Xian Famous Foods', 'descr'=>'A taste of Shaanxi\'s delicious culinary traditions, with delights like spicy cold noodles and lamb burgers.'],
    ['icon'=>'check', 'title'=>'Sapporo Haru', 'descr'=>'Greenpoint\'s best choice for quick and delicious sushi'],
]);
```

screen shot 2017-12-21 at 04 18 07

or

``` php
$view = $app->add(['View', 'template' => new \atk4\ui\Template('
<div class="ui header">Top 20 countries (alphabetically)</div>
{List}<div class="ui icon label"><i class="{iso}ae{/} flag"></i> {name}andorra{/}</div>{/}
</div>')]);

$view->add('Lister', 'List')
    ->addHook('beforeRow', function($l){ $l->current_row['iso']=strtolower($l->current_row['iso']); })
    ->setModel(new Country($db))
    ->setLimit(20);
```

screen shot 2017-12-21 at 04 18 17

 - Documentation: [docs/lister.rst](docs/lister.rst)
 - Demo: [demos/lister.php](demos/lister.php)

### DIFF
--- a/src/Lister.php
+++ b/src/Lister.php
@@ -44,17 +44,9 @@ class Lister extends View
                 continue;
             }
 
-            if ($this->t_row->hasTag('_title')) {
-                $this->t_row->set('_title', $this->model[$this->model->title_field]);
-            }
-
-            if ($this->t_row->hasTag('_href')) {
-                $this->t_row->set('_href', $this->app->url(['id'=>$this->current_id]));
-            }
-
-            if ($this->t_row->hasTag('_id')) {
-                $this->t_row->set('_id', $this->current_id);
-            }
+            $this->t_row->trySet('_title', $this->model[$this->model->title_field]);
+            $this->t_row->trySet('_href', $this->app->url(['id'=>$this->current_id]));
+            $this->t_row->trySet('_id', $this->current_id);
 
             if ($this->t_row == $this->template) {
                 $rowHTML .= $this->t_row->set($this->current_row)->render();
--- a/src/Lister.php
+++ b/src/Lister.php
@@ -9,9 +9,6 @@ class Lister extends View
     // @var Template
     public $t_row = null;
 
-    // @var Template
-    public $t_totals = null;
-
     public $defaultTemplate = null;
 
     public function init()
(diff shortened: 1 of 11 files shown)


src/Lister.php
```diff
@@ -33,15 +37,31 @@ public function renderView()
...
                 continue;
             }
 
-            $rowHTML = $this->t_row->set($this->current_row)->render();
-            $this->template->appendHTML('rows', $rowHTML);
+            $this->t_row->trySet('_title', $this->model[$this->model->title_field]);
+            $this->t_row->trySet('_href', $this->app->url(['id'=>$this->current_id]));
+            $this->t_row->trySet('_id', $this->current_id);
+
+            if ($this->t_row == $this->template) {
```
DarkSide666: I slightly don't like this condition to be executed on every loop. That can have bad performancevespeciallyif templates are longer. What if we normalize template in `initChunks()` - if template don't have `{rows}{row}` tags, then add them and further work with only them in `renderView()`?

src/View.php
```diff
@@ -197,12 +197,15 @@ public function setModel(\atk4\data\Model $m)
...
+     * @param array $data   Array of data
+     * @param array $fields Limit model to particular fields
      *
      * @return \atk4\data\Model
      */
-    public function setSource(array $data)
+    public function setSource(array $data, $fields = null)
     {
+        $this->setModel(new \atk4\data\Model(new \atk4\data\Persistence_Static($data)), $fields);
+        /*
```
DarkSide666: Should we leave all that uncommented block here?

src/Lister.php
romaninsh: i actually had some problems with it. Don't, mind if that's in a separate PR. Performance wise, it's just comparison, shouldn't be too bad.

src/View.php
romaninsh: I'll remove.

PR #297, opened 2017-12-21
Feature/fix 295

romaninsh:
Fixes #295 and #296.

Issue #296, opened 2017-12-21
Adding child and re-rendering has no effect


romaninsh:
When render() is executed, it sets a flag (as a fix to #85) to prevent it form being re-rendering. It is still possible to add some children into the view, but when you call render() again the result is not what you expect.

``` php
var_Dump(htmlspecialchars($v->render()));
$v->add('LoremIpsum');
var_Dump(htmlspecialchars($v->render()));
```

The second var_dump currently does not contain LoremIpsum. We do not want to re-introduce #85, so instead there should be a check preventing from adding any children into view that was already rendered.

Issue #295, opened 2017-12-19
CRUD edit form is broken


romaninsh:
CRUD stopped working after, probably, #293. The cause is that adding Modal from "renderView" no longer works. Proof of concept:

``` php
class MyView extends \atk4\ui\View {
    protected $button;

    function init() {
        parent::init();
        $this->button = $this->add(['Button', 'Click me!', 'green']);
    }

    function renderView() {
        $this->button->on('click', $this->add('Modal')->set(function($p) {
            $p->add('LoremIpsum');
        })->show());
        parent::renderView();
    }
}

$app->add(new MyView());
```

The result is that "Edit" window is not populated:

<img width="1326" alt="screen shot 2017-12-19 at 13 46 12" src="https://user-images.githubusercontent.com/453929/34159993-26743558-e4c3-11e7-8cfe-da09f672cfac.png">

romaninsh:
I have finally found the root of the problem. Basically adding Modal from "renderView" no longer works. Proof of concept:

``` php
class MyView extends \atk4\ui\View {
    protected $button;

    function init() {
        parent::init();
        $this->button = $this->add(['Button', 'Click me!', 'green']);
    }

    function renderView() {
        $this->button->on('click', $this->add('Modal')->set(function($p) {
            $p->add('LoremIpsum');
        })->show());
        parent::renderView();
    }
}


$app->add(new MyView());
```

PR #314, opened 2018-01-06
resolve #313

### LINKED ISSUE #313: Paginator with self-reloading option is broken after change to CallbackLater

romaninsh:
Fixes #313 by reverting a change introduced in #297 .

https://github.com/atk4/ui/pull/297/files#diff-d36b8180c676ceb4cd32cea1f15f2de2L20

 - [ ] double check that #295 still works
 - [x] double check that #296 still works


romaninsh:
CRUD was changed not to use code as demonstrated in #295  (adding modal in renderView). Ideally it should throw exception, but not sure why it does not.

Finally tests broke because dependency was broken (oracle tests), should be fixed soon.

PR #320, opened 2018-01-10
Feature/js package

ibelar:
This PR is an enhancement to js package.

This should be npm package ready if we decide to go that route. 

- The package.json file has been update to reflect more the ownership, package name and description from the original;
- Build file name has been change from atk4JS to atkjs-ui ;
     - think it is better to use atkjs-ui or eventually the new framework name, remove 4 in name since it is representing the old version number. (I think)
- Entry file name has been changed from index.js to agile-toolkit-ui.js 
  - Since the package now export the library, if we decide to create an npm module, then we could use import atk from ‘agile-toolkit-ui’
  - global var named atk is also available globally in js


romaninsh:
should this be merged before upload field?

ibelar:
Yes please do. Remember, file name atk4JS.js has been changed to atkjs-ui, if your ok with the change than merge.

ibelar:
Sorry, forgot the previous comment. In fact, fileUpload should be merge first. Then you can merge this PR.

romaninsh:
OK. understood.

ibelar:
### Update apiService to display all error in Modal.
- replace call to alert(error) for apiService.showErrorModal(error)

### Update with change on PR #331

PR #325, opened 2018-01-11
Always pass Console as 1st parameter to model callback method

DarkSide666:
We almost always need to pass Console object to model method because we use Console for output.


DarkSide666:
closing

Issue #318, opened 2018-01-10
Form error reloads in new popup window


DarkSide666:
See video: https://www.useloom.com/share/d2d10646e443491895448f9b306f3f5c

romaninsh:
I've seen this in the past. Issue is with modal dialog zIndex..

DarkSide666:
But shouldn't it lreload in existing window where we have form?
Now it reloads in new window.

romaninsh:
re-opening because this issue is still worked on.

PR #326, opened 2018-01-11
fix/#318 Form error

ibelar:
This PR fix modal behaviour when an exception is thrown during Form::save()

Somehow, semantic-ui create the modal generated by the exception with css position attribute set to ‘static’, therefore static positioned element are alway stacked under other positioned element, regardless of the z-index attribute.

The fix will close others open modal in order to only show the modal containing the exception thrown by Form::save()

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -375,7 +375,7 @@ class Form extends View //implements \ArrayAccess - temporarily so that our buil
 
         $cb->set(function () {
             $caught = function ($e) {
-                return new jsExpression('$([html]).modal("show")', [
+                return new jsExpression('$([html]).modal("hide others").modal("show")', [
                     'html' => '<div class="ui fullscreen modal"> <i class="close icon"></i> <div class="header"> '.
                     htmlspecialchars(get_class($e)).
                     ' </div> <div class="content"> '.


romaninsh:
a much needed fix. Don't worry about coverage for JS-related patches until #264 is solved.

DarkSide666:
tested - it works good. merging.
fixes #318

romaninsh:
so when you close the other modal (with exception) does the form modal re-appear?

DarkSide666:
no

romaninsh:
so if i fill out the form and it results in exception, when i close exception, i can't get back to the form? Meaning i loose data?? ouch!

DarkSide666:
yes, it looks like that.
I guess will have to test more carefully. Actually it should not hide anything, just add another popup which is over all existing popups.

PhilippGrashoff:
Hi,

when ****ing up a callback inside a modal, e.g. putting an echo inside form submit, the error modal is displayed correctly above the other modal:
modalovermodal

PhilippGrashoff:
If I **** up a callback inside a modal, e.g. putting an echo inside form submit, the error modal gets displayed correctly above the other modal:
modalovermodal

PR #328, opened 2018-01-11
fix #327: title_field and id_field

### LINKED ISSUE #327: AutoComplete does not work if model id_field != 'id' or title_field' != name
the Dropdown relies on getting 'name' and 'id':
```
        $chain->dropdown([
            'fields'      => ['name' => 'name', 'value' => 'id'/*, 'text' => 'description'*/],
```

But if model has an id_field not named 'id' or the title_field is not named 'name', the Autoloader breaks.
Currently, getData() creates the data using model->export():

`$data = $this->model->export([$this->model->id_field, $this->model->title_field]);`

The fix has to happen here, as init() may not know the model yet.

PhilippGrashoff:
defined in init(), the Autocomplete field requires getting data 'name' and 'id'.
If model->title_field was not 'name' or model->id_field was not 'id', the AutoComplete broke.
I fixed getData() to always return the data as ['id' =>..., 'name' =>...]
A fix in init() was not possible as init() might not know the model.

fixes #327

### DIFF
--- a/src/FormField/AutoComplete.php
+++ b/src/FormField/AutoComplete.php
@@ -128,14 +128,13 @@ class AutoComplete extends Input
             }
         }
 
-      
         //fix #327: https://github.com/atk4/ui/issues/327
         $data = [];
-		$res = $this->model->export([$this->model->id_field, $this->model->title_field]);
-		foreach($res as $item) {
-			$data[] = ['id' => $item[$this->model->id_field], 'name' => $item[$this->model->title_field]];
-		}
-		
+        $res = $this->model->export([$this->model->id_field, $this->model->title_field]);
+        foreach ($res as $item) {
+            $data[] = ['id' => $item[$this->model->id_field], 'name' => $item[$this->model->title_field]];
+        }
+        
         if ($this->empty) {
             array_unshift($data, ['id' => 0, 'name' => $this->empty]);
         }
--- a/src/FormField/AutoComplete.php
+++ b/src/FormField/AutoComplete.php
@@ -128,10 +128,16 @@ class AutoComplete extends Input
             }
         }
 
-        $data = $this->model->export([$this->model->id_field, $this->model->title_field]);
-
+      
+        //fix #327: https://github.com/atk4/ui/issues/327
+        $data = [];
+		$res = $this->model->export([$this->model->id_field, $this->model->title_field]);
+		foreach($res as $item) {
+			$data[] = ['id' => $item[$this->model->id_field], 'name' => $item[$this->model->title_field]];
+		}
+		
         if ($this->empty) {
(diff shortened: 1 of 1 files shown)


src/FormField/AutoComplete.php
```diff
@@ -128,10 +128,15 @@ public function getData()
             }
         }
 
-        $data = $this->model->export([$this->model->id_field, $this->model->title_field]);
-
+        //fix #327: https://github.com/atk4/ui/issues/327
+        $data = [];
+        $res = $this->model->export([$this->model->id_field, $this->model->title_field]);
+        foreach ($res as $item) {
```
DarkSide666: This could be slow.
Simple tweak would be to wrap foreach in
`if ($this->model->id_field != 'id' || $this->model->title_field != 'name')`

src/FormField/AutoComplete.php
```diff
@@ -128,10 +128,15 @@ public function getData()
...
+            $data[] = ['id' => $item[$this->model->id_field], 'name' => $item[$this->model->title_field]];
+        }
+        
         if ($this->empty) {
-            array_unshift($data, [$this->model->id_field => 0, $this->model->title_field => $this->empty]);
+            array_unshift($data, ['id' => 0, 'name' => $this->empty]);
```
DarkSide666: id = 0 could have another meaning. Maybe you need `id = ''` here?

src/FormField/AutoComplete.php
romaninsh: agreed, would deal with majority of cases.

src/FormField/AutoComplete.php
romaninsh: i think @ibelar made it so id=0 works just fine.

DarkSide666: You know, theoretically there can be models with id=0 as real id value. Theoretically :)
OK, need to test this.

PhilippGrashoff: just took over id => 0 from the original implementation. But you are right, a real ID could be 0.

src/FormField/AutoComplete.php
PhilippGrashoff: The wrap makes sense, though the performance issue shouldn't be massive as long as a sensible LIMIT (standard 10) is set.

src/FormField/AutoComplete.php
romaninsh: yes, leave it, it is a separate issue, you are not making things worse.

DarkSide666: ok i agree

src/FormField/AutoComplete.php
DarkSide666: well ... if limit is always set and it's such small (10, 20 rows), then wrapping will not make performance better. So I accept these changes and let Romans decide to merge or not to merge :)

PR #329, opened 2018-01-15
feature/Error-modal(fix#318)

ibelar:
## Enhance error display in modal by hiding the stack trace by default

This PR will change how modal display error, by hiding stack trace in order for Error modal to stay on top of other modals.
It allow for other modal to stay open and get a chance to re-save form data.

Stack trace view can be toggle on or off by clicking the Trace button on top of modal.
 - note that stack trace view has a maximum height and must be scroll to view it’s content.

### Note: 
seem like if modal content is longer than window content, i.e. when displaying trace content, semantic-ui will position the modal as ‘static’. This means that it will always show under other modal. 

## Important
For this pr to work, atk4JS need rebuild and Core Exception::getHtml() method need to be changed as follow:

```
public function getHtml()
{
//previous code…
//add class atk-stack to this div
$output .= '<div class="ui top attached segment atk-stack">';
$output .= '<div class="ui top attached label">Stack Trace</div>';
//previous code…
}
```

trace-close
trace-open

### DIFF
--- a/js/src/services/ApiService.js
+++ b/js/src/services/ApiService.js
@@ -148,12 +148,21 @@ class ApiService {
         //catch application error and display them in a new modal window.
         var m = $("<div>")
             .appendTo('body')
-            .addClass('ui scrolling modal')
+            .addClass('ui modal')
             .css('padding', '1em')
             .html(errorMsg);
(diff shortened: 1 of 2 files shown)


romaninsh:
Not a fan of hiding things on error messages. In practice, users forget to "expand" them before taking screenshot.

romaninsh:
closing in favour of #331

Issue #307, opened 2018-01-04
refactor url() into View


romaninsh:
Currently to find out target url() we used `$app->url()`. When we are working in context of a view, sometimes we need to know "what's the url that renders this view?". Here is proposed implementation:

``` php
$msg = $tab->add('Message');

$button->link($msg->url());
```

When `View::url()` is called, it would:

 - consult `$this->owner` when unsure.
 - if enclosed in VirtualPage, will add the trigger
 - if enclosed on Tab, add #hash for activating the tab
 - if enclosed on a Loader, also make sure trigger is set
 - if no proper $app is set, throw exception

Some Views can add arguments based on state, for example `Grid` can include page number for #110 to help avoid reliance on session.

Subsequently jsReload() would be refactored to use this `View::url()` method. This also will help developers easily identify which Views cannot be 're-traced' back.

Finally this should help us avoid problems, where beginners are messing up with sticky get.

Cons: slower execution, so perhaps we should think on how to use for TableColumn\Link.

DarkSide666:
StickyGET and URLs as such is a big confusion in UI. So I will vote for having view->url() if that makes it less confusing.
On other hand - I still remember you said that will be bad for security to have view->url() when I proposed such approach back in "toolkit times" :)

ibelar:
If doing so, please make sure that App may have the last word on url somehow and url does not rely entirely on View. As an example, Wordpress integration require to add specific argument in order for ajax to properly run in WP.

```
class AtkWpApp extends App
{
    public function url()
   {
       $this->page = 'admin-ajax';
        if (!WpUtil::isAdmin()) {
            $this->page = WpUtil::getBaseAdminUrl().'admin-ajax';
        }
        $sticky = $this->sticky_get_arguments;
        $sticky['action'] = $this->plugin->getPluginName();
        $sticky['atkwp'] = $this->plugin->getWpComponentId();

        if ($this->plugin->getComponentCount() > 0) {
            $sticky['atkwp-count'] = $this->plugin->getComponentCount();
        }

        if ($this->plugin->config->getConfig('plugin/use_nounce', false)) {
            $sticky['_ajax_nonce'] = helpers\WpUtil::createWpNounce($this->plugin->getPluginName());
        }
  }
}
```

romaninsh:
Here is how i think it would work.

1. each View would have "stickyGetArgs" property = [] by default
2. calling $view->stickyGet() simply adds into that property.
3. adding a child would inherit stickyGetArgs
[...]

PR #333, opened 2018-01-16
Implement $view->url() functionality as per #307

romaninsh:
Implementation of #307. Allows to set view-specific stickyGet options:

``` php
$form->stickyGet('model');
```

Now if form will be submitted it will retain $_GET['model'], but the rest of the callbacks around the form (before or after) will not have that sticky.

This makes it much more reliable to set sticky for virtual pages.

## Refactor jsReload()

Although jsReload still used to create reload event, the callback is not handled by jsReload anymore but rather by the view itself. This was possible to implement thanks to the new url() handling. This is more consistent and should solve many reload-related issues.

## Catching runaway callbacks

Previously if for some reason callback is not executed, you'd get a full HTML render back. Now this is tracked and you get error about unhandled callback request.

## urlTrigger property (for paginator and callback)

You can now specify which URL attribute be used to trigger callback or paginate. (name of the view by default)

## Other

Cleanups and new documentaiton

 - POST_trigger property renamed into postTrigger
 - simplified callback names for crud's edit and crud's add pages.

TODO:
 - [x] need tests for few changes i added
 - [x] need some fixes in docs
 - [x] do we need sticky2.php demo at all? (making them nicer)
 - [x] in demos/crud.php we get atk4\ui\Exception: Unable to set stickyGet after url() has been used here or by a child when trying to edit and save any record
 - [x] Travis tests fail on PHP 5.6

### DIFF
--- a/demos/sticky2.php
+++ b/demos/sticky2.php
@@ -14,6 +14,8 @@ if (isset($_GET['name'])) {
     // frame will generate URL with sticky parameter
     $frame->add(['Label', 'Name:', 'detail'=>$_GET['name'], 'black'])->link($frame->url());
 
+    $frame->add(['Menu'])->addItem('Test', ['foo'=>'bar']);
+
     // app still generates URL without localized sticky
     $frame->add(['Label', 'Reset', 'iconRight'=>'close', 'black'])->link($app->url());
     $frame->add(['ui'=>'hidden divider']);
--- a/demos/sticky2.php
+++ b/demos/sticky2.php
@@ -19,17 +19,16 @@ if (isset($_GET['name'])) {
     $frame->add(['ui'=>'hidden divider']);
 
     // nested interractive elemetns will respect lockal sticky get
-    $frame->add(['Button', 'Triggering callback here will inherit color'])->on('click', function() {
+    $frame->add(['Button', 'Triggering callback here will inherit color'])->on('click', function () {
         return new \atk4\ui\jsNotify('Color was = '.$_GET['name']);
     });
 }
-
 
 $t = $app->add(['Table']);
 $t->setSource(['Red', 'Green', 'Blue']);
 $t->addDecorator('name', ['Link', [], ['name']]);
 
 $frame = $app->add(['ui'=>'green segment']);
-$frame->add(['Button', 'does not inherit sticky get'])->on('click', function() {
+$frame->add(['Button', 'does not inherit sticky get'])->on('click', function () {
     return new \atk4\ui\jsNotify('$_GET = '.json_encode($_GET));
 });
--- a/demos/sticky2.php
+++ b/demos/sticky2.php
@@ -1,0 +1,12 @@
+<?php
+
+require 'init.php';
+
+$b = $app->add(['Button', 'b1='.@$_GET['b1']]);
+$b->link(['b1'=>$b->stickyGet('b1')+1]);
+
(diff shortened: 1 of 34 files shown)


src/Paginator.php
```diff
@@ -29,6 +29,13 @@ class Paginator extends View
      */
     public $range = 4;
 
+    /**
+     * Set this if you want GET argument name to look beautifully.
+     *
+     * @var null|string
+     */
+    public $getArgName = null;
```
DarkSide666: Probably need to add something in docs about this property and getPageURL() method

src/Lister.php
```diff
@@ -45,7 +45,7 @@ public function renderView()
             }
 
             $this->t_row->trySet('_title', $this->model[$this->model->title_field]);
-            $this->t_row->trySet('_href', $this->app->url(['id'=>$this->current_id]));
+            $this->t_row->trySet('_href', $this->url(['id'=>$this->current_id]));
```
DarkSide666: need to test this change

src/Menu.php
```diff
@@ -38,7 +38,7 @@ public function addItem($item = null, $action = null)
         $item = $this->add($item)->setElement('a');
 
         if (is_array($action)) {
-            $action = $this->app->url($action);
+            $action = $this->url($action);
```
DarkSide666: need to test this change

src/Tabs.php
```diff
@@ -50,7 +50,7 @@ public function addTab($name = null, $action = null)
             } else {
                 // otherwise treat it as URL
                 //# TODO: refactor this ugly hack
-                $item->setPath(str_replace('.php.', '.', $this->app->url($action)).'#');
+                $item->setPath(str_replace('.php.', '.', $this->url($action)).'#');
```
DarkSide666: need to test this change

demos/sticky2.php
```diff
@@ -0,0 +1,9 @@
+<?php
```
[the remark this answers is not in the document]
romaninsh: addressed.

src/Lister.php
romaninsh: Tested locally with:

``` php
    $frame->add(['Lister', "defaultTemplate"=>"lister.html"])->setSource(['foo','bar','baz']);
```

so works fine, but no good demo for this.

src/Menu.php
romaninsh: Tested with 

``` php
    $frame->add(['Menu'])->addItem('Test', ['foo'=>'bar']);
```

but no good demo here.

src/Tabs.php
romaninsh: did a manual test by adding sticky on form.php / form2.php, worked fine.

src/Form.php
```diff
@@ -366,7 +366,7 @@ public function renderView()
      */
     public function ajaxSubmit()
     {
-        $this->_add($cb = new jsCallback(), ['desired_name' => 'submit', 'POST_trigger' => true]);
+        $this->_add($cb = new jsCallback(), ['desired_name' => 'submit', 'postTrigger' => $this->name.'_submit']);
```
romaninsh: @ibelar  - can you confirm that form's submission includes a post parameter which is consistent with this naming? Previously we used $callback->name but ATK shortens long names, this way it's more correct.

Issue #324, opened 2018-01-11
Modal should be refreshed after adding dynamic content


DarkSide666:
If we have dynamic content in modal (for example Console), then modal should be refreshed to correctly position it.

Please read https://forums.meteor.com/t/position-of-modal-semantic-ui-popup/28429/2 and maybe that can help you somehow to get idea what's needed.

ibelar:
Imants;

Have you tried refreshing the modal from inside console after updating content.
something like:
```
$console->send($modal->js()->modal('refresh'));
```

Also can you provide a code sample on how to reproduce.

thanks;

DarkSide666:
Page:
```
// Set up Modal
$mod = $app->add(['Modal', 'title' => 'Generate']);
$mod->addScrolling(); // I wonder if this works at all
$mod->notClosable(); // Don't allow to close Modal by clicking outside of it
$mod->set(function ($modal) use ($m){
    $modal->add('Console')->setModel($m, 'gen');
});

// Add button
$app->add(['Button','Generate'])->on('click', $mod->show());
```

Model:
```
    public function gen() {
        $this->log('info', 'Starting long process');

        for($i=0;$i<25;$i++) {
            $this->debug('test = '.$i);
            sleep(1);
        }

        return 'Success';
    }
```

Video:
https://www.useloom.com/share/668785beb4e34fc5ae735dd950c68ab5

See Console is growing infinitely and there are no vertical scrollbars not in browser, not Modal, not Console. But when I resize browser window and set it back to full, then vertical scrollbar appears.

ibelar:
Add this option to your modal and it should be working:

```
$mod->setOptions(['observeChanges' => true]);
```

DarkSide666:
Yeah !!! That helped.
https://www.useloom.com/share/b5951efe581f44829f69063d95096c6b

Can you add this info in documentation? Or maybe even set this as default option?
Or maybe we need method $modal->observeChanges() which will enable this switch and document it?

romaninsh:
btw console can be limited with max-height in theory.

PR #334, opened 2018-01-16
feature/Modal observes changes

ibelar:
## New method Modal::observeChanges()

Shortcut method to add observeChanges setting in semantic-ui modal.
observesChange setting allow for re positioning when DOM content changes in modal.

See example in issue: https://github.com/atk4/ui/issues/324

### DIFF
--- a/src/Modal.php
+++ b/src/Modal.php
@@ -148,6 +148,19 @@ class Modal extends View
     }
 
     /**
+     * Whether any change in modal DOM should automatically refresh cached positions.
+     * Allow modal window to add scrolling when adding content dynamically after modal creation.
+     *
+     * @return $this
+     */
+    public function observeChanges()
+    {
+        $this->setOptions(['observeChanges' => true]);
+
+        return $this;
+    }
+
+    /**
      * Add scrolling capability to modal.
      *
      * @return $this


src/Modal.php
```diff
@@ -147,6 +147,19 @@ public function setOptions($options)
...
 
+    /**
+     * Whether any change in modal DOM should automatically refresh cached positions.
+     * Allow modal window to add scrolling when adding content dynamically after modal creation.
+     *
+     * @return $this
+     */
+    public function observeChanges()
+    {
+        $this->setOptions(['observeChanges' => true]);
```
mvorisek, line 158: Hello @lubber-de, when is `observeChanges: false` helpful? Are there any negative performance/functional implications when set `observeChanges: true` by default?

the only related discussion I found: https://github.com/fomantic/Fomantic-UI/issues/564

lubber-de: observeChanges is also helpful if your code possibly changes input fields (adding/removing) to make sure focus trap will remain working. https://github.com/fomantic/Fomantic-UI/pull/2036
If setting this to true by default might has a  signifant performance impact depends on how complex your code/Dom is.
https://stackoverflow.com/questions/31659567/performance-of-mutationobserver-to-detect-nodes-in-entire-dom

PR #335, opened 2018-01-17
feature/Autocomplete-with-dropdown-settings

ibelar:
## Add ability to customize semantic-ui dropdown module for Autocomplete.

You can now pass your own semantic-ui dropdown module settings in AutoComplete.

```
$ac->setDropdownSettings($settings_array);
```

### Let say you need Autocomplete to submit form when a value has changed. You can do so by passing your own onChange event handler in the dropdown module.

```
$ac = $form->addField('country3', [
    'AutoComplete',
    'model'       => new Country($db),
    'placeholder' => 'Search for country by code, LV or UK',
    'search'      => ['name', 'iso', 'iso3'],
]);

$ac->setDropdownSettings( [
                            'allowReselection' => true,
                            'selectOnKeydown' => false,
                            'onChange'        => new atk4\ui\jsExpression('function(value,t,c){
                                                  if ($(this).data("value") !== value) {
                                                    $(this).parents(".form").form("submit");
                                                    $(this).data("value", value);
                                                  }
                                                }')
                          ]);

$form->onSubmit(function ($f) {
    return new atk4\ui\jsNotify($f->model['country3']);
});
```


src/FormField/AutoComplete.php
```diff
@@ -157,4 +156,59 @@ public function getInput()
...
+     *                                                            $(this).data("value", value);
+     *                                                          }
+     *                                                         }'),
+     *                          ]);
+     *
+     * @param array $settings
+     *
+     * @return $this
+     */
+    public function setDropdownSettings($settings)
```
romaninsh: we don't generally use setters in Agile UI unless they do more than assign a variable, so lets follow that pattern.

src/FormField/AutoComplete.php
```diff
@@ -157,4 +156,59 @@ public function getInput()
...
+    }
+
+    /**
+     * Set Semantic-ui Api settings to use with dropdown.
+     *
+     * @param array $config
+     *
+     * @return $this
+     */
+    public function setApiConfig($config)
```
romaninsh: same thing as before.

src/FormField/AutoComplete.php
```diff
@@ -157,4 +156,59 @@ public function getInput()
...
+    }
+
+    public function renderView()
+    {
+        $chain = new jQuery('#'.$this->name.'-ac');
+
+        $settings = array_merge([
+            'fields'      => ['name' => 'name', 'value' => 'id'/*, 'text' => 'description'*/],
+            'apiSettings' => array_merge($this->apiConfig, ['url' => $this->getCallbackURL().'&q={query}']),
+        ], $this->dropdownSettings);
```
romaninsh: my proposal was to move this line into a function and call a function, so if user would want to do something differently he could technically override anything. This way it's still not so flexible as you can't modify some things. 

Also overal implementation became more complex.

ibelar: You should be able to override everything using array_merge no? Not sure to exactly understand what you mean by adding a method and how to render the call to $('#ac_id').dropdown([]) using such method, other then in renderView. Can you provide a sample?

romaninsh: well for example i can't omit 'apiSettings'. I can probably set it to false. OOP-approach is more flexible.

romaninsh:
Added necessary changes and updated doc.

PR #343, opened 2018-01-22
fix tab callback

romaninsh:
Just fixes tab callback, currently in includes layout too, but must cut page content only.


romaninsh:
merging as this is a hotfix.

PR #345, opened 2018-01-22
Feature/api redirect

romaninsh:
Implementation of App::redirect and App::jsRedirect.

It's exactly what you think it is. Just wrapper around header: Location and document.location=. 

This is helpful for our new users.


romaninsh:
messed this up and fix went into readme-update 😬

PR #353, opened 2018-01-30
Feature/add dbconnect

romaninsh:
Implementation of `app::dbConnect()`. While this method has a sentimental value to ATK4 legacy users, it also has some some reasons to exist:

 - UI depends on Data anyway, so might as well introduce a method for it.
 - Most of the docs use $app->db = blah, but this property is not defined.
 - this method also does $app->add($db) meaning $db->app property is set.

This should be THE DEFAULT way now to connect to database from Agile UI, so if there is any
documentation, please update.

Workaround: `$app->add($app->db)`.

### DIFF
--- a/docs/app.rst
+++ b/docs/app.rst
@@ -243,6 +243,18 @@ so well.  Example::
 
 No much magic in these methods.
 
+Database Connection
+-------------------
+
+.. php:method:: dbConnect(dsn)
+
+This method should be used instead of manually calling Persistence::connect. This will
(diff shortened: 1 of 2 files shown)


docs/app.rst
```diff
@@ -243,6 +243,18 @@ so well.  Example::
 
 No much magic in these methods.
 
+Database Connection
+-------------------
+
+.. php:method:: dbConnect(dsn)
```
DarkSide666: you lost $user, $password and $args parameters here.

PR #354, opened 2018-01-30
Feature/various fixes

### LINKED ISSUE #363: CRUD does not work when placed inside Dynamic Tab
Some trouble with callbacks prevent it from working:

``` php
$t = $app->add('Tabs');
$t->addTab('first')->add(['Message', 'Go to next tab and click add']);

$t->addTab('second', function($p) {
    $p->add('CRUD')->setModel(new Client($p->app->db));
});
```

When modal opens, it displays same CRUD inside:

screen shot 2018-02-08 at 00 50 04

I think the problem is if there are multiple "cut" arguments, the top-most takes precedence, even though the bottom-one should do.

romaninsh:
A lot changes and fixes, mainly was required for https://github.com/atk4/mastercrud

 - Implemented BreadCrumb view with documentation and demo
 - refactored CRUD, now much more consistent, more demos comments
 - Added documentation to CRUD, after refactoring some undocumented properties
 - Console::set() second argument can override trigger event
 - Input::addAction() also adds action class (looks nicer)
 - Grid Quicksearch is now sticky
 - injecting 'quickSearch' into Grid will initialize quickSearch after you set model
 - Table::getColumnDecorators for easier access to $table->column
 - Status column now respects class
 - VirtualPages can now be nested properly: Fix #363 
 - refactored jsNotfy

### DIFF
--- a/demos/notify2.php
+++ b/demos/notify2.php
@@ -57,8 +57,6 @@ $form->onSubmit(function ($f) {
              ->setTransition($f->model['transition'])
              ->setIcon($f->model['icon']);
 
-    var_dump($notifier->options);
-
     if ($f->model['attach'] !== 'Body') {
         $notifier->attachTo($f);
     }
--- a/demos/notify2.php
+++ b/demos/notify2.php
@@ -57,6 +57,8 @@ $form->onSubmit(function ($f) {
              ->setTransition($f->model['transition'])
              ->setIcon($f->model['icon']);
 
+    var_dump($notifier->options);
+
     if ($f->model['attach'] !== 'Body') {
         $notifier->attachTo($f);
     }
--- a/docs/components.rst
+++ b/docs/components.rst
@@ -41,11 +41,11 @@ control over even the small elements.
     text
     loremipsum
     header
+    beadcrumb
     icon
     image
     message
     tabs
-    wizard
     helloworld
 
 
--- a/docs/crud.rst
+++ b/docs/crud.rst
@@ -108,7 +108,7 @@ Custom Page
 .. php:attr:: pageUpdate
 
 You can also specify a custom class for your Page. Normally it's a :php:class:`VirtualPage` but you
-can extend it to introduce your own style or add more components that just a form::
+can extend it to introduce your own style or add more components than just a form::
 
     class TwoPanels extends \atk4\ui\VirtualPage {
 
@@ -140,7 +140,7 @@ Notification
 .. php:attr:: notifyCreate
 .. php:attr:: notifyUpdate
 
-When data is saved, property `$notify` can contain a custom notification action. By default it uses :php:class:`jsNotify`
+When data is saved, properties `$notifyDefault` can contain a custom notification action. By default it uses :php:class:`jsNotify`
 which will display green strip on top of the page. You can either override it or add additional actions::
 
     $crud=$this->add([
--- a/docs/crud.rst
+++ b/docs/crud.rst
@@ -1,0 +1,156 @@
+
+.. _crud:
+
+====
+CRUD
+====
+
+.. php:namespace:: atk4\ui
+.. php:class:: CRUD
+
+CRUD class offers a very usable extension to :php:class:`Grid` class, which automatically adds actions for deleting,
+updating and adding records as well as linking them with corresponding Model actions.
+
+.. important:: If you only wish to display a non-interractive table use :php:class:`Table` class. If you need to
+    display Data Grid with some custom actions (not update/delete/add) or if you want to use your own editing
+    mechanism (such as edit data on separate page, not inside a modal), use :php:class:`Grid`
+
+
+.. important:: ATK Addon - MasterCRUD implements a higher-level multi-model management solution, that takes
+    advantage of model relations and traversal to create multiple levels of CRUDs: https://github.com/atk4/mastercrud
+
+Using CRUD
+==========
+
+The basic usage of CRUD is::
+
+    $app->add('CRUD')->setModel(new Country($app->db));
+
+Users are now able to fully interract with the table. There are ways to restrict which "rows" and which "columns" user
+can access. First we can only allow user to read, manage and delete only countries that are part of European Union::
+
+    $eu_countries = new Country($app->db);
+    $eu_countries->addCondition('is_eu', true);
+
+    $app->add('CRUD')->setModel($eu_countries);
+
+After that column `is_eu` will not be editable to the user anymore as it will be marked `system` by `addCondition`.
+
+You can also specify which columns you would like to see on the grid::
+
+    $crud->setModel($eu_countries, ['name']);
+
+This restriction will apply to both viewing and editing, but you can fine-tune that by specifying one of many
+parameters to CRUD.
+
+Disabling Actions
+=================
+
+.. php:attr: $canCreate
+.. php:attr: $canUpdate
+.. php:attr: $canDelete
+
+By default CRUD allows all the four operations - reading, creating, updating and deleting. CRUD cannot function
+without read operation, but the other operations can be explicitly disabled::
+
+    $app->add(['CRUD', 'canDelete'=>false]);
+
+Specifying Fields
+=================
+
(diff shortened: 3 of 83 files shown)


src/Console.php
```diff
@@ -59,7 +59,7 @@ class Console extends View implements \Psr\Log\LoggerInterface
      *
      * While inside a callback you may execute runCommand or setModel multiple times.
      */
-    public function set($callback = null, $junk = null)
+    public function set($callback = null, $event = true)
```
DarkSide666: Please add description of arguments with @param

src/Table.php
```diff
@@ -204,6 +204,16 @@ public function addDecorator($name, $decorator)
         $this->columns[$name][] = $decorator;
     }
 
+    public function getColumnDecorators($name)
```
DarkSide666: Please add this method in documentation and add comments and @param

docs/components.rst
```diff
@@ -41,11 +41,11 @@ control over even the small elements.
     text
     loremipsum
     header
+    beadcrumb
     icon
     image
     message
     tabs
-    wizard
```
romaninsh: duplicate..

docs/table.rst
```diff
@@ -394,11 +411,11 @@ to make a note on how template caching works then,
  - values are encapsulated for named fields.
  - values are concatinated by anonymous fields.
  - <td> properties are stacked
- - last formatter will convert array with td properties into an actual tag.
+ - last decorator will convert array with td properties into an actual tag.
 
 Header and Footer
 -----------------
-When using with multiple formatters, the last formatter gets to render Header column.
+When using with multiple decorators, the last decorator gets to render Header column.
```
DarkSide666: Really last decorator is responsible of header cell? I thought first decorator do that and last decorator maybe formats footer cell ... Please check this - sounds strange.

src/CRUD.php
```diff
@@ -12,12 +12,12 @@ class CRUD extends Grid
     public $formDefault = null;
     public $fieldsDefault = null;
 
-    public $fieldsGrid = null;
+    public $fieldsRead = null;
```
DarkSide666: Please add appropriate @var comments for all these properties.

src/CRUD.php
```diff
@@ -12,12 +12,12 @@ class CRUD extends Grid
...
 
-    public $columnEdit = null;
-    public $fieldsEdit = null;
-    public $formEdit = null;
-    public $pageEdit = null;
+    public $columnUpdate = null;
+    public $fieldsUpdate = null;
+    public $formUpdate = null;
```
DarkSide666: I think to make code below simpler all defaults can be set here like this:
```
/** @var mixed Form to use when updating record */
public $formUpdate = ['Form', 'layout' => 'FormLayout/Columns'];

/** @var mixed Page to use when updating record */
public $pageUpdate = 'VirtualPage';
// or
public $pageUpdate = ['VirtualPage', 'short_name'=>'edit']; // not sure if short_name should be here or it should be unchangeable and hard-coded somewhere below in code

and so on ...
```

This way it still will be extendable/overwritable, but all such hard-coded "configuration" will be in the beginning of class code so it's easier to understand what's going on here and what property i have to overwrite to get my desired result.

src/Grid.php
```diff
@@ -131,8 +131,7 @@ public function addQuickSearch($fields = [])
         $this->quickSearch = $form->add(new \atk4\ui\FormField\Input(['placeholder' => 'Search', 'short_name' => $this->name.'_q', 'icon' => 'search']))
             ->addClass('transparent');
 
-        if (isset($_GET[$this->name.'_q'])) {
-            $q = $_GET[$this->name.'_q'];
+        if ($q = $this->stickyGet($this->name.'_q')) {
```
DarkSide666: probably better use `$this->quickSearch->short_name` here as parameter.

src/Grid.php
```diff
@@ -196,6 +195,10 @@ public function setModel(\atk4\data\Model $model, $columns = null)
             $this->applySort();
         }
 
+        if ($this->quickSearch && is_array($this->quickSearch)) {
```
DarkSide666: so you can define fields for quicksearch as parameter while creating Grid and then when you setModel, then it will be converted to actual QuickSearch object? Well .... yeah, why not. :)

src/CRUD.php
romaninsh: Agreed.

src/Grid.php
romaninsh: yes, although it won't work with JsSearch in #368.

demos/notify2.php
```diff
@@ -57,6 +57,8 @@ public function init()
              ->setTransition($f->model['transition'])
              ->setIcon($f->model['icon']);
 
+    var_dump($notifier->options);
```
DarkSide666: var_dump is left here !

src/CRUD.php
```diff
@@ -84,52 +103,129 @@ public function setModel(\atk4\data\Model $m, $defaultFields = null)
...
+            if (!is_object($this->formCreate) || !$this->formCreate->_initialized) {
+                $this->formCreate = $this->pageCreate->add($this->formCreate ?: $this->formDefault);
+            }
+
+           // $form = $page->add($this->formCreate ?: ['Form', 'layout' => 'FormLayout/Columns']);
+            $this->formCreate->setModel($this->model, $this->fieldsCreate ?: $this->fieldsDefault);
+
+            // set save handler with reload trigger
+            $this->formCreate->onSubmit(function ($form) {
+                $this->formCreate->model->save();
```
DarkSide666: shouldn't use `$form` here instead of `$this->formCreate` ?

src/CRUD.php
```diff
@@ -84,52 +103,129 @@ public function setModel(\atk4\data\Model $m, $defaultFields = null)
...
+            $this->model->load($this->app->stickyGet($this->name));
+
+            // Maybe developer has already created form
+            if (!is_object($this->formUpdate) || $this->formUpdate->_initialied) {
+                $this->formUpdate = $this->pageUpdate->add($this->formUpdate ?: $this->formDefault);
+            }
+
+            $this->formUpdate->setModel($this->model, $this->fieldsUpdate ?: $this->fieldsDefault );
+            $this->formUpdate->onSubmit(function ($form) {
+                $this->formUpdate->model->save();
```
DarkSide666: shouldn't use `$form` here instead of `$this->formUpdate` ?

src/jsNotify.php
```diff
@@ -45,8 +33,9 @@ public function setOptions($options)
      *
      * @return $this
      */
-    public function setOption($option, $value)
+    public function setMissingProperty($option, $value)
     {
+        echo 'here';exit;
```
DarkSide666: what's this ?

romaninsh:
I have no clue why TravisCI is failing and not able to find "numcode" field. It works fine on local (mac). Can someone help?

docs/crud.rst
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    $crud=$this->add([
+        'CRUD',
+        'formDefault'=>new UserForm();
+    ])->setModel($big_model);
+
+
+.. todo:: add example / test implementation
```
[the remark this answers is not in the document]
romaninsh: now on crud page.

docs/crud.rst
```diff
@@ -0,0 +1,156 @@
...
+    $crud=$this->add([
+        'CRUD',
+        'pageDefault'=>new TwoPanels();
+    ])->setModel(new Client($app->db));
+
+
+Notification
+============
+
+.. php:attr:: notify
```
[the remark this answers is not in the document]
romaninsh: done.

src/CRUD.php
romaninsh: Technically it's the same, but ill change.

DarkSide666:
Fixed few bugs and that bug with `numcode` too.
Phpunit tests now run successfully, but only feature tests fail in PHP 7.1 (https://travis-ci.org/atk4/ui/builds/348432245)

PR #360, opened 2018-02-05
Fix/#356 Upload Field Set

ibelar:
## Allow to set field value on page load.

$field->set(‘name’, ‘token’) 

Enable setting field name and token to specify pre existing file name and/or token.

```
   /**
     * Allow to set file id and file name
     *  - fileId will be the file id sent with onDelete callback.
     *  - fileName is the field value display to user.
     *
     * @param string      $fileId   // Field id for onDelete Callback.
     * @param string|null $fileName // Field name display to user.
     * @param mixed       $junk
     *
     * @return $this|void
     */
    public function set($fileId, $fileName = null, $junk = null)
    {
        $this->setFileId($fileId);

        if (!$fileName) {
            $fileName = $fileId;
        }

        return $this->setInput($fileName, $junk);
    }
```
## Enable to set thumbnail src for pre existing field.

```
$img->setThumbnailSrc('./images/logo.png');
```

### DIFF
--- a/src/FormField/Upload.php
+++ b/src/FormField/Upload.php
@@ -7,8 +7,7 @@ use atk4\ui\Template;
 use atk4\ui\View;
 
 /**
- * Class Upload
- *
+ * Class Upload.
  */
 class Upload extends Input
 {
--- a/src/FormField/Upload.php
+++ b/src/FormField/Upload.php
@@ -209,7 +209,7 @@ class Upload extends Input
 
                         return $this->jsActions;
                     });
-                } elseif ($action === null || $files['file']['error']) {
+                } elseif ($action === null || isset($files['file']['error'])) {
                     $this->cb->set(function () use ($fx, $files) {
                         return call_user_func($fx, 'error');
                     });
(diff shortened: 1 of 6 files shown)


romaninsh:
does this also allow me to set "filename" to be shown to user?

romaninsh:
ok, reviewed your comment, looks like some mixup.

"field value" should be submitted back to form and only display to the user if "original filename" is not provided.

so set should be like this:

set($fileId, $original_filename = null)

src/FormField/Upload.php
```diff
@@ -87,6 +87,26 @@ public function init()
...
+     *
+     * @param mixed|null  $value  // Field value, display to user.
+     * @param string|null $fileId // Field id.
+     *
+     * @return $this|void
+     */
+    public function set($value, $fileId = null)
+    {
+        if ($fileId) {
+            $this->setFileId($fileId);
```
romaninsh: well i think there is still inconsistency throughout here. I'll give analogy with form fields:

$form->addField('name');

Now - form appears as "Name" if caption is not specified. However name is mandatory and must be always set. 


In your case, it would be understandable if the 'fileId' would be the primary identifier and the only required parameter, not only here, everywhere. "original_filename" is an optional parameter which can be passed from the onUpload() but it would be insignificant and wouldn't be stored anywhere.  Maybe in the future we also add "icon" or "size" parameters which are also insignificant and are not really needed at form submission.


In the `demo` you can use this callback:

```
move_uploaded_file([..], 'files/'.$original_filename);
return $original_filename;
```

so in this case $original_filename acts as ID and is passed to the form which can, for example, do some more stuff with the file.


Probably a refactoring is needed, which we can do in the future, but it's important that we swap the arguments to set() here.

romaninsh:
Declaration of atk4\ui\FormField\Upload::set($value, $fileId = NULL) should be compatible with atk4\ui\FormField\Generic::set($value = NULL, $junk = NULL)

romaninsh:
looks better, going to test.....

Issue #270, opened 2017-11-17
CRUD should focus first input field when add/edit dialog is open


romaninsh:
makes entry easier. Dialog pops up, enter data, press enter, dialog closes. 

if this is implemented as enhancement to form/modal, then the approach should be documented for other uses.

gartner:
In addition to focus on first element when opened, I think that the 'Save'-button should be included in the tab-order.
I like to use <tab>-key to walk through the form, but here, when I tab out of the last input-field in the dialog, focus is sent to the browsers url-bar. (Using Chrome).

I would like it to be placed on the 'Save'button, where I can then press <space> to save the record.

acicovic:
This is already implemented so I think the issue should be closed.

DarkSide666:
agreed, works

PR #368, opened 2018-02-09
Fix/#364 #270

ibelar:
Will autofocus on first field in modal using form.

### Change on ModalService

On dynamic modal, like when use in CRUD, the onShow and onVisible callback are fire prior modal content is loaded.
Therefore semantic-ui modal autofocus has no effects. 

To fix this, we need to manually trigger our autofocus handler function (doAutoFocus) after content is loaded in modal.

The doAutoFocus function is call when modal autofocus is set to true. True is also the default value for modal settings.

### Change on Form.php

To allow for save button to be focusable, the html tag for button is change from div to button. 

### Note

js package need to be rebuild.


romaninsh:
@ibelar although this looks good, it breaks the test-suite (behat) and something should be done wit that. Can you check it out? Behat is only used on 7.1 php it runs a "demos" on a web server while trying various tests against it.

Running behat is pretty trivial, but if documentation is needed, we can add it to README, please let me know.

ibelar:
ok let me take a look at test suite.

Issue #365, opened 2018-02-09
Come up with CSS for reseting quicksearch


romaninsh:
http://ui.agiletoolkit.org/demos/grid.php has a QuickSearch bar. When user types something and click on the search icon, grid reflects that. If previous search exists user would have to delete text manually if he wanted to reset search.

We need to improve usability of QuickSearch by allowing:

 - perform search without pressing enter (button)
 - reset search without touching keyboard

## Concept 1: JS action substitution

Using Action buttons show "reset" button by default after successful search. If user clicks into the field (usually with intent to change), then hide reset and show search action.

![quicksearch](https://user-images.githubusercontent.com/453929/36026638-e8419cc6-0d8f-11e8-89a8-2d97f8604467.gif)

``` php
$qs=$app->add('FormField\Input');
$act_search = $qs->add(['Button', 'icon'=>'search'], 'AfterInput')->addStyle('display', 'none');
$act_reset = $qs->add(['Button', 'icon'=>'red cancel'], 'AfterInput');
$qs->addClass('action');
$qs->set('prevsearch');

$act_search->on('click', $qs->jsInput()->val('searching'));
$act_reset->on('click', $qs->jsInput()->val('reset'));

$qs->on('focus', 'input', new \atk4\ui\jsExpression('[];[]',[$act_search->js()->show(), $act_reset->js()->hide()]));
```

## Concept 2: Corner reset

Use less-intrusive corner label to display cancel, while still providing search action:

![quicksearch2](https://user-images.githubusercontent.com/453929/36026903-ea75d4de-0d90-11e8-9c14-ee8ff1d25169.gif)

``` php
$i=$app->add('FormField\Input');
$i->addAction(['icon'=>'search']);
$label = $i->add(['Label','', 'left red corner'], 'AfterInput')->addStyle('cursor', 'pointer');
$label->add(['Icon','cancel']);
$i->addClass('left corner labeled');
$i->on('click', '.label', $label->js()->hide());
$i->on('click', '.label', $i->js()->removeClass('left corner labeled'));
$i->on('click', '.label', $i->jsInput()->val('')->focus());
$i->set('prevsearch');
```

# Problem:

Neither concept actually works when placed in Grid. Input field is located inside Menu with class "transparent" to avoid extra spacing. Labels appear miss-aligned:

[...]

PR #371, opened 2018-02-10
Feature/jsSearch-update

ibelar:
## Add jsSearch input field for grid.

This is an attempt to fix issue #365 and another alternative to existing QuickSearch for grid.

A new input field that will reload grid dynamically using semantic-ui api. 

You can test using demos/jssearch.php

### Field feature

- reload filtered grid on pressing enter;
- using escape, when in input field, will clear field and reload grid without filter data;
- display loading indicator when loading data;
- display button erase icon when data is filtered;
- display button search icon when not filtered;
- button is also clickable;
- keep sorting value when filtering data;

## Grid modification

Grid had to be modified to allow reloading of Grid and Paginator only, ie without reloading menu bar. A new container view that contain Grid and paginator is added, so the jsSearch action will reload container view.

The grid is backward compatible with the old QuickSearch

## Fix

This PR also fix paginator when data is filtered in new jsSearch or existing QuickSearch.

## New js plugin

This pr use a new js plugin atkJsSearch that perform user action on new jsSearch input field.

## Note

atkjs need to be rebuild. 
Atn: this PR does not included fix in modal view. Better to merge this PR after Modal input focus fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,15 @@ For detailed explanation read this article: https://medium.com/@romaninsh/atk-ne
 - Implementing Console [\#280](https://github.com/atk4/ui/pull/280) ([romaninsh](https://github.com/romaninsh))
 - Improvements for Tables, CRUDS and few other areas. [\#277](https://github.com/atk4/ui/pull/277) ([romaninsh](https://github.com/romaninsh))
 
+### 1.4.1
+
+Added `$app->dbConnect()` as a simpler way to connect to the database. Some improvements in JS libraries.
+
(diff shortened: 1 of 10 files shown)


ibelar:
Made some changes to better indicate ui state. 
- now show filter icon when filtering is on.
- changing search field input value will now reset search button but filter icon will remain on until clear.

ibelar:
just improve ui
screen shot 2018-02-14 at 12 17 12 pm
screen shot 2018-02-14 at 12 17 00 pm

romaninsh:
This looks slightly better, and works better.

PR #372, opened 2018-02-13
feature/jsUrl

ibelar:
## Add new method App::jsUrl()

This method is add in order to ease integration with other frameworks, like WP, when generating url. This will allow to differentiate between an ajax request Callbalck or regular Callback for generating proper url.

### DIFF
--- a/src/View.php
+++ b/src/View.php
@@ -1017,6 +1017,20 @@ class View implements jsExpressionable
     public $_triggerBy = null;
 
     /**
+     * Build an URL which this view can use for js call-backs. It should
+     * be guaranteed that requesting returned URL would at some point call
+     * $this->init().
+     *
+     * @param array $page
+     *
+     * @return string
+     */
+    public function jsUrl($page = [])
+    {
+        return $this->app->jsUrl($page, false, array_merge($this->_getStickyArgs($this->name), $this->stickyArgs));
+    }
+
+    /**
      * Build an URL which this view can use for call-backs. It should
      * be guaranteed that requesting returned URL would at some point call
      * $this->init().
--- a/src/View.php
+++ b/src/View.php
@@ -1016,19 +1016,8 @@ class View implements jsExpressionable
 
     public $_triggerBy = null;
 
-    /**
-     * Build an URL which this view can use for js call-backs. It should
-     * be guaranteed that requesting returned URL would at some point call
-     * $this->init().
-     *
-     * @param array $page
-     *
-     * @return string
-     */
-    public function jsUrl($page = [])
-    {
-        return $this->app->jsUrl($page, false, array_merge($this->_getStickyArgs($this->name), $this->stickyArgs));
-    }
+    /** @var bool Whether url need to be set using ajax */
+    public $needAjax = false;
 
     /**
      * Build an URL which this view can use for call-backs. It should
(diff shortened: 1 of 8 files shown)


src/View.php
```diff
@@ -1016,6 +1016,9 @@ public function getJS($force_echo = false)
 
     public $_triggerBy = null;
 
+    /** @var bool Whether url need to be set using ajax */
+    public $needAjax = false;
```
romaninsh: I don't like this approach. Some URLs need to be ajax and some does not.

src/View.php
```diff
@@ -1027,7 +1030,7 @@ public function getJS($force_echo = false)
      */
     public function url($page = [])
     {
-        return $this->app->url($page, false, array_merge($this->_getStickyArgs($this->name), $this->stickyArgs));
+        return $this->app->url($page, false, array_merge($this->_getStickyArgs($this->name), $this->stickyArgs), $this->needAjax);
```
romaninsh: i would prefer to have $view->jsUrl() also.

src/jsCallback.php
```diff
@@ -127,4 +127,18 @@ public function getAjaxec($response, $chain = null)
...
+     *
+     * @param string $mode
+     *
+     * @return string
+     */
+    public function getURL($mode = 'callback')
+    {
+        $this->owner->needAjax = true;
+
+        return parent::getURL($mode);
```
romaninsh: can call $this->jsURL() directly

ibelar:
@romaninsh - I have add jsUrl() method on my initial commit but think that it might just get more complicated than needed. I think simply adding a view property and act accordingly might be easier. 
You can see my initial commit with jsUrl():
https://github.com/atk4/ui/pull/372/commits/5f5b412ebff8a9991a6cc83f252a589d197b2049

And then, I add to add this hack for jsModal to work with jsUrl():
https://github.com/atk4/ui/pull/372/commits/f820eba2e479034f37c54adda0d6ea46922b0469

Also realized that if you approve PR#371 for jsSearch, we will need to change how the uri is generated as opposed to just setting the property to make it work with ajax. 

And what if a certain view later on need to generated url for ajax in certain circonstance but not always, I believe setting the flag on the fly might be easier then redefining $view->url() all the time.
 
```
$panel->add([new View(), 'needAjax'=>true']);
```
Let me know which one you prefer, I can easily switch it back to jsUrl()

romaninsh:
I think that any View should be capable of creating both AJAX and non AJAX requests if it wants. That weird reliance on parent is awkward. 

I like your initial commit, but I'll need to analyze what kind of problem you have faced there.

romaninsh:
@ibelar what is the status of this ticket? Is this doable at all?

ibelar:
This is working fine in WP as it is and it would be great to have for url generation in WP. Waiting for your investigation on [f820eba](https://github.com/atk4/ui/commit/f820eba2e479034f37c54adda0d6ea46922b0469) when using jsUrl() method as opposed to using a view property: needAjax.

romaninsh:
ok, how about if you create a new branch off that commit and create a separate PR? then I'll be able to test it and optimize..

ibelar:
Ok - see new [PR#382](https://github.com/atk4/ui/pull/382) 
Everything work fine in WP using PR#382 also. 
It's just that calling for url in jsModal look a bit strange. May be there is better way....

PR #380, opened 2018-02-22
Fix/#379

ibelar:
Revert button element to « div » instead of « button » and add ability for
button to use focus on tab and submit form via js.


romaninsh:
Thanks!

PR #382, opened 2018-02-23
Feature/jsUrl Rev

### LINKED ISSUE #369: app method url() should have an argument indicating a regular or ajax url
Some integrations (like ibelar/atk-wordpress) need to differentiate between ajax and regular requests when creating a url. MasterCrud does a lot with regular requests. (quickSearch as well I believe)

I suggest the url interface to be:
```php
/**
     * @param bool $ajax
     * @return string
     */
    public function url($ajax = true)
```
@ibelar, do you agree?

ibelar:
Revision using App:jsUrl() without view property.

resolves #369

### DIFF
--- a/src/Tabs.php
+++ b/src/Tabs.php
@@ -30,7 +30,7 @@ class Tabs extends View
         if ($callback) {
             // if there is callback action, then use VirtualPage
             $vp = $sub->add('VirtualPage');
-            $item->setPath($vp->getUrl('cut'));
+            $item->setPath($vp->getJSURL('cut'));
 
             $vp->set($callback);
 
--- a/src/Tabs.php
+++ b/src/Tabs.php
@@ -72,11 +72,9 @@ class Tabs extends View
             $tab = new Tab($name);
         }
 
-        $item = $this->add([$tab, 'class' => ['item']], 'Menu');
-        $item->setElement('a');
-        $item->setAttr('data-tab', $tab->name);
-
-        return $item;
+        return $this->add([$tab, 'class' => ['item']], 'Menu')
+                ->setElement('a')
+                ->setAttr('data-tab', $tab->name);
     }
 
     /**
@@ -90,10 +88,7 @@ class Tabs extends View
      */
     private function addSubView($name)
     {
-        $sub = $this->add(['View', 'class' => ['ui tab']], 'Tabs');
-        $sub->setAttr('data-tab', $name);
-
-        return $sub;
+        return $this->add(['View', 'class' => ['ui tab']], 'Tabs')->setAttr('data-tab', $name);
     }
 
     /**
--- a/src/Tabs.php
+++ b/src/Tabs.php
@@ -41,11 +41,11 @@ class Tabs extends View
     }
 
     /**
-     * Adds dynamic tab in tabs widget.
-     * Dynamic tabs are loaded via a virtual page callback or url.
+     * Adds dynamic tab in tabs widget which will load a separate
+     * page/url when activated.
      *
      * @param mixed $name      Name of tab or Tab object
-     * @param url   $needJsURL Supply URL of another page which will open in the tab
+     * @param string|array $url    URL to open inside a tab
      *
      * @throws Exception
      */
--- a/src/Tabs.php
+++ b/src/Tabs.php
@@ -53,6 +53,8 @@ class Tabs extends View
             //# TODO: refactor this ugly hack
             $item->setPath(str_replace('.php.', '.', ($needJsURL) ? $this->jsURL($action) : $this->url($action)).'#');
         }
+
+        return $sub;
     }
 
     /**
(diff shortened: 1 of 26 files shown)


ibelar:
@romaninsh - Here is another example of using jsUrl() method vs needAjax property.

When using Tab with VirtualPage in WP, generating url would need to be done via jsUrl() in order to work properly in WP. Tabs::AddTab() method is easy to update for VirtualPage to generate jsUrl() when using needAjax property. See commit [5bcd1b2](https://github.com/atk4/ui/pull/372/commits/5bcd1b270dab57ad9abdb2fd0f38519c3d5c5f4f) in atk4\ui#372.

Not sure how to do so using jsUrl() without altering VirtualPage. Let me know your thoughts.

ibelar:
@romaninsh - Never mind my last (delete) comments. 

Manage to make it work by using a new property on Callback ($needJsUrl)  instead of View. 
Now, callback will call proper owner url or jsUrl method according to this new property. 
This way, Tabs and jsModal are easy to modify and a lot clearer while jsCallback just need a property value change to generate jsUrl.

Hope this is ok. Let me know your thoughts.

src/Tabs.php
```diff
@@ -15,12 +15,13 @@ class Tabs extends View
...
-     * @param mixed $name   Name of tab or Tab object
-     * @param mixed $action Optional callback action or URL (or array with url + parameters)
+     * @param mixed $name      Name of tab or Tab object
+     * @param mixed $action    Optional callback action or URL (or array with url + parameters)
+     * @param bool  $needJsUrl Whether the virtual page should generate a jsUrl or not.
      *
      * @return View
      */
-    public function addTab($name = null, $action = null)
+    public function addTab($name = null, $action = null, $needJsUrl = false)
```
romaninsh: how do i know if virtual page should generate jsURL or not?

romaninsh:
Yes, this looks much cleaner. I still don't understand why i need to pass extra argument to addTab though, can you clarify?

romaninsh:
OH one note - we need a paragraph in the docs regarding jsURL. What's the policy on the uppercase ? should we use jsURL? I think we uppercase acronyms entirely (getURL) unless they are in the beginning (dbConnect).

ibelar:
In regards to extra argument for addTab. When using a regular url to load tab in WP, this will output the entire page instead of the tab portion needed, so it must use an ajax request which is route via jsUrl. It's really needed for WP to output tab dynamically.

I will update PR to us jsURL instead.

romaninsh:
Oh i think we shouldn't be using regular URL in tab. I know one of the tests may be using that, but I am not happy about it. It's either a callback or a virtual page in a second argument.

Perhaps adding addTabURL() is better?

ibelar:
Now good to go!

romaninsh:
@ibelar - I have added some cleanups, restored original addTab() functionality and moved url generation outside of the tab. If user does pass the array, then $app->url() is used, but he can call `addTabURL('bleh', $app->jsURL(..))`

I wonder if this implementation is still OK for wordpress admin and if it would still solve #369

ibelar:
@romaninsh - I see, understand the changes.

The changes will still work in WP but we will need to define VirtualPage prior to set Tab and use VirtualPage url like this:

```
        $vp = $this->add(['VirtualPage']);
        $vp->cb->needJsURL = true;
        $vp->add(['LoremIpsum', 'size' => 2]);

        // dynamic tab
        $t->addTabURL('Dynamic Lorem Ipsum', $vp->getURL('cut'));
```

as opposed to just do:

```
        // dynamic tab
        $t->addTabURL('Dynamic Lorem Ipsum', function ($tab) {
            $tab->add(['LoremIpsum', 'size' => 2]);
        }, true);
```

I can also add WpTabs, extending Tabs, in my integration that will take care of this so we can still use it as before.

```
        $t = $this->add('WpTabs');
        // dynamic tab
        $t->addWpTab('Dynamic Lorem Ipsum', function ($tab) {
            $tab->add(['LoremIpsum', 'size' => 2]);
        }, true);
```

Should still solve #369

romaninsh:
Can you explain WHEN VirtualPage does not need "jsURL" ? 

Can we set Callback::needJsURL = true permanently ?

ibelar:
Wizard callback need a regular url for next and previous step buttons. Probably because it does not render json but html only. When pressing next using a link generate via jsUrl, it will only output the form html not the entire page.


May be other in future...

wizard-jsurl

romaninsh:
@ibelar , i have done some thinking and I'm proposing to maintain getURL and getJSURL inside callbacks and virtual pages. Here is my reasoning:

When we create Virtual page or a callback it does not matter what is inside or where it is placed. It actually matters how will we invoke said callback. There are two options:

1. load it dynamically (would need jsURL)
2. direct user to it (would need url)

So technically the same content could be available either directly or through JS. So my proposed solution this time is to update ATK to use 'getJSURL' whenever we specify callback that is used inside JavaScript.

I hope this is better approach. What do you think, @ibelar

ibelar:
Yes! Totally agree with this change. It really make more sense and believe this will also be more flexible for future use.
I have perform some basic testing in WP using the atk-wordpress-demo and atk-demo-sample plugin and all seems to work fine. I am also able to use Wizard out of the box and will be able to include it in atk-wordress-demo.  
Thanks.

PR #395, opened 2018-03-06
fix/#387-Allow a js function to run after reload is complete.

ibelar:
Will allow a js function to be run after api onSuccess.

ex: 
```
$val = $app->add(['FormField/Line', '']);
$val->addAction('Set Custom Value')->js('click', new \atk4\ui\jsReload($v, ['val' => $val->jsInput()->val()], $val->jsInput()->focus()));
```

## File Change 
 - jsReload.php
 - reloadView.js
 - apiService.js file.

Note: atkjs-ui need to be rebuild.


romaninsh:
Please investigate test-suite. Otherwise change looks OK

ibelar:
Looks like problem is from jsReload class name not using CamelCase. What should I do?

romaninsh:
that's ok.

PR #397, opened 2018-03-08
Added getField-method to Generic FormLayout

FabulousGee:
Added getField-method to Generic Form Layout

See also:
https://forum.agiletoolkit.org/t/split-existing-form-into-tab/447/4


romaninsh:
see #398

PR #407, opened 2018-03-19
feature/Draggable

ibelar:
## Description
Allow a grid table to be draggable for reordering purpose.

### Use

```
$g = $app->add(['Grid']);
$g->setModel(new Country($db));

$dragHandler = $g->addDragHandler();

$dragHandler->onReorder(function($newOrder) {
    //newOrders contains an array of grid ids;
});
```
## Note

atkjs-ui need to be rebuild.


ibelar:
@romaninsh - Did a bit of refactoring and add jsSortable.php as a jsCallback.

Think this will be more flexible this way. Can be apply to grid but to whatever container, like ul element. 
Also add new information sent to callback, i.e. the source element id and it's final position.

see /demos/jssortable.php for a complete demo.

romaninsh:
(ibelar i sent you PM on gitter, is that only way to get in touch with you directly?)

romaninsh:
merged, but not deleting the branch yet, in case some further fixes are introduced.

PR #408, opened 2018-03-21
Implement "Multiformat" column decorator for table

romaninsh:
Implementation of Multiformat column for the table. It allows to use different formatters depending on cell value. Simple use:

``` php
    $table->addColumn('amount', ['Multiformat', 'callback'=>function($a, $b) {

        if($a['is_invoiced'] > 0) {
            return ['Link', 'invoice', ['invoice_id'=>'id']];
        } elseif (abs($a['is_refunded']) < 50) {
            return ['Template', 'Amount was <b>refunded</b>'];
        }

        return ['Money'];
    }]);
```

Screenshot does not really correspond to example above, but demonstrates 3 different decorators for a same column, including Link+Money, Money and Template.

screen shot 2018-03-21 at 17 00 16

 - Documentation: docs/table.rst
 - Demo: demos/table2.php

### DIFF
--- a/demos/table2.php
+++ b/demos/table2.php
@@ -3,33 +3,70 @@
 date_default_timezone_set('UTC');
 include 'init.php';
 
+$app->add(['Header', 'Table with various headers', 'subHeader'=>'Demonstrates how you can add subheaders, footnotes and other insertions into your data table', 'icon'=>'table']);
+
 $data = [
     ['id'=>1, 'action'=>'Salary', 'amount'=>200],
     ['id'=> 2, 'action'=>'Purchase goods', 'amount'=>-120],
     ['id'=> 3, 'action'=>'Tax', 'amount'=>-40],
 ];
 
-$p = new \atk4\data\Persistence_Array($data);
-$m = new \atk4\data\Model($p);
-
-$m->addField('action');
-$m->addField('amount', ['type'=>'money']);
+$m = new \atk4\data\Model(new \atk4\data\Persistence_Static($data));
+$m->getElement('amount')->type = 'money';
 
 $table = $app->add('Table');
-$table->setModel($m);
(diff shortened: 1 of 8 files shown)


demos/table2.php
```diff
@@ -3,33 +3,70 @@
 date_default_timezone_set('UTC');
 include 'init.php';
 
+$app->add(['Header', 'Table with various headers', 'subHeader'=>'Demonstrates how you can add subheaders, footnotes and other insertions into your data table', 'icon'=>'table']);
```
DarkSide666: table2 demo page is not included in left side menu.

DarkSide666: also opening that demo page throws exception
```
Error: Class '\atk4\data\Field\Callback' not found (in C:\wamp64\www\git\atk\ui\vendor\atk4\core\src\FactoryTrait.php:153)
#0 C:\wamp64\www\git\atk\ui\vendor\atk4\data\src\Model.php(373): atk4\data\Model->factory(Array, Array, '\\atk4\\data\\Fiel...')
#1 C:\wamp64\www\git\atk\ui\demos\table2.php(49): atk4\data\Model->addField('amount_copy', Array)
#2 {main}
```
If it doesn't do that on your environment, then try running `composer update`

src/TableColumn/Multiformat.php
```diff
@@ -0,0 +1,80 @@
+<?php
```
DarkSide666: Please add at least some type-hint comments for properties and methods :)

demos/table2.php
romaninsh: fixed both.

src/TableColumn/Multiformat.php
romaninsh: adding..

PR #416, opened 2018-03-22
Feature/add card

romaninsh:
Implementation of Card view. Simple use:

``` php
$app->add('Card')->setModel($user->load(1));
```

This is intended to display data of a single record.

screen shot 2018-03-22 at 13 46 04

This implementation is not final. I want this to look more like this:

screen shot 2018-03-22 at 13 10 05

with ability to interact with the individual card, adding corner labels, totals, images, etc.

### DIFF
--- a/demos/card.php
+++ b/demos/card.php
@@ -1,0 +1,7 @@
+<?php
+include 'init.php';
+include 'database.php';
+
+$app->add(['Header', 'Card displays read-only data of a single record']);
+
+$app->add(['Card'])->setModel((new Stat($db))->tryLoadAny());
--- a/src/Card.php
+++ b/src/Card.php
@@ -40,7 +40,7 @@ class Card extends Table
         $mm = parent::setSource($data);
         $this->addDecorator('value', ['Multiformat', function ($row, $field) use ($m) {
             $field = $m->getElement($row->data['id']);
-            $ret =  $this->decoratorFactory($field);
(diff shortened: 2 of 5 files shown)


demos/card.php
```diff
@@ -0,0 +1,8 @@
+<?php
```
DarkSide666: Demo is not added in left side menu :)

src/Card.php
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Card class displays a single record data.
+ *
+ * IMPORTANT: Although the purpose of the "Card" component will remain the same, we do plan to
+ * improve implementation of a card to to use https://semantic-ui.com/views/card.html.
+ */
+class Card extends Table
+{
+    protected $_bypass = false;
```
DarkSide666: need description of this property

src/Card.php
```diff
@@ -0,0 +1,54 @@
...
+
+namespace atk4\ui;
+
+/**
+ * Card class displays a single record data.
+ *
+ * IMPORTANT: Although the purpose of the "Card" component will remain the same, we do plan to
+ * improve implementation of a card to to use https://semantic-ui.com/views/card.html.
+ */
+class Card extends Table
```
DarkSide666: Really name it just - Card? Well, maybe that's OK at least for a start.

PR #425, opened 2018-04-05
Adding virtualpage into Form makes sure we inherit stickyGet

romaninsh:
Currently using "plus" inside crud pop-up does not work due to 2 reasons:

 - virtual page is added into app, therefore loosing some stickyGet set on the modal
 - semanticUI pop-up looks messed up appears not on top of existing popup but below it.


ibelar:
@romaninsh - made minor changes when no autocomplete is not inside a form. Then virtual page will be add to owner if no form.
Also fix submit error in autocomplete.php demo file.

PR #431, opened 2018-04-10
Fix #418

### LINKED ISSUE #418: Form Dropdowns do not show ... (empty selection) once something is selected
If a Form is associated with a model, it displays hasOne References as Dropdown.

When no value for this dropdown is set, it shows an empty selection (... placeholder).
As soon as a value is selected, this option disappears. So its no possible to un-select this hasOne Reference any more.
The ... option should only disappear when the field is set to` required = true`

Empty selection available
empty_selection_there

And gone after something was selected
empty_selection_gone

PhilippGrashoff:
Fix for #418, allows empty selection in a dropdown once a "real" value is selected (was not possible before).
As there is an open bug in Semantic UI for this, this might change with future sematic UI Versions:
https://github.com/Semantic-Org/Semantic-UI/issues/2072

### DIFF
--- a/src/FormField/DropDown.php
+++ b/src/FormField/DropDown.php
@@ -18,7 +18,17 @@ class DropDown extends Input
     public function init()
     {
         parent::init();
-        $this->jsInput(true)->dropdown();
+        //See https://github.com/atk4/ui/issues/418
+        //if field is required, disable empty selection once a value is
+        //selected. Currently standard behaviour of sematic ui dropdown
+        if(isset($this->field) && $this->field->required) {
+            $this->jsInput(true)->dropdown();
+        }
+        //add any (does not have to be $this->empty) placeholder to allow
+        //empty selection even after a value was selected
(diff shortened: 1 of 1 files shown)


romaninsh:
Asking iBelar to review if this. I hope this fixes #418

ibelar:
Yes, this does fix 418

But while we are here, we should probably move js action in Dropdown::renderView() and add method to setup/override existing options for Dropdown.

```
function renderView()
{
   $this->js(true)->dropdown($this->options)
  parent::renderView();
}
```
add method Dropdown::setOptions($options);

and set $this->options['placeholder'] in init as you do now.

this will add more flexibility in future.

PhilippGrashoff:
@ibelar good idea, in AutoComplete it already is this way if I recall correctly.

romaninsh:
@ibelar I think if this solves the issue lets merge this. The further refactoring can be done later.

PR #435, opened 2018-04-12
Allow to set a custom Limit to loaded records

PhilippGrashoff:
Adds a property `limit` which can be used to set a custom limit of records loaded.
This can be very handy in the following case:
Imagine in total its 50 records for this model, 20 of them are used often, the others hardly. Setting the limit to 20 and the sorting to frequency makes the 20 most used records available without typing anything in the text field.


romaninsh:
yes, this is a good PR, thanks!

PhilippGrashoff:
@romaninsh just had the described case in real life :)

PR #446, opened 2018-04-17
Add function getValue() to Radio

### LINKED ISSUE #445: FormField Radio does not have getValue() method
Reading the value from a Radio Form field like:
`$my_precious_value = $form->getField('someradioinput)->getValue();`
does not work because Radio class nor its parents have this method.

The solutions seems pretty straightforward, will create.

PhilippGrashoff:
Fixes #445 

Please check if the code makes sense to you. To me its still a bit unclear if a form field can exist without a field set, and what the form field content property is.

### DIFF
--- a/src/FormField/Radio.php
+++ b/src/FormField/Radio.php
@@ -49,4 +49,15 @@ class Radio extends Generic
 
         return parent::renderView();
     }
+
+
+    /**
+     * Returns value if field if a field is set to this input, null else.
+     *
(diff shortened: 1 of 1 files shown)


romaninsh:
Actually getValue() should be protected everywhere else. It's only for internal use. How about we refactor this PR to properly deal with it? 

Also use $form->model['field'] or if you are working with field, then $field->get().

PR #448, opened 2018-04-18
Feature/Table-Column-Filter

ibelar:
## Add filtering to grid column using popup.

Filtering can be set by columns on a grid. Each grid’s column filter will be set according the their model field type.
The state of the filter is display via it’s icon colour.

Finally an action button is added to the grid for clearing all filter at once.

column-filter
column-filter-on

## Usage

This will add filtering to all columns in the grid.
```
$g = $view->add(['Grid']);
$g->setModel(new Country($db));

$g->addFilterColumn();
```

You may specify column by passing an array to Grid::addFilterColumn()

```
$g->addFilterColumn([‘name’, ‘iso’]);
```

This way, only ‘name’ and ‘iso’ column will have filtering.

## demo 

In demos/tablefilter.php


romaninsh:
terrible - decrease by 4% of code coverage just by adding code without tests.

romaninsh:
i'll leave branch, to add tests and create new PR.

PR #454, opened 2018-04-30
feature/async-defer

ibelar:
## JS file loading using async and defer method.

Allow javascript file to use async / defer option when loading by adding attribute tag to script tag.

## Usage

if Async is need, set second argument to true, 
if Defer is need, set third argument to true.

```
  $this->app->requireJs( $url, false, true );
```

## Note

This is require for google-address add-on.


src/App.php
```diff
@@ -555,12 +555,15 @@ public function stickyForget($name)
...
      *
      * @param string $url
+     * @parma bool   $isAsync Whether or not you want Async loading.
+     *
+     * @param bool $isDefer Whether or not you want Defer loading.
      *
      * @return $this
      */
-    public function requireJS($url)
+    public function requireJS($url, $isAsync = false, $isDefer = false)
```
romaninsh: can there be a trigger callback after loaded?

ibelar:
this can be set in url, at least with Google api:

```
$this->app->requireJs( "https://maps.googleapis.com/maps/api/js?key=""&callback=initMap", false, true );
```

PR #455, opened 2018-05-01
Added setApiConfig() function to Form.php

PhilippGrashoff:
Using setApiConfig(), additional parameters can be passed to Semantic UI .api function which does the AJAX submission of the Form. Code is cloned from AutoComplete.php

### DIFF
--- a/docs/form.rst
+++ b/docs/form.rst
@@ -467,7 +467,8 @@ Form Submit Handling
 .. php:method:: setApiConfing($config)
 
     Add additional parameters to Semantic UI .api function which does the AJAX submission of the form.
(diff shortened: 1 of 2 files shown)


docs/form.rst
```diff
@@ -464,6 +464,11 @@ Form Submit Handling
 
     Create and return :php:class:`jsChain` action, that will replace form with a success message.
 
+.. php:method:: setApiConfing($config)
+
+    Add additional parameters to Semantic UI .api function which does the AJAX submission of the form.
+    All available parameters can be found here: https://semantic-ui.com/behaviors/api.html#/settings
+
```
romaninsh: could you add some short example here?

romaninsh:
Thank you @PhilippGrashoff for PR.

PR #458, opened 2018-05-03
Feature/Admin Menu

ibelar:
## Change Admin menu in admin layout

This improve the way admin menu behave in Admin layout. Also add ability to change layout css class for main content. 

layout-1

 - Burger icon will display on specific media query only and will allow to toggle  menu on or off when visible. Burger icon is hide when not in use.
- Menu Show or hide using transition animation;
- Ability to easily target other device screen size;
- App title is move to top admin menu. This allow to display the app title at all time, even when side menu is scroll.
- Remove dependency to agilejs.js file;

## Fix
- no need to add an extra view for table column popup or dropdown.

## Note
Need to recompile agileui.less file

layout-2
layout-3


mayack:
Hey there,

Can give some insight on why the menu's layout had to be changed in the first place? Not sure if this change works for AgileUI. Burger icon's behaviour can be implemented though.

romaninsh:
Some of the cons:

 - i think burger should be always visible. It's useful sometimes to see info on screen.
 - there should be way to hide top menu. quite often there are errors dumped and very hard to see them.
 - can't say i like new toolbar/menu positioning.

romaninsh:
crud2 is broken

footer positioning is wrong.

mobile size menu squeezes content.

@mayack said he will create a separate branch to address this.

ibelar:
@mayack - Here are the reason that motivated the changes:
- try to simplify css in order to fix popup positioning without the need of an extra view;
- ability to set different css class, like fluid, for expanding content to full width;
- unified the menu behaviour: 
  - try to unified the experience when using menu on desktop and mobile size. On desktop size, menu is open and close via burger icon, but on mobile size, the menu is open using burger icon and close using X icon; Plus, an overlay is shown in mobile size, but not on desktop size.
  - on mobile size, when you open menu and scroll up and if you need to close the menu, then you need to scroll down to get the X icon back as opposed to just click the burger icon.
- minimize javascript for menu to it's minimum;
- Application title disappear when menu is scroll up on desktop size, and show only when menu is open on mobile size. I think it is better when your application title is shown all the time.

Most important point for me I guess is to have a unified experience between mobile and desktop and that application title, specially as an application developer, should be showed at all times; but agreed that this is also personal preference.

PhilippGrashoff:
I really like the idea, and also agree with romans that burger should be always there but independent from top menue

romaninsh:
closing this in favour of #460

romaninsh:
#460

 - fixes CSS for popup positioning
 - is fluid by default. I'll add option to remove fluidity, but we think it's a good deafult.

@mayack - can you look into other replies, perhaps those can be re-integrated into #460.

PR #460, opened 2018-05-08
Feature/admin layout2

romaninsh:
Implementation of a cleaner Admin layout. Features added:

 - fix for the issue with the popups not aligning correctly in the table
 - use flexible mode by default for admin

### DIFF
--- a/demos/tablecolumnmenu.php
+++ b/demos/tablecolumnmenu.php
@@ -6,10 +6,7 @@ require 'database.php';
 
 $app->add(['Header', 'Table column may contains popup or dropdown menu.']);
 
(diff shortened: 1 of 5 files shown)


demos/tablecolumnmenu.php
```diff
@@ -7,7 +7,7 @@
 $app->add(['Header', 'Table column may contains popup or dropdown menu.']);
 
 //For popup positioning to work correctly, table need to be inside a view segment.
-$view = $app->add('View', ['ui' => 'basic segment']);
+$view = $app->add('View');
```
romaninsh: I wonder if those Views are still needed.

Issue #466, opened 2018-05-16
Conditional form rule doesn't run with model default value


DarkSide666:
Model:
```
$this->addField('type', [
    'required'  => true,
    'default'   => 'text', // conditional form doesn't pick up this value automatically
    'enum'      => ['switch', 'list', 'text', 'number', 'money', 'datetime', 'percent', 'rating'],
]);
$this->addField('text_long_text', ...);
$this->addField('text_rich_text', ...);
```

Form:
```
public $fieldsDisplayRules = [
    'text_long_text' => ['type' => 'isExactly[text]'],
    'text_rich_text' => ['type' => 'isExactly[text]'],
];
```

Open form, in dropdown value "text" is selected, but fields text_long_text and text_rich_text doesn't show up.
When I change dropdown value and select another value, then conditions are executed.
When I select value "text" again, then fields text_long_text and text_rich_text shows up.

Probably condition rules are checked only on onChange event, but with default values it should be checked also on init. Most likely solution is to also run condition rules for all fields as soon as form is initialized + on each field onChange event as it most likely is now.

ibelar:
fix in #472

PR #472, opened 2018-05-16
fix/issue#466

ibelar:
- set field rule according to their initial value. #466


romaninsh:
does this require recompile of JS ?

PR #474, opened 2018-05-17
Feature/Dropdown-Plus

ibelar:
## Dropdown input field

Refacror Dropdown form field

 - allow to pass icon in menu item;
 - allow to set multiple selection;
 - allow for dropdown option to be pass as semantic-ui dropdown options;

dropdown-plus

## Usage

Use as regular Dropdown input field except for values that can use an array as a value argument for using icon.

Here is an example of displaying icon with dropdown item value.
```
$form->addField('icon',
                ['DropDownPlus',
                    'caption' => 'Using icon',
                    'empty' => 'Choose an icon',
                    'values'  => ['tag' => ['Tag', 'icon' => 'tag icon'], 'globe' => ['Globe', 'icon' => 'globe icon'], 'registered' => ['Registered', 'icon' => 'registered icon'], 'file' => ['File', 'icon' => 'file icon']],
,
                ]);
```

## Demo

More sample available in dropdown-plus.php demos file.

## Important

atkjs-ui.js file need to be rebuild for conditionalForm to work using new dropdown,


ibelar:
ac933698e2b1680851a3f5bc94c4351041ccf450

Fix issue with conditional formatting. Now compatible with atkConditionalForm js plugin.

PR #475, opened 2018-05-22
bugfix if icon is not set

DarkSide666:
otherwise we get
`Notice: Undefined index: icon in C:\wamp64\www\saasty.io\vendor\atk4\ui\src\Menu.php on line 125`

### DIFF
--- a/src/Menu.php
+++ b/src/Menu.php
@@ -122,7 +122,7 @@ class Menu extends View
         if (is_string($title)) {
             $group->set('title', $title);
         } else {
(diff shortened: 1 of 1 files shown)


src/Menu.php
```diff
@@ -122,7 +122,7 @@ public function addGroup($title)
         if (is_string($title)) {
             $group->set('title', $title);
         } else {
-            if ($title['icon']) {
+            if (isset($title['icon'])) {
```
romaninsh: && $title['icon']

PR #482, opened 2018-06-03
Make View->on(event, [jsAction1, jsAction2]) work

PhilippGrashoff:
If an array containing only JS actions was used as second parameter of View->on, it failed and threw the error message:
atk4\ui\Exception: You cannot add anything into the view after it was rendered

Now JS actions are merged into $actions array and it works as intended.

Simple test code:
        $v_test= $app->add(['Button']);
        $v_test->on('click', [$v_test->js()->html('sdfsf'), $v_test->js()->html('ddddd')]);


romaninsh:
You are up early. Or late :)

PR #484, opened 2018-06-04
Allow $field->ui['placeholder'] to affect form fields' placeholder directly.

romaninsh:
There is a perfectly working solution to set a Field's placeholder value:

``` php
$this->addField('surname', ['ui'=>['Form'=>['placeholder'=>'e.g. Smith']]]);
```

However, it appears that defining this value directly in 'ui' is desired by users:

``` php
$this->addField('surname', ['ui'=>['placeholder'=>'e.g. Smith']]);
```

This PR implements a short-hand method:

screen shot 2018-06-04 at 10 23 12

### DIFF
--- a/demos/form2.php
+++ b/demos/form2.php
@@ -72,7 +72,7 @@ $form->onSubmit(function ($f) {
         {
             parent::init();
             $this->addField('name', ['required' => true]);
-            $this->addField('surname');
+            $this->addField('surname', ['ui'=>['placeholder'=>'e.g. Smith']]);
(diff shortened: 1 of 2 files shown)


PhilippGrashoff:
Hi, my issue was about adding a FormField to a Form, not a Data Field to a Model.
At least the Input Formfield has the property $placeholder. And as Upload extends Input, setting the placeholder property using a seed should work.

In general, Form Fields of Forms and Data Fields of Models can be mixed up easily I think :)

As for Model->addField, I think the way it is using 'ui' => 'form => [] is totally ok as its a Data Field.

PR #492, opened 2018-06-14
atk4/ui/fix/#491 auto complete not initializing with proper text

skondakov:
I am trying to load the AutoComplete value from its model and not the forms model. Also trying to show its selected title field.

### DIFF
--- a/src/FormField/AutoComplete.php
+++ b/src/FormField/AutoComplete.php
@@ -238,7 +238,7 @@ class AutoComplete extends Input
 
         $this->js(true, $chain);
 
-        if ($this->field->get()) {
+        if ($this->field && $this->field->get()) {
             $id_field = $this->id_field ?: $this->model->id_field;
             $title_field = $this->title_field ?: $this->model->title_field;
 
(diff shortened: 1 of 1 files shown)


romaninsh:
@ibelar if this looks ok to you - merge.

skondakov:
I have added a check for $this->field .. about setting to "0", I set it to this value as this seems to be the default empty value for the autocomplete. We can set it to null as well, but shouldn't leave it without changing as I have noticed that in case of multiple autocompletes - there is a bug with loading the same value all over them.

src/FormField/AutoComplete.php
```diff
@@ -238,10 +238,18 @@ public function renderView()
...
-            $chain->dropdown('set value', $data['id'])->dropdown('set text', $data[$this->model->title_field]);
-            $this->js(true, $chain);
+        if ($this->field && $this->field->get()) {
+            $id_field = $this->id_field ?: $this->model->id_field;
+            $title_field = $this->title_field ?: $this->model->title_field;
+
+            $this->model->tryLoadBy($id_field, $this->field->get());
+
+            if (!$this->model->loaded()) {
+                $this->field->set('0');
```
romaninsh: why not null?

skondakov:
Done, it is null now

PR #501, opened 2018-06-20
Fix problem with multiple callbacks being not executed

### LINKED ISSUE #500: Modal in Tab not working
When placing Modal in a Tab it's not reaching the correct callback.

screen shot 2018-06-20 at 09 26 45

When clicking on the button, I expect to see "LoremIpsum" in the popup, but the same tab repeats again.

screen shot 2018-06-20 at 09 26 51

The cause of this seems to be in the URLs:

```
http://localhost/ui/demos/tabs.php?atk_admin_tabs_view_4_virtualpage=cut&__atk_callback=1&_=1529483293534#/atk_admin_tabs_tab_4
http://localhost/ui/demos/tabs.php?atk_admin_tabs_view_4_virtualpage=cut&atk_admin_tabs_view_4_virtualpage_modal_view_callbacklater=ajax&__atk_callback=1&json=true
```

the "cut" of view 4 virtualpage overrides modal_view_callbacklater's response.

Code:

``` php
$t->addTab('Modal popup', function ($tab) {
    $tab->add(['Button', 'Load Lorem'])->on('click', $tab->add('Modal')->set(function($p){
        $p->add(['LoremIpsum', 'size' => 2]);
    })->show());
});
```

romaninsh:
Various UI element use CallbackLater. It will divert it's execution until later time, but what if it is already a "later time" ? Currently CallbackLater would divert until "more later time" beforeRender / beforeOutput, but that's not a correct approach, because it would fail third time. 

This PR will execute CallbackLater instantly if it's already late.

resolve #500 

IMPORTANT: this fix was reverted in the past due to: https://github.com/atk4/ui/issues/313. I have tried paginator examples and they seem to work fine.

### DIFF
--- a/demos/tabs.php
+++ b/demos/tabs.php
@@ -18,6 +18,13 @@ $t->addTab('Dynamic Lorem Ipsum', function ($tab) {
     $tab->add(['LoremIpsum', 'size' => 2]);
 });
 
+// modal tab
+$t->addTab('Modal popup', function ($tab) {
+    $tab->add(['Button', 'Load Lorem'])->on('click', $tab->add('Modal')->set(function($p){
+        $p->add(['LoremIpsum', 'size' => 2]);
+    })->show());
+});
+
 // dynamic tab
 $t->addTab('Dynamic Form', function ($tab) {
     $tab->add(['Message', 'It takes 2 seconds for this tab to load', 'warning']);
--- a/demos/tabs.php
+++ b/demos/tabs.php
@@ -20,7 +20,7 @@ $t->addTab('Dynamic Lorem Ipsum', function ($tab) {
 
 // modal tab
 $t->addTab('Modal popup', function ($tab) {
-    $tab->add(['Button', 'Load Lorem'])->on('click', $tab->add('Modal')->set(function($p){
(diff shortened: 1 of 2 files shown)


demos/tabs.php
```diff
@@ -18,6 +18,13 @@
     $tab->add(['LoremIpsum', 'size' => 2]);
 });
 
+// modal tab
+$t->addTab('Modal popup', function ($tab) {
+    $tab->add(['Button', 'Load Lorem'])->on('click', $tab->add('Modal')->set(function ($p) {
+        $p->add(['LoremIpsum', 'size' => 2]);
+    })->show());
+});
```
DarkSide666: demo don't work well
* try open modal popup, then close it by clicking X and then open it again - it will open modal dialog, but empty without LoremIpsum.
* also there is quite ugly flickering of modal window while it is opening (re-opening), but I think it's because of window being small while it shows loader spinner.

I found out that in cases when no LoremIpsum shows up, it's because it don't make Ajax call to server at all. So it's probably some issue with how JS executes.

romaninsh: I'll try to ask @ibelar for help..

Issue #507, opened 2018-06-25
Date picker doesn't respect APP->ui_persistence->date_format


DarkSide666:
Date picker doesn't respect APP->ui_persistence->date_format.

```
$app->ui_persistence->date_format = 'd.m.Y';
$app->ui_persistence->time_format = 'H:i:s';
$app->ui_persistence->datetime_format = 'd.m.Y H:i:s';

// Initialize model
$m = new Model_Order($p);

// CRUD
$c = $app->add(['CRUD', 'ipp' => 10]);
$c->setModel($m);
```

Form still shows date in default `M d, Y` format and as result date can't be saved - shows "Incorrectly formatted date/time"

DarkSide666:
I could not find respective option in Semantic Datetime library. There is `inputFormat`, but I could not make it work as I expected.

Also please add `firstDayOfWeek` property in UI persistence + make Datetime pick this up.

DarkSide666:
Tested. Now works fine.

PR #508, opened 2018-06-25
Feature/Calendar-input-using-ui-persistence

ibelar:
Let calendar input field use format from ui persistence.

This is a fix for #507 

This will now work:
```
$app->ui_persistence->date_format = 'd.m.Y';
$app->ui_persistence->time_format = 'H:i';
$app->ui_persistence->datetime_format = 'd.m.Y H:i';

$c = $app->add(['CRUD', 'ipp' => 10]);
$c->setModel($m);
```

## Note 

atkjs-ui need to be rebuild.

### DIFF
--- a/src/FormField/Calendar.php
+++ b/src/FormField/Calendar.php
@@ -3,9 +3,17 @@
 namespace atk4\ui\FormField;
 
 use atk4\ui\Form;
+use atk4\ui\jsExpression;
+use atk4\ui\jsFunction;
 
 /**
  * Input element for a form field.
+ *
+ * 2018-06-25 : Add Locutus js library for formatting date as per php format.
+ * http://locutus.io/php/datetime/
+ *
+ * Locutus date function are available under atk.phpDate function.
+ * ex: atk.phpDate('m.d.Y', new Date());
  */
 class Calendar extends Input
 {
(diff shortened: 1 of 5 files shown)


romaninsh:
@ibelar if not hard - could you add first day of week?

src/FormField/Calendar.php
```diff
@@ -29,10 +47,21 @@ public function renderView()
...
-            $this->options['type'] = $this->type;
+        if (!$this->type) {
+            $this->type = 'datetime';
+        }
+
+        $typeFormat = $this->type.'_format';
+        if ($format = $this->app->ui_persistence->$typeFormat) {
+            $formatter = "function(date, settings){
+                            if (!date) return;
+                            return atk.phpDate('${format}', date);
```
romaninsh: can this be passed as an argument into jsExpression?

DarkSide666: good point

ibelar:
Oups forgot about firstDayOfWeek - just added now.
@romans - also properly escaping argument in jsExpression

src/FormField/Calendar.php
```diff
@@ -55,13 +55,17 @@ public function renderView()
...
-                            return atk.phpDate('${format}', date);
+                            return atk.phpDate([format], date);
                         }";
-            $this->options['formatter'][$this->type] = new jsExpression($formatter);
+            $this->options['formatter'][$this->type] = new jsExpression($formatter, ['format' => $format]);
         }
 
         $this->options['type'] = $this->type;
 
+        if ($dayOfWeek = $this->app->ui_persistence->firstDayOfWeek) {
```
DarkSide666: while 0 (Sunday) is default for JS library then this will work, otherwise you have to validate it with `isset` or `is_int` here.

ibelar: Yes, default is 0 for Calendar options, so will start on Sunday like the actual implementation.

Issue #524, opened 2018-07-09
CRUD edit form reload don't use stickyGet parameters


DarkSide666:
Example in demo/crud.php first CRUD.

1. open demo/crud.php page.
2. sort first CRUD by phone number by clicking on column header.
3. open edit form (in this URL there still is atk4_admin_crud_sort=phonecode sticky parameter).
4. when you click Save in edit form, then it simply triggers js reload with `trigger('reload')`, which was added in CRUD view on init() phase and don't contain stickyGet arguments in it's URL. So CRUD reloads without taking current stickyGet arguments in account.

CRUD on init generates this js code and at that moment there was no stickyGet sort argument:
```
$("#atk_admin_crud").on("reload",function(event) {
  event.preventDefault();
  event.stopPropagation();
  $("#atk_admin_crud").atkReloadView({"uri":"crud.php?__atk_reload=atk_admin_crud","uri_options":[],"afterSuccess":null});
  });
```

Later on when you call `trigger('reload')` it simply runs this js reload method and reloads view without sort argument and that  way _forgets_ it.

Possible solutions:
1. add ability to call `trigger('reload', $this->stickyArgs)` and that would inject arguments in original reload URL
2. or move on('reload' , ...) somewhere else so it also reloads on each view/crud reload.

romaninsh:
@DarkSide666 so in short - "grid reload looses current sort order" right?

DarkSide666:
yeah :)

now fixed - sort order, paginator page and quicksearch are memorized on grid reload

PR #527, opened 2018-07-09
Fix/issue#524 crud

ibelar:
Fix issue #524 - Keep sort and pagination value when editing field,

### DIFF
--- a/src/CRUD.php
+++ b/src/CRUD.php
@@ -69,6 +69,7 @@ class CRUD extends Grid
         parent::init();
 
         $this->stickyGet($this->paginator->name);
+        $this->stickyGet('_q');
 
         if ($this->canUpdate) {
             $this->pageUpdate = $this->add($this->pageUpdate ?: $this->pageDefault, ['short_name'=>'edit']);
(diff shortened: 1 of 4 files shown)


src/CRUD.php
```diff
@@ -165,16 +165,7 @@ public function initCreate()
...
     {
-        return [
-            // reload Grid
-            (new jQuery($this))->trigger('reload'),
-
-            // close modal
-            new jsExpression('$(".atk-dialog-content").trigger("close")'),
-
-            // display notification
-            $this->factory($this->notifyCreate ?: $this->notifyDefault),
```
DarkSide666: You are loosing notifyCreate message now.

src/CRUD.php
```diff
@@ -68,7 +68,7 @@ public function init()
     {
         parent::init();
 
-        $this->on('reload', $this->jsReload());
+        $this->stickyGet($this->paginator->name);
```
DarkSide666: probably same issue is with QuickSearch / Filter

src/CRUD.php
ibelar: should be ok now!

src/CRUD.php
ibelar: yes!

PR #532, opened 2018-07-25
Atk4/ui/feature/tabs selected tab

skondakov:
Integrate selected tab feature for tabs control

``` php
$tabs->addTab('tab name', null, [true if you want to set this tab as selected, default is false]);
..
```

### DIFF
--- a/demos/tabs.php
+++ b/demos/tabs.php
@@ -12,6 +12,11 @@ $tab = $t->addTab('Static Tab');
 $tab->add(['Message', 'Content of this tab will refresh only if you reload entire page']);
 $tab->add('LoremIpsum');
 
+// set the default active tab
+$t->addTab('Default Active Tab', function ($tab) {
+    $tab->add(['Message', 'This is the active tab by default']);
+})->setActive();
+
 // dynamic tab
 $t->addTab('Dynamic Lorem Ipsum', function ($tab) {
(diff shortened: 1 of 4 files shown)


DarkSide666:
Looks good to me, but not sure if it's good to add this as 3rd parameter. Maybe we should better use seed mechanism for consistency? Something like `$tabs->addTab('tab name', ['active' => true])`

romaninsh:
i also thought - seeds, but our tabs don't have a predictable way to reference them. The only other way is 

$t->addTab('..')->makeActive().

@DarkSide666 , @ibelar your thoughts on this?

ibelar:
makeActive() method sounds good. 
We could also probably use this method definition to enable loading of dynamic tab when made active. As of now, if a dynamic tab is add and set active, it is not loading dynamic content automatically. User still need to click Tab to load content.

romaninsh:
yes this PR should accommodate dynamic tabs also.

DarkSide666:
`makeActive()` sounds good and surely it should also fix dynamic tabs (automatic loading of them when made active).

skondakov:
I agree with all of you guys. 

- Refactored the feature to work with method setActive() after adding a tab.
- Fixed problem with dynamic tabs - loading when a tab is selected
- Added example in demo

PR #534, opened 2018-07-25
fix/exception using wizard

ibelar:
Refactor caughtException

Problem was that Wizard use __atk_callback request too but required exception to be send in html, because not using ajax request.

This will first look for ajax request instead of __atk_callback parameter in request. 
Will send exception in json for all ajax request except for tab because  even if tab use ajax, it require an html output.

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -229,13 +229,12 @@ class App
     {
         $ajax = false;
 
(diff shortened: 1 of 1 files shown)


ibelar:
Method name changed to: isJsonRequest()

Issue #538, opened 2018-07-30
Add option to jsModal to show or hide header


PhilippGrashoff:
Hi,

as of now, jsModals always have the `<div class="header"`> section. It would be great to have an option to show or not show this in jsModal, like a parameter called `showHeader` which can be set to true or false.

At the moment, the header is hard-coded in atk4/ui/js/src/plugins/createModal.js

Best Regards
Philipp

romaninsh:
This is merged

PR #539, opened 2018-07-30
Feature/jsModal options and Modal option

ibelar:
Address issue #538 

 - Option to customize header css class for Modal and jsModal;
 - Option to customize Loading label in jsModal;

### Customizing header css class name for jsModal and Modal:

- using jsModal::setOption($options, $value) 
```
       $jsModal->setOption('headerCss', 'ui blue header')  //will make header color blue.
```
- using Modal::headerCss property for regular modal.

### Removing header in jsModal

 - at creation time passing an empty, null or false value in title parameter;
 - by calling jsModal::removeHeader() method;

### Customizing Loading label in jsModal

- using jsModal::setOption($options, $value) 
```
   $jsModal->setOption('label', '')  //set loading label to an empty string
```

### DIFF
--- a/src/jsModal.php
+++ b/src/jsModal.php
@@ -7,7 +7,15 @@ namespace atk4\ui;
  */
 class jsModal extends jsExpression
 {
-    public function __construct($title, $url, $args = [], $mode = 'json', $hasHeader = true)
+    /**
+     * jsModal constructor.
+     *
+     * @param $title       //When empty, header will be remove in modal.
+     * @param $url
+     * @param array $args
+     * @param string $mode
+     */
+    public function __construct($title, $url, $args = [], $mode = 'json')
     {
         if ($url instanceof VirtualPage) {
             $url = $url->getJSURL('cut');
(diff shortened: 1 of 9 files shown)


src/jsModal.php
```diff
@@ -7,12 +7,61 @@
  */
 class jsModal extends jsExpression
 {
-    public function __construct($title, $url, $args = [], $mode = 'json')
+    public function __construct($title, $url, $args = [], $mode = 'json', $hasHeader = true)
```
DarkSide666: Why don't we use seed mechanism and init() method?

src/jsModal.php
```diff
@@ -7,12 +7,61 @@
...
     {
         if ($url instanceof VirtualPage) {
             $url = $url->getJSURL('cut');
         }
 
         parent::__construct('$(this).atkCreateModal([arg])', ['arg' => ['uri' => $url, 'title' => $title, 'mode' => $mode, 'uri_options' => $args]]);
+
+        if (!$hasHeader) {
```
DarkSide666: I think this can go to init() method or shouldn't this be before parent::construct?
Anyway having 5 parameters in constructor method call looks not well.

romaninsh:
if you make header empty, will it remove the surroinding markup also?

romaninsh:
can you also add support here: https://github.com/atk4/ui/blob/develop/demos/modal.php#L8

passing title as false/empty string or omitting

romaninsh:
and setting first argument to false/null here : https://github.com/atk4/ui/blob/develop/demos/modal.php#L49

src/jsModal.php
romaninsh: no need for extra arg, setting first argument to false/null/empty string would be a intuitive way to switch off header.

Issue #514, opened 2018-07-02
DropDown with icons don't look nice


DarkSide666:
If you put dropdown with icons, then it don't look nice. See this
![9f96efd6f1c01d0d1bb065e81e29745b33f5a44d27ffbf3352 pimgpsh_fullsize_distr](https://user-images.githubusercontent.com/1969119/42158220-84505a3e-7df8-11e8-9691-bb011d589c55.png)

ibelar:
This is more of a css issue. semantic ui dropdown will only use space that it's container is using. 
If this is add in form using fiedlGroup then try not using 'width' option for dropdown field or setup a wider option.

DarkSide666:
Ok, setting `$field->width = 'four'` or using `$field->setStyle('width', '12em')` works for me.

So if there is no other option how to set dropdown menu elements as `no-wrap`, then I guess there is nothing we can do about it :(

DarkSide666:
Adding `white-space: nowrap;` to menu elements helps putting them in single line, but that doesn't make menu container wider and text goes outside of menu container.

romaninsh:
this may also be issue with the surrounding ui or multiple CSS files. Can you try to reproduce this issue in ui/demos/ ?

PR #545, opened 2018-08-19
Feature/demo for 514

DarkSide666:
test case (demo/form) for #514.

See dropdown field "Likes".

### DIFF
--- a/demos/form.php
+++ b/demos/form.php
@@ -39,15 +39,14 @@ $g->addField('surname');
 $g->addField('gender', ['DropDown', 'values' => ['Female', 'Male']]);
 $g->addField('likes', ['DropDown', [
     'isValueRequired' => true,
-    'values' => [
-        'car' => ['Cars', 'icon' => 'car icon'],
+    'values'          => [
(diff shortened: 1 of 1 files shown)


romaninsh:
is that adding demo for a problem? in that case we should keep this open and then include fix and merge.

ibelar:
@DarkSide666 - Not sure what the problem is. The dropdown field is enclosed in a 'Two Wide' css rule  because of the 'width' property being set to two.

The 'two wide' css rule from semantic restricted the width of all elements inside that container to be 12.5%. For sure, dropdown element will not look nice. 

In order to overcome this and make dropdown items look nice, then items element will need to be placed outside that container, probably with absolute positioning. This is not how semantic dropdown module is build. 

Maybe we should ask @mayack to take a look at this, css wise, because I do not see an easy way to fix this without hacking semantic dropdown module.

romaninsh:
@DarkSide666 could it be that your "demo" breaks but it's how it is supposed to be?

romaninsh:
closing as irrelevant.

PR #553, opened 2018-08-31
add scrollbar if needed in modal

DarkSide666:
it's impossible to add it in later phase


romaninsh:
it appears we need to observe in any case :) maybe it should be in the modal itself..?

DarkSide666:
I would vote "YES" on that, but bad thing is that it's impossible to nicely remove that option if in some case we don't need it at all.

PR #557, opened 2018-09-04
add commented example for quicksearch - how to search on each keypress

DarkSide666:
add commented example for quicksearch usage - how to search on each keypress

### DIFF
--- a/demos/grid.php
+++ b/demos/grid.php
@@ -7,6 +7,9 @@ $g = $app->add(['Grid']);
 $g->setModel(new Country($db));
 $g->addQuickSearch();
 
+// Example how to filter quickSearch on each keypress
+//$g->quickSearch->on('keyup', 'input', $g->quickSearch->js()->find('.atk-search-button')->click());
+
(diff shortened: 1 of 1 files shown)


DarkSide666:
Maybe `autoChange` method could be useful in this case too.
Let's see what @ibelar will say about that.

romaninsh:
@ibelar i don't like the keystroke. If i type aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa it will query server 30 times. Also you have navigational keys. This should be quite simple really - set timer 0.5sec by default after keystroke and if nothing was pressed in this time, trigger "onchange" early.

This method is simple but was doing wonders. What do you think?

DarkSide666:
I think it should be option there and allow developer to choose what should trigger action:
- only pressing Enter (default)
- onChange event (when you loose focus for example, or - see next one)
- on each keystroke (+ timeout to debounce required here always) - fires onChange (see above)

Probably last two options are actually the same - _trigger action on change_ (same as keystroke in this implementation)

romaninsh:
closing in favour of #563

PR #559, opened 2018-09-06
Feature/Table Column Resizable

ibelar:
Allow table column to be resizable using drag via new Table method Table::resizableColumn()

 - You can supply callback function that will return new width set via dragging;
-  you can supply an array of column widths when creating the table;

``` php
    /**
     *
     * Make columns resizable by dragging column header.
     *
     * @param null $fx              A callback function with columns widths as parameter.
     * @param null $widths          An array of widths value, integer only. ex: [100,200,300,100]
     * @param null $resizerOptions  An array of column-resizer module options. see https://www.npmjs.com/package/column-resizer
     *
     * @throws Exception
     */
    public function resizableColumn($fx = null, $widths = null, $resizerOptions = null)
    {
        $options = [];
        if ($fx  && is_callable($fx)) {
            $cb = $this->add('jsCallBack');
            $cb->set(function() use ($fx) {
                $widths = [];
                if (isset($_POST['widths'])) {
                    //$widths = explode(',', $_GET['widths']);
                    $widths = json_decode($_POST['widths']);
                }
                return call_user_func($fx, $widths);
            });
            $options['uri'] = $cb->getJSURL();
        } else if ($fx && is_array($fx)) {
            $widths = $fx;
        }

        if ($widths) {
            $options['widths'] = $widths;
        }

        if ($resizerOptions) {
            $options = array_merge($options, $resizerOptions);
        }

        $this->js(true, $this->js()->atkColumnResizer($options));

        return $this;
    }
```

screen shot 2018-09-06 at 11 07 04 am

## Demo
Available in table2.php file.

## Important 
atkui-js need to be rebuild.

### DIFF
--- a/src/Table.php
+++ b/src/Table.php
@@ -316,30 +316,30 @@ class Table extends Lister
     ];
 
     /**
-     *
      * Make columns resizable by dragging column header.
      *
-     * @param null $fx              A callback function with columns widths as parameter.
-     * @param null $widths          An array of widths value, integer only. ex: [100,200,300,100]
-     * @param null $resizerOptions  An array of column-resizer module options. see https://www.npmjs.com/package/column-resizer
+     * @param null $fx             A callback function with columns widths as parameter.
+     * @param null $widths         An array of widths value, integer only. ex: [100,200,300,100]
+     * @param null $resizerOptions An array of column-resizer module options. see https://www.npmjs.com/package/column-resizer
      *
      * @throws Exception
      */
     public function resizableColumn($fx = null, $widths = null, $resizerOptions = null)
     {
         $options = [];
-        if ($fx  && is_callable($fx)) {
+        if ($fx && is_callable($fx)) {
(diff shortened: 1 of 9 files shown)


src/Table.php
```diff
@@ -315,6 +315,47 @@ public function decoratorFactory(\atk4\data\Field $f, $seed = [])
...
+     * @param null $resizerOptions An array of column-resizer module options. see https://www.npmjs.com/package/column-resizer
+     *
+     * @throws Exception
+     */
+    public function resizableColumn($fx = null, $widths = null, $resizerOptions = null)
+    {
+        $options = [];
+        if ($fx && is_callable($fx)) {
+            $cb = $this->add('jsCallBack');
+            $cb->set(function () use ($fx) {
```
romaninsh: use $cb->set($fx)

ibelar: ok done!

romaninsh:
No documentation at all.

PR #560, opened 2018-09-11
Feature/form submit hook in crud

### LINKED ISSUE #558: In CRUD form->error() is not treated correctly
Full example (copy in /demos folder and you can test right away):
```
<?php
require 'init.php';
require 'database.php';

$fn = function ($form)  {
    if ($form->model['numcode'] < 10) {
        return $form->error('numcode', 'Code is to small');
    }
};

// custom form
$f = new \atk4\ui\Form();
$f->onSubmit($fn);			
$c = $app->add(['CRUD', 'formDefault' => $f]);
$c->setModel(new Country($db));
```

**Issue**

When you enter small numcode value in form and try to submit, then as response from server you get:
```
{"success":true,"message":null,"atkjs":"$(\"#atk_admin_crud_edit_userform\").form(\"add prompt\",\"numcode\",\"Code is to small\");\n$(\".atk-dialog-content\").trigger(\"close\");\n$.atkNotify({\"content\":\"Data is saved!\",\"color\":\"green\"});\n$(\"#atk_admin_crud_view\").atkReloadView({\"uri\":\"a.php?atk_admin_crud=2\\x26__atk_reload=atk_admin_crud_view\",\"uri_options\":[],\"afterSuccess\":null})"}
```
As you can see - it contains _add prompt_ part, but right after that it closes dialog and reloads. It also saves that small code in data model / database.

**Expected**

It should simply throw error prompt saying that code is to small and not storing and closing anything. Like in normal Form case.

DarkSide666:
This is one approach how to fix #558 , but doesn't look to good because then developer have to save form model himself and what's even worse - he have to take care of showing CRUD notifier message and closing modal dialog. See in https://github.com/atk4/ui/issues/558#issuecomment-420386692

### DIFF
--- a/src/CRUD.php
+++ b/src/CRUD.php
@@ -152,11 +152,14 @@ class CRUD extends Grid
             $this->formCreate->setModel($this->model, $this->fieldsCreate ?: $this->fieldsDefault);
 
             // set save handler with reload trigger
-            $this->formCreate->onSubmit(function ($form) {
-                $form->model->save();
+            // adds default submit hook if it is not already set for this form
+            if (!$this->formUCreate->hookHasCallbacks('submit')) {
+                $this->formCreate->onSubmit(function ($form) {
+                    $form->model->save();
 
-                return $this->jsSaveCreate();
-            });
+                    return $this->jsSaveCreate();
+                });
+            }
         });
     }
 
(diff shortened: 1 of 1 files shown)


src/CRUD.php
```diff
@@ -152,11 +152,14 @@ public function initCreate()
             $this->formCreate->setModel($this->model, $this->fieldsCreate ?: $this->fieldsDefault);
 
             // set save handler with reload trigger
-            $this->formCreate->onSubmit(function ($form) {
-                $form->model->save();
+            // adds default submit hook if it is not already set for this form
+            if (!$this->formUCreate->hookHasCallbacks('submit')) {
```
romaninsh: formUCreate - typo?

DarkSide666: oh, fixed

Issue #523, opened 2018-07-06
Switch from Semantic-UI to Fomantic-UI


romaninsh:
Fomantic is a community fork of Semantic UI. Currently they have delivered 2.4 release with various improvements and are more likely to progress forward.

I propose that we use Fomantic instead of Semantic.

DarkSide666:
If Fomantic is active project nd it doesn't break compatibility with Semantic then I agree.

vegax87:
I agree

hammy2899:
If you make the decision to switch to FUI please feel free to contact the fomantic team if you need anything or want to know any information.

Great to see you considering fomantic 😃

romaninsh:
@hammy2899 we could be interested to collaborate on some themes / style packs but that would be end of 2018.. Will see how our side-project will go :)

ibelar:
Although Fomantic seem promising, I do not see the advantage of switching now. My understanding is that there lists of features is practically the same then Semantic except for a few more feature in there 2.4 version.
I think we should wait and see what Fomantic will bring in term of new features vs Semantic and base our decision then. 
I do not see any substantial benefit of making the switch now.

DarkSide666:
More or less I agree with @ibelar . Let's allow some time to pass and see how active and promising Fomantic will be.

PhilippGrashoff:
I vote for Fomantic for 2 reasons:
1) they have constant development and releases. This offers us the chance to get bugs that interfere with ATK to get fixed quickly.
2) More important, each person using Fomantic UI and giving feedack will motivate the developers to keep going. Just ask yourselves what keeps you going developing ATK without directly getting any money for it. Aint a part of your motivation the people that use ATK, like it, give new input, share ideas? 

Best regards

DarkSide666:
Good point @PhilippGrashoff

prudho:
In case you want another reason to move to Fomantic, 2.5 was released this morning with new core components: https://fomantic-ui.com/introduction/new.html

romaninsh:
Thanks everyone for the feedback. Glad to see 2.5 out there. Calendar is finally a significant argument in favour of fomantic.

romaninsh:
NOTE: semantic released 2.4 (https://semantic-ui.com/introduction/new.html#/twoFour). But it seems romantic 2.5 is more important for our purposes (calendars), so does not affect our desire to change.

hammy2899:
[...]

PR #567, opened 2018-09-18
Upgrading to Fomantic-UI

romaninsh:
Implements #523 

CDN: https://cdn.jsdelivr.net/npm/fomantic-ui@2.5.0/dist/semantic.css


src/App.php
```diff
@@ -300,21 +299,15 @@ public function initIncludes()
...
 
-        // Calendar
-        $url = isset($this->cdn['calendar']) ? $this->cdn['calendar'] : '../public';
-        $this->requireJS($url.'/calendar.min.js');
-        $this->requireCSS($url.'/calendar.css');
-
         // Agile UI
         $url = isset($this->cdn['atk']) ? $this->cdn['atk'] : '../public';
         $this->requireJS($url.'/atkjs-ui.min.js');
-        $this->requireJS($url.'/agileui.js');
```
romaninsh: this no longer needed? (yey)

romaninsh: if yes, shouldn't we remove file also.

PR #569, opened 2018-09-27
Set view as not rendered if there is exception

DarkSide666:
If View is already rendered, but throws exception in json_encode, then you can't add more elements in View so need to set it as not rendered here.

### DIFF
--- a/src/View.php
+++ b/src/View.php
@@ -729,6 +729,7 @@ class View implements jsExpressionable
                                 'html'    => $this->template->render(),
                                 'id'      => $this->name, ]);
         } catch (\Exception $exception) {
+            $this->_rendered = false;
             $l = $this->add(new self());
             if ($exception instanceof \atk4\core\Exception) {
                 $l->template->setHTML('Content', $exception->getHTML());


romaninsh:
Please explain to me why is this relevant / needed? 

Technically you could say that if rendering generated exception, then we can't consider to be "rendered" and it's not much of a chance, but I'm trying understand what you're saying with adding elements and I can't. 

See also this:

``` php
<?php
/**
 * A very basic view which may contain other elements that are likely to throw exception during render. In which case,
 * the exception will be contained within the view and won't break any of the other UI.
 */
class Volatile extends \atk4\ui\View
{
    public $fx = null;
    function set($cb) {
        $this->fx = $cb;
        return $this;
    }
    function recursiveRender() {
        try {
            parent::recursiveRender();
        } catch(\Throwable $e) {
            //$this->output('ouch');
            //$v = $this->add(['Message', 'We have a problem', 'error']);
            if ($e instanceof \atk4\core\Exception) {
                $msg = $e->getMessage();
                $content = $e->getHTML();
            } elseif ($e instanceof \Error) {
                $msg = get_class($e).': '.$e->getMessage().' (in '.$e->getFile().':'.$e->getLine().')';
                $content = nl2br($e->getTraceAsString());
            } else {
                $msg = get_class($e).': '.$e->getMessage();
                $content =  null;
            }
            if ($this->fx) {
                $v = $this->add('View');
                call_user_func($this->fx, $v, $msg, $content);
                $this->template->appendHTML('Content', $v->getHTML());
                $this->app->html->template->appendHTML('HEAD', $v->getJS());
            } else {
                $this->template->appendHTML('Content', $content);
            }
        }
    }
}
```

DarkSide666:
If view is already marked as rendered and only then throws exception, then this line will throw another exception https://github.com/atk4/ui/pull/569/files#diff-248b7e363055cfdcca5a52260252ef41R733 because you can't add more elements in View if it's already marked as rendered.
That's the issue here.

romaninsh:
oh i see yes that's not good.

PR #570, opened 2018-10-08
Properly implement disabled / read_only fields

### LINKED ISSUE #126: disabled does not work when adding field into form
$f_names->addField('middle_name', ['width'=>'three', 'disabled'=>true]); // 'disabled' attribute doesn't work here (bug?)

### LINKED ISSUE #376: If model field is read-only, it should be reflected on a form
Readonly field should be disabled on a form.

``` php
$m = new Country($app->db);
$m->getElement('name')->read_only = true;

$app->add('Form')->setModel($m);
```

if you try to submit this, you get error as expected:

screen shot 2018-02-21 at 07 51 07

However it's better if form would recognize read-only field and display it's control as disabled.

DarkSide666:
Properly implement disabled / read_only fields.

fix #126
fix #376

### DIFF
--- a/src/FormLayout/Columns.php
+++ b/src/FormLayout/Columns.php
@@ -10,26 +10,27 @@ use atk4\ui\Form;
  */
 class Columns extends Generic
 {
+    /** @var int count of columns */
     public $col = null;
+    
+    /** @var string size CSS class */
     public $size = '';
 
+    /**
+     * Sets form model and adds form fields.
+     *
+     * @param \atk4\data\Model $model
+     * @param array|null       $fields
+     *
+     * @return \atk4\data\Model
+     */
     public function setModel(\atk4\data\Model $model, $fields = null)
     {
         // dont add any fields automatically
         parent::setModel($model, false);
 
         if ($fields === null) {
-            $fields = [];
-            foreach ($model->elements as $f) {
-                if (!$f instanceof \atk4\data\Field) {
-                    continue;
-                }
-
-                if (!$f->isEditable()) {
-                    continue;
-                }
-                $fields[] = $f->short_name;
-            }
+            $fields = $this->getModelFields($model);
         }
 
         $cnt = count($fields);
--- a/src/FormLayout/Columns.php
+++ b/src/FormLayout/Columns.php
@@ -12,7 +12,7 @@ class Columns extends Generic
 {
     /** @var int count of columns */
     public $col = null;
-    
+
     /** @var string size CSS class */
     public $size = '';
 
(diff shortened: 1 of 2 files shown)


src/FormLayout/_Abstract.php
```diff
@@ -87,6 +87,37 @@ protected function _addField($decorator, $field)
...
+     */
+    protected function getModelFields(\atk4\data\Model $model)
+    {
+        $fields = [];
+        foreach ($model->elements as $f) {
+            if (!$f instanceof \atk4\data\Field) {
+                continue;
+            }
+
+            if ($f->isEditable() || $f->read_only) {
```
DarkSide666: Automatically include read_only fields in form too if array of field names is not passed as 2nd argument of setModel()

src/FormLayout/Columns.php
```diff
@@ -10,26 +10,27 @@
...
-                if (!$f instanceof \atk4\data\Field) {
-                    continue;
-                }
-
-                if (!$f->isEditable()) {
-                    continue;
-                }
-                $fields[] = $f->short_name;
-            }
+            $fields = $this->getModelFields($model);
```
DarkSide666: no more code duplication

src/FormLayout/_Abstract.php
```diff
@@ -96,25 +127,27 @@ public function setModel(\atk4\data\Model $model, $fields = null)
...
+            $fields = $this->getModelFields($model);
+        }
 
-                if (!$f->isEditable()) {
-                    continue;
-                }
-                $fields[] = $f->short_name;
+        // prepare array of fields - check if fields are editable or read_only/disabled
+        $add_fields = [];
+        foreach ($fields as $field) {
```
DarkSide666: Checks fields and if field is not editable, then will set disabled=true attribute.
This will work if there is not 2nd argument in setModel() method and also when you pass array of field names. That's for consistency - if field is defined as not editable in model, then there should not be a way to set it as editable in form (you can't save it anyway).

src/FormLayout/_Abstract.php
romaninsh: This may break some backwards compatibility. For example some read-only fields may be sensitive and shouldn't be displayed.

DarkSide666: will change to: `if ($f->isEditable() || ($f->read_only && $f->isVisible())`

this also can have some backwards compatibility issues, but it's impossible to fix without that.
Readonly sensitive fields then should be set as not visible `ui`=>['visible'=>false]

Issue #575, opened 2018-10-08
Disabled fields can still be focused and edited


DarkSide666:
**Issue:**

Disabled (read only) form fields adds CSS class `disabled` and these fields can't be clicked anymore. So if you can't click, you can't edit, right? WRONG !

You can easily navigate to such field with Tab key and edit it's value because `disabled` class is added to field container (div) and field as such doesn't have `disabled="disabled"` attribute set.

**Possible solution:**

I think simplest solution would be to add `disabled='disabled'` attribute to input fields in case their container has `disabled` CSS class. That way fields can't be focused anymore and will not be submitted when form will submit. That's what I call read_only or disabled form field and not just some gray background field with something like `on('click', function(){this.blur();});

romaninsh:
ATK Data shouldn't save those fields because they are read-only , right? Or is it actually possible to change the values in the database?

DarkSide666:
Data don't save them, but if form submits them (input element don't have disabled attribute), then Data will throw appropriate _validation_ exception.

PR #576, opened 2018-10-08
fix/ set input attribute to readonly or disable.

ibelar:
Add two new attribute to input tag: readonly and disabled

fixes: #575 

screen shot 2018-10-08 at 5 43 20 pm

Use:

```
$f = $app->add(new \atk4\ui\Form('segment'));

$f->addField('readonly', ['readonly' => true])->set('read only');
$f->addField('disabled', [ 'disabled' => true])->set('disabled');
```


src/FormField/Input.php
```diff
@@ -24,6 +24,9 @@ class Input extends Generic
 
     public $iconLeft = null;
 
+    public $disabled = false;
```
romaninsh: php doc needed  /**

PR #581, opened 2018-10-18
add formConfig to pass parameters to FUI .form()

PhilippGrashoff:
alike apiConfig and setApiConfig() for adding parameters to FUI .api() function, formConfig and setFormConfig() can be used to add parameters to FUI .form() function, e.g.
```
$orm->setFormConfig(['keyboardShortcuts' => false]);
```

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -551,7 +551,7 @@ class Form extends View //implements \ArrayAccess - temporarily so that our buil
         //var_dump($cb->getURL());
         $this->js(true)
             ->api(array_merge(['url' => $cb->getJSURL(), 'method' => 'POST', 'serializeForm' => true], $this->apiConfig))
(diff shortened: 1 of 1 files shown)


src/Form.php
```diff
@@ -529,7 +551,7 @@ public function ajaxSubmit()
         //var_dump($cb->getURL());
         $this->js(true)
             ->api(array_merge(['url' => $cb->getJSURL(), 'method' => 'POST', 'serializeForm' => true], $this->apiConfig))
-            ->form(['inline' => true, 'on' => 'blur']);
+            ->form(array_merge($this->formConfig, ['inline' => true, 'on' => 'blur']));
```
romaninsh: it's generally better if the formConfig is second argument and can override the defaults (such as 'on'=>blur) Can you please swap it?

PR #585, opened 2018-10-24
Fix/#573 Memorize url

### LINKED ISSUE #573: Grid sorting memorizes previous sort order in URL
When you sort grid by clicking on columns, then it adds sticky parameter in URL.
BUT also it adds (not overwrites) it one more time _manually_ in code.

untitled

First parameter is set by PHP because of it being sticky. Second time (active) parameter is set by JavaScript here https://github.com/atk4/ui/blob/develop/src/Grid.php#L376

ibelar:
Fixes #573 

Also fix pageSelector sorting.


ibelar:
@DarkSide666 - Fix sorting issue when Editing record and also when adding new record.

PR #587, opened 2018-10-31
Feature/dynamic scroll

ibelar:
Implementation of Dynamic scrolling for Lister and Table.

Adding dynamic scroll to a view holding a lister:

``` php
$container = $app->add('View');
$v = $container->add(['View', 'template' => $listerTemplate]);
$m = $l->setModel(new Country($db));
// add dynamic scrolling.
$l->addJsPaginator(20, $container);
```

Adding dynamic scrolling to a table:

``` php
$table = $app->add(['Table']);
$m = $table->setModel(new Country($db));
$table->addJsPaginator(30);
```

Demo File:
- scroll-lister.php;
- scroll-table.php;


ibelar:
@DarkSide666 - Thanks for the feedback
- I have arrange method parameters in different order to satisfy PHP param declaration. Good catch on this one. I have kept parameter separate because $options in this case is really the js plugin $options and do not want to mix with other params.
- I do not have too much choice of using action count in order to determine next page request or not. This is because the table row rendering will output an empty "No record found" html row otherwise and since this is a valide output, the plugin will try to send another request and append multiple No record Found in table.

js/src/plugins/scroll.js
```diff
@@ -79,7 +79,17 @@ export default class scroll extends atkPlugin {
    * @return bool
    */
   hasScrollbar() {
-    return this.$el.height() > this.$scroll.height(); // @todo - implement this properly. This way it somehow already works
+    let innerHeight = this.isWindow ? Math.ceil(this.$el.height()) : Math.ceil(this.$inner.height());
+    let scrollHeight = Math.ceil(this.$scroll.height());
+    if (this.isWindow) {
+      return innerHeight > scrollHeight;
+    } else {
+      if ((innerHeight === scrollHeight) || (innerHeight < scrollHeight)){
```
DarkSide666: can't it simply be like this?
`if (innerHeight <= scrollHeight){`

PR #590, opened 2018-11-08
Feature/Form-Layout-Section with Accordion

ibelar:
## Note 

This PR add ability to setup a layout  section in Form where field can be added to the view.

screen shot 2018-11-20 at 11 50 30 am
screen shot 2018-11-20 at 11 50 38 am
screen shot 2018-11-20 at 11 50 46 am
screen shot 2018-11-20 at 11 50 52 am


src/Accordion.php
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @param string        $title
+     * @param null|callable $callback
+     * @param string        $icon
+     *
+     * @throws Exception
+     *
+     * @return View
+     */
+    public function addItem($title, $callback = null, $icon = 'dropdown')
```
romaninsh: i think addSection() makes more sense.

romaninsh: (naming issue)

src/Accordion.php
```diff
@@ -0,0 +1,164 @@
...
+     * Ex: toggle an accordion from it's index value.
+     * $accordion->jsBehavior('toggle', 1).
+     *
+     * @param string $behavior The name of the behavior for the module.
+     * @param array  $args     The behaviors argument as an array.
+     * @param bool   $when     When this js action is render.
+     *
+     * @return mixed
+     */
+    public function jsBehavior($behavior, $args, $when = null)
```
romaninsh: this is pretty cool functionality... perhaps this could be used across other views too ? Just a thought for now, we can refactor after.

src/AccordionItem.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace atk4\ui;
+
+/**
+ * An accordion item in Accordion.
+ */
+class AccordionItem extends View
```
romaninsh: "Section" makes more sense than "Item".

src/Form.php
```diff
@@ -143,6 +143,47 @@ protected function initLayout()
...
+     */
+    public function addLayoutView($view, $hasDivider = true)
+    {
+        if ($this->layout === null) {
+            throw new Exception('Layout need to be initialized prior to add View.');
+        }
+
+        $v = $this->layout->add($view);
+        if ($hasDivider) {
+            $this->layout->add(['ui' => 'hidden divider']);
```
romaninsh: this makes assumption about layout of the form.

how about Form_Layout->addSection() there ?

also, technically calling form->add(something) and have it inserted into layout is confusing.

src/FormLayout/Generic.php
```diff
@@ -88,6 +88,38 @@ public function addGroup($label = null)
         return $this->add(new self($label));
     }
 
+
+    public function addSection($hasDivider = true)
+    {
+        $v = $this->add('View');
+        if ($hasDivider) {
+            $this->add(['ui' => 'hidden divider']);
```
DarkSide666: shouldn't divider go before View ?

src/FormLayout/Generic.php
```diff
@@ -88,6 +88,38 @@ public function addGroup($label = null)
...
+        }
+
+        return $v->add(['FormLayout/Generic', 'form' => $this->form]);
+    }
+
+    public function addColumnSection($hasDivider = true)
+    {
+        $v = $this->add(['FormLayout/ColumnSection', 'form' => $this->form]);
+        if ($hasDivider) {
+            $this->add(['ui' => 'hidden divider']);
```
DarkSide666: shouldn't divider go before View?

src/FormLayout/Generic.php
ibelar: If not place after, then Submit button is placed right after field in last section.

PR #598, opened 2018-11-20
Garrettcook patch

garrettcook:
Addresses [https://forum.agiletoolkit.org/t/addfiltercolumn-causes-fatal-error/564](https://forum.agiletoolkit.org/t/addfiltercolumn-causes-fatal-error/564).  Also now allows filtering on field of type "text".


romaninsh:
is renaming namespaced "Generic" class into "FilterGeneric" beneficial in some way?

DarkSide666:
I don't see any advantage of this renaming (except adding TypeText which is good).

Also your example doesn't give me error.

Try creating this file in in /demos folder and run it - no errors:
```
require 'init.php';
require 'database.php';

$view = $app->add('View', ['ui' => 'basic segment']);
$g = $view->add(['Grid']);
$m = new Country($db);

$g->setModel($m);
$g->addFilterColumn();
```

So maybe there is something wrong with your other code or other composer dependency which is not namespaced and contains Generic class.
Can you show us your composer.json file and also your myModel class.

garrettcook:
Github won't let me attach the composer.json file, but it is the same as what comes with 1.6.0.  What is meant by myModel class?

DarkSide666:
Did my example above work fine for you?
If it work fine, then please try to make it fail and post it here so I can see what's wrong with it. Try to keep test case as simple as possible to show me your error.

garrettcook:
I don't have the default database set up and maybe there's some problem with another control I'm using.  For now we can just set aside my namespace issue and merge the new TypeText.php file.

romaninsh:
PHP Fatal error:  Cannot declare class atk4\ui\TableColumn\FilterModel\Generic, because the name is already in use in /home/travis/build/atk4/ui/src/TableColumn/FilterModel/FilterGeneric.php on line 14

from test-suite.

garrettcook:
Does this mean the change will not be implemented?  I thought the test-suite example indicated that the issue was present outside of just my environment,

PR #600, opened 2018-11-21
Feature/form section

ibelar:
## Note 
Replace previous Feature/Accordion.

This PR add ability to setup a layout  section in Form where field can be added to the view.

screen shot 2018-11-20 at 11 50 30 am
screen shot 2018-11-20 at 11 50 38 am
screen shot 2018-11-20 at 11 50 46 am
screen shot 2018-11-20 at 11 50 52 am


docs/form.rst
```diff
@@ -104,14 +104,32 @@ Extensions
 
 Starting with Agile UI 1.3 Form has a stable API and we expect to introduce some extensions like:
 
- - Capcha decorator
+ - Captcha decorator
  - File Upload field (see https://github.com/atk4/filestore)
- - Multi-record form
- - Multi-tab form
+ - Multi-record form (see https://github.com/atk4/mastercrud)
+ - Multi-tab form (see https://github.com/atk4/mastercrud)
```
romaninsh: not sure why URL here is relevant.

docs/form.rst
```diff
@@ -104,14 +104,32 @@ Extensions
...
+a View layout for it in order to create their html element. In other words, layout attached to the form
+is responsible of rendering html for fields.
+
+When Form is first initialized, it will provide and set a default Generic layout within the form.
+Then using Form::addField() will rely on that layout to add field View to it and render it properly.
+You may also supply your own layout when creating your form.
+
+Form layout may contain sub layouts. Each sub layout being just another layout view, it is possible
+to nest them, by adding other sub layout to them. This allows for great flexibility on how to dispose
+your fields within Form.
```
romaninsh: dispose?

docs/form.rst
```diff
@@ -335,17 +353,35 @@ needs a bit more info:
...
+Using setModel() on a sub layout
+--------------------------------
+
+You may add field to sub layout directly using setModel method on the sub layout itself.::
+
+    $f = $app->add('Form');
+    $f->setModel($m, false);
+
+    $sub_layout = $f->layout->addSubLayout();
+    $sub_layout->setModel($m, ['first_name', 'last_name']);
```
romaninsh: can $m be optional here btw?

romaninsh: also what happens if a different model is used here... ?

docs/form.rst
```diff
@@ -335,17 +353,35 @@ needs a bit more info:
...
+
+
+When using setModel() on a sub layout to add fields per sub layout instead of entire layout,
+make sure you pass false as second argument when setting the model on the Form itself, like above.
```
romaninsh: ...otherwise........

src/Accordion.php
```diff
@@ -166,7 +166,7 @@ public function renderView()
         }
 
         //Only set Accordion in Top container. Otherwise Nested accordion won't work.
-        if (!$this->owner instanceof AccordionSection && !$this->isDynamicSection()) {
+        if (!$this->getClosestOwner($this, '\atk4\ui\AccordionSection') && !$this->isDynamicSection()) {
```
romaninsh: you can specify \atk4\ui\AccordionSection::class btw... not sure how it would work though. no need to fix!

DarkSide666: let's leave it as is for now because we don't use new `::class` syntax in other places too.

docs/form.rst
DarkSide666: Theoretically $m can be optional and then sub-layout could take model from it's parent layout ...

About different model - no idea, ave not tested that :)

PR #612, opened 2018-12-02
Feature/refactor template

DarkSide666:
* Refactor Template getTagRef() and getTagRefList() methods and related methods
* Add tests for these methods
* Fix Lister rendering. Now correctly using Template `_top` tag
* Make Lister class a bit more consistent with Table class. Introduce renderRow() method like it's in Table class.

### DIFF
--- a/src/Lister.php
+++ b/src/Lister.php
@@ -49,7 +49,7 @@ class Lister extends View
     }
 
     /**
-     * From the current template will extract {row} into $this->t_row and {empty} into $this->t_empty.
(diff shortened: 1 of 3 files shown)


PhilippGrashoff:
nice one!

romaninsh:
Upon further inspection and impact of this PR in 1.6.2 release some things to consider:

 - PR introduces cloning inside rendering loop. That's expensive and may impact performance.
 - PR changes meaning of $lister->t_row, which is public and may be used by other components.

I'd like to benchmark rendering performance and how "cloning" impacts it.

PR #614, opened 2018-12-04
Feature/add dropdown tests

romaninsh:
Using contribution from @PhilippGrashoff to create a method for selecting items in drop-downs.

### DIFF
--- a/demos/autocomplete.php
+++ b/demos/autocomplete.php
@@ -78,7 +78,7 @@ $form->add(['Label', 'Input new country information here', 'top attached'], 'Abo
 $c = new Country($db);
 $c->addExpression('letter1', 'concat("Ends with ", substring([name], -1))');
 
-$form->addField('country1', [
+$form->addField('country_a', [
     'Lookup',
     'model'       => new Country($db),
     'hint'        => 'Lookup field is just like AutoComplete, supports all the same options.',
@@ -86,7 +86,7 @@ $form->addField('country1', [
     'search'      => ['name', 'iso', 'iso3'],
 ]);
 
-$lookup = $form->addField('country2', [
+$lookup = $form->addField('country_b', [
     'Lookup',
     'model'       => $c,
     'hint'        => 'However one or few "filtering" options can be added narrowing down the final result set',
--- a/demos/autocomplete.php
+++ b/demos/autocomplete.php
@@ -32,11 +32,14 @@ $form->addField('country3', [
     'search'      => ['name', 'iso', 'iso3'],
 ]);
 
-//$acc = $form->getField('country_id');
-//$acc->actionRight = ['Button', 'Hello htere'];
 
 $form->onSubmit(function ($f) use ($db) {
-    return $f->model->ref('country1')['name'].' / '.$f->model->ref('country2')['name'].' / '.(new Country($db))->load($f->model['country3'])->get('name');
+    $str = $f->model->ref('country1')['name'].' '.$f->model->ref('country2')['name'].' '.(new Country($db))->tryLoad($f->model['country3'])->get('name');
+    $view = new \atk4\ui\Message('Select:'); // need in behat test.
+    $view->init();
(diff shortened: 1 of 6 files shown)


PhilippGrashoff:
Hi,

this needs one change to make it easily usable: In my version, the css_selector param is for the main <div> containing all Dropdown contents. This is easy in my code as I always set the html ID of this div to the field name, e.g. <div id="firstname" class="dropdown ....">
To make it work with standard atk, the only way to easily get from field name to the html element is the <input>'s name attribute which matches the field name.
So instead of $css_selector param a $field_name param would make sense for easy use of this function. Using CSS sibling selectors should do for finding dropdown icon and the menu.
Will investigate later.

romaninsh:
would need to implement at least one test-case. Ideally it should be with the form submit, but after looking at some dropdown i haven't found anything in our demos that could do that.

romaninsh:
@ibelar - can you look into finishing this up when you get a moment?

PhilippGrashoff:
This css could do the job finding dropdown icon and menue by field name:
'input[name="'.$field_name.'"] ~ 
e.g. `'input[name="'.$field_name.'"] ~  i.dropdown.icon'`
will test locally tonight

PhilippGrashoff:
Hi there, just tested using the css selector mentioned above, it works perfect. So this way the function can be used by just passing the field name as parameter.

Here's the working code for me:
```
/*
     * Selects a dropdown. If $select_option is defined, it will select the item where data-value matches,
     * if not, an option which is not the placeholder "..." and which was not selected before.
     *
     * @param string css_selector         The selector for the main <div>
     * @param string select_option        A specific option to select, can be left null
     * @paran bool   wait_menu_disappear  If true, function waits until Dropdown menu disappeared again
     */
    public function selectDropDown($field_name, $select_option = null, $wait_menu_disappear = false) {
        //find arrow and click
        $dropdown_arrow = $this->webDriver->wait(5)->until(
            \WebDriverExpectedCondition::elementToBeClickable(\WebDriverBy::cssSelector('input[name="'.$field_name.'"] ~ i.dropdown.icon'))
        );
        usleep(100000);
        $dropdown_arrow->click();

        //now options should be visible
        $this->webDriver->wait()->until(
            \WebDriverExpectedCondition::visibilityOfElementLocated(\WebDriverBy::cssSelector('input[name="'.$field_name.'"] ~ div.menu.visible'))
        );

        //store current value for later rollback
        $this->dropDownInitiallySelected = $this->webDriver->findElement(\WebDriverBy::cssSelector('input[name="'.$field_name.'"] ~ div.menu div.selected'))->getAttribute('data-value');

        //select a non-selected element
        $dropdown_options = $this->webDriver->findElements(\WebDriverBy::cssSelector('input[name="'.$field_name.'"] ~ div.menu div'));
        $option_selected = false;

        //if no option to select was specified, select some which is not
        //empty and not selected yet
        if($select_option === null)  {
            foreach($dropdown_options as $option) {
                //do not select the show all option (....)
                if($option->getAttribute('data-value') == '') {
                    continue;
                }
                //do not select the option already active
                if(strpos($option->getAttribute('class'), 'selected') !== false) {
                    continue;
                }
                $option_selected = $option->getAttribute('data-value');
                $option->click();
                break;
            }
        }
        //select
        else {
            $option = $this->webDriver->findElement(\WebDriverBy::cssSelector('input[name="'.$field_name.'"] ~ div.menu div[data-value="'.$select_option.'"]'));
            $option_selected = $option->getAttribute('data-value');
            $option->click();
        }

        //wait for dropdown menu to disappear
        if($wait_menu_disappear) {
            $this->webDriver->wait()->until(
                 \WebDriverExpectedCondition::invisibilityOfElementLocated(\WebDriverBy::cssSelector('input[name="'.$field_name.'"] ~ div.menu.visible'))
			);
        }

        return $option_selected;
    }
```

romaninsh:
awesome!!

PR #615, opened 2018-12-04
Feature/cond form fixes

DarkSide666:
We need some support of conditional forms when using Form sub-layouts, including accordion.

For example, hiding form sub-layout when some checkbox is not checked.
That should be possible by using 2nd ($selector) parameter of `setGroupDisplayRules` method, but we can't show/hide accordion section in that way for example.

Maybe it's somehow possible to include callback mechanism in `setFieldDisplayRules` and `setGroupDisplayRules`? Then we would be able to do almost anything.

**In this PR:**
* fix Radio field input element name (depends on Lister and we refactored template there)
* adds ability to pass object as selector parameter to `setGroupDisplayRules`
* adds test for advanced conditional rules (including sub-layouts and accordion)


DarkSide666:
Our client said he don't need to implement this for form accordion layout. He have done something himself to work around this.

So let's merge this PR in - it still contains other good fixes/changes.

PR #616, opened 2018-12-04
Make demo for dynamic scroll in container better

DarkSide666:
See `demos/scroll-grid-container.php`

@ibelar Is it possible to fix load spinner? It should show up in the middle of container like it is doing when you sort grid by some column. Otherwise it's messing up page structure and flickering.

### DIFF
--- a/demos/scroll-container.php
+++ b/demos/scroll-container.php
@@ -11,7 +11,9 @@ $app->add(['View', 'ui' => 'ui clearing divider']);
 
 $app->add(['Header', 'Dynamic scroll in Container']);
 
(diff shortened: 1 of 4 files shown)


DarkSide666:
```
  addLoader(){
    this.$inner.closest('.segment').addClass('loading');
  }

  /**
   * Remove loader.
   */
  removeLoader() {
    this.$inner.closest('.segment').removeClass('loading');
  }
```

This works perfect for my changed demo file, but fails in other cases when there is no segment element or it's not related with parent view at all.

PR #617, opened 2018-12-07
Feature/dynamic scroll crud

DarkSide666:
Dynamic scroll in Table and Grid works fine, but in CRUD:
1. data table column width doesn't match table header widths.
2. edit, delete buttons move to right and are not visible.

screenshot

Solution which works - remove `collapsing` CSS class from table cells with buttons.
But maybe that way we break something else.

### DIFF
--- a/demos/scroll-grid-container.php
+++ b/demos/scroll-grid-container.php
@@ -3,7 +3,7 @@
 require 'init.php';
 require 'database.php';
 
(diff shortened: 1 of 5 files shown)


DarkSide666:
Works perfect except that it reloads CRUD to page 1 after edit. But that's OK for now I guess. It's not for this PR anyway.
So please somebody approve this PR. I can't because I initially created it.

PR #619, opened 2018-12-12
Fixes #618: Process fields with integer value 0 correctly

### LINKED ISSUE #618: Ternary operator messes up integer with value = 0 in DropDown::getInput()
Hi there,
I do have a dropdown for a field which can have either 0 or 1 as value. The bug is that if the 0 value is selected and stored, on the next load the placeholder ... is selected instead of the 0 value.

The issue is in DropDown::getInput(): The ternary operator( ? : ) messes up int(0) to an empty string in this line:
`$value = isset($this->field) ? $this->app->ui_persistence->typecastSaveField($this->field, $this->field->get()) : $this->content ?: '';`

Adding this code to debug:
```
var_dump(isset($this->field));
var_dump($this->field->get());
var_dump($this->app->ui_persistence->typecastSaveField($this->field, $this->field->get()));
var_dump($value);
```

produces
```
bool(true)
int(0)
int(0)
string(0) ""
```

When changing the ternary operator to normal if/else:
```
               if(isset($this->field)) {
			$value = $this->app->ui_persistence->typecastSaveField($this->field, $this->field->get());
		}
		else {
			if($this->content) {
				$value = $this->content;
			}
			else {
				$value = '';
			}
		}
```

the result is as expected:
```
bool(true)
int(0)
int(0)
int(0)
```

So the fix for this issue would be pretty simple, just change to if/else. The reason for this issue is a strange behaviour of the ternary operator when stacked, see post below.

PhilippGrashoff:
The stacked ternary operator led to a wrong processing of integer 0 values, this fixes it and the value of the dropdown's hidden input is set correctly to 0 instead of an empty string.

Fixes #618

### DIFF
--- a/src/FormField/DropDown.php
+++ b/src/FormField/DropDown.php
@@ -105,7 +105,13 @@ class DropDown extends Input
      */
     public function getInput()
     {
-        $value = isset($this->field) ? $this->app->ui_persistence->typecastSaveField($this->field, $this->field->get()) : $this->content ?: '';
+        //fix for https://github.com/atk4/ui/issues/618
+        if(isset($this->field)) {
+            $value = $this->app->ui_persistence->typecastSaveField($this->field, $this->field->get());
+        }
+        else {
+            $value = $this->content ?: '';
+        }
 
         return $this->app->getTag('input', [
             'name'        => $this->short_name,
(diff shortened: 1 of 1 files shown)


PhilippGrashoff:
Yeah, I could have put braces, but unless you know about that ternary stacked mess, you won't understand the braces. And as probably hardly anyone knows about that totally stupid behavior of stacked ternarys, I thought this is more clear.

PR #627, opened 2019-01-03
Feature/implement tryappend

### LINKED ISSUE #626: Template: tryAppend() and tryAppendHTML() woud be nice
Hi there,

in Template class, trySet(), trySetHTML() and tryDel() already exist which are nice. The same for append() and appendHTML() would be good, then all template editing functions would have their try... complement.
This is very useful if templates are editable for someone else.

DarkSide666:
* Refactor set(), trySet(), setHTML(), trySetHTML(), append(), appendHTML()
* Implement tryAppend(), tryAppendHTML(), fix #626 
* few other related fixes - in Lister and in demo scripts


romaninsh:
all ok.

PR #631, opened 2019-01-10
add ability to set buttonSave=false to not create it

DarkSide666:
Now can use like this:
`$this->add(new Form(['buttonSave'=>false]))`

Maybe even could add ability that if buttonSave is array, then pass it as seed to addButton() method, but not sure about that. How you think?

### DIFF
--- a/demos/form2.php
+++ b/demos/form2.php
@@ -10,6 +10,10 @@ $app->add(['Header', 'Database-driven form with an enjoyable layout']);
 
 // create form
 $form = $app->add(new \atk4\ui\Form(['segment']));
+//$form = $app->add(new \atk4\ui\Form(['segment', 'buttonSave'=>false]));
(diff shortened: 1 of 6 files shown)


romaninsh:
Yes, i'd like ability to pass either object or seed for the button. See how Button handles 'icon' property. Needs to be consistent.

DarkSide666:
Ok will implement that shortly.
Passing as object should already work, but seeds not.

DarkSide666:
* Form constructor now supports seed parameters
* changes in addButton method to support seed parameters
* buttonSave can be passed as seed. See all these cases below now work properly:
```
$app->add(new \atk4\ui\Form(['segment', 'buttonSave'=>false]));
$app->add(new \atk4\ui\Form(['segment', 'buttonSave'=>new \atk4\ui\Button(['Import', 'secondary', 'iconRight'=>'list'])]));
$app->add(new \atk4\ui\Form(['segment', 'buttonSave'=>[null, 'Import', 'secondary', 'iconRight'=>'list']]));
$app->add(['Form', 'segment', 'buttonSave'=>[null, 'Import', 'secondary', 'iconRight'=>'list']]);
```

PR #637, opened 2019-01-18
Use Lookup field for hasOne, fix #636

### LINKED ISSUE #636: hasOne relation should use Lookup field instead of DropDown by default
Currently we use \atk4\ui\FormField\DropDown field by default for hasOne fields.
See https://github.com/atk4/ui/blob/develop/src/Form.php#L383 and below.

But most of the time data models contain a lot of data and DropDown field always loads all records from referenced model and renders all its elements. This is very time and resource heavy and in case referenced data model is huge - it can even crash.

```
// this will by default use \atk4\ui\FormField\DropDown field in forms
// DropDown field loads all its values instantly and renders them, so
// in case of huge data model this will be slow and resource heavy.
$this->hasOne('a_id', new HugeModel());

// To use much nicer and faster alternative use \atk4\ui\FormField\Lookup
// field. It will load only few records and use ajax to load other records
// as soon as you start typing something in lookup field. So this will work
// fast and will not be resource heavy and most likely should be as default
// field for hasOne relation sometime in future.
$this->hasOne('b_id', [new HugeModel(), 'ui'=>['form'=>new \atk4\ui\FormField\Lookup()]]);
```

In case of big referenced data model we should use Lookup field by default to help developers avoid these issues mentioned above.

So my proposal is to replace this:
```
        if ($f->type != 'boolean') {
            if ($f->enum) {
                $fallback_seed = ['DropDown', 'values' => array_combine($f->enum, $f->enum)];
            } elseif ($f->values) {
                $fallback_seed = ['DropDown', 'values' => $f->values];
            } elseif (isset($f->reference)) {
                $fallback_seed = ['DropDown', 'model' => $f->reference->refModel()];
            }
        }
```
with this:
```
        if ($f->type != 'boolean') {
            if ($f->enum) {
                $fallback_seed = ['DropDown', 'values' => array_combine($f->enum, $f->enum)];
            } elseif ($f->values) {
                $fallback_seed = ['DropDown', 'values' => $f->values];
            } elseif (isset($f->reference)) {
                $fallback_seed = ['Lookup', 'model' => $f->reference->refModel()];
            }
        }
```
So it will still use DropDown for enum and values because these are _static ones_and most likely not a big data amount, but use Lookup field by default in case of hasOne reference, because models most likely will have a lot of data.

DarkSide666:
Use Lookup field for hasOne, fix #636

### DIFF
--- a/demos/form2.php
+++ b/demos/form2.php
@@ -80,8 +80,7 @@ class Person extends \atk4\data\Model
         $this->addField('name', ['required' => true]);
         $this->addField('surname', ['ui'=>['placeholder'=>'e.g. Smith']]);
         $this->addField('gender', ['enum' => ['M', 'F']]);
-        $this->hasOne('country_1_id', new Country());
-        $this->hasOne('country_2_id', [new Country(), 'ui'=>['form'=>new \atk4\ui\FormField\Lookup()]]);
+        $this->hasOne('country_id', new Country());
     }
 
     public function validate($intent = null)
--- a/demos/form2.php
+++ b/demos/form2.php
@@ -80,7 +80,8 @@ class Person extends \atk4\data\Model
         $this->addField('name', ['required' => true]);
         $this->addField('surname', ['ui'=>['placeholder'=>'e.g. Smith']]);
         $this->addField('gender', ['enum' => ['M', 'F']]);
-        $this->hasOne('country_id', new Country());
+        $this->hasOne('country_lookup_id', new Country()); // this works fast
+        $this->hasOne('country_dropdown_id', [new Country(), 'ui'=>['form'=>new \atk4\ui\FormField\DropDown()]]); // this works slow
     }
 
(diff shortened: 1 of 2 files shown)


demos/form2.php
```diff
@@ -80,8 +80,7 @@ public function init()
         $this->addField('name', ['required' => true]);
         $this->addField('surname', ['ui'=>['placeholder'=>'e.g. Smith']]);
         $this->addField('gender', ['enum' => ['M', 'F']]);
-        $this->hasOne('country_1_id', new Country());
-        $this->hasOne('country_2_id', [new Country(), 'ui'=>['form'=>new \atk4\ui\FormField\Lookup()]]);
+        $this->hasOne('country_id', new Country());
```
romaninsh: how about a test with Dropdown?

DarkSide666: ok will add

Issue #503, opened 2018-06-22
confirm in View->on() only works with Callback, not with jsReload for example


PhilippGrashoff:
Hi there,

I tried the following (reload a view but only if user confirms):
```
$button_delete->on(
    'click', 
    (new \atk4\ui\jsReload($this->mainForm, ['action' => 'delete', 'delete_id' => $this->mainForm->model->get('id')])), 
    ['confirm' => $this->_tConfirmDelete->render()]);
```

This does not work - the reload is executed as soon as the button is clicked. View->on() only takes care of 'confirm' if a php callback is defined. It would be great if it also worked out of the box otherwise.

romaninsh:
I agree, that would be a good addition.

DarkSide666:
fixed by #642

PR #642, opened 2019-01-30
Add "confirm" option support for js->on() case

DarkSide666:
Now will support 'confirm' option also in this case not only when you use callback.

```
$btn->on(
    'click',
    [
        new jsReload($installer, ['start' => 1], null, ['stateContext' => '#'.$btn->name]),
        $btn->js()->addClass('disabled'),
    ],
    ['confirm' => 'Are you sure you want to install this add-on?']
);
```

### DIFF
--- a/src/View.php
+++ b/src/View.php
@@ -997,6 +997,12 @@ class View implements jsExpressionable
         $chain = new jQuery();
         $actions[] = $chain;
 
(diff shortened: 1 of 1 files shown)


PhilippGrashoff:
fixes #503

PR #644, opened 2019-02-05
bugfix for confirm

DarkSide666:
Sometimes you pass `['confirm'=>false]`


romaninsh:
@DarkSide666 I dropped support for 5.6 not because atk ui fails on it, but because composer crashes in travis, hitting memory limit:

```
The command "composer install" failed and exited with 255 during .
```

PR #648, opened 2019-02-11
make templates-dir easier to customize

DarkSide666:


DarkSide666:
yes array is supported - you can pass multiple template directories.
typehint is here: https://github.com/atk4/ui/pull/648/files#diff-58980ac23a3a9a0b92abd9805eed1e78R37

Issue #649, opened 2019-02-11
Double confirm alert


DarkSide666:
```
$this->add(['Button', 'Disconnect', 'icon'=>'remove'])
    ->on('click', function(){
        ...
    }, ['confirm'=>'Do you really want to disconnect?'])
;
```
This shows confirm dialog 2 times. Shows it and then when you click OK button shows it one more time.

abbadon1334:
i'm on it.
the problem is in generated javascript code

```
$("#_a1183b66__page_crud_button").on("click",function(event) {
--
event.preventDefault();
event.stopPropagation();
if(!confirm("Do you really want to disconnect?")){return;};
$(this).atkAjaxec({"uri":"6?atk_backend_main_course_tabs_tabssubview_2_virtualpage=cut\x26_a1183b66__page_crud_button_click=ajax\x26__atk_callback=1","uri_options":[],"confirm":"Do you really want to disconnect?"});
});

```
this is the first confirm 
```
if(!confirm("Do you really want to disconnect?")){return;};
```
and the second come from this
```
"confirm":"Do you really want to disconnect?"});
```
i debug the function **View->on** with xdebug and $defaults are correctly set

i try to comment #642 changes in ***View->on*** and the generated code is correct

from what i see jsCallback set a confirm indipendent by the logic of View->on.

i look in the history and found that #642 add this bug, adding a confirm prompt to any callback.

to maintain this feature on all callback and use it like #642 i try to comment **confirm** in **atkAjaxec** defaults
``
    public function jsRender()
    {
        if (!$this->app) {
            throw new Exception(['Call-back must be part of a RenderTree']);
        }

        return (new jQuery())->atkAjaxec([
            'uri'         => $this->getJSURL(),
            'uri_options' => $this->args,
            //'confirm'     => $this->confirm,
        ])->jsRender();
    }
``
and with grid buttons works, but i think is a big change any change that refer to callbacks and must be tested more extensively

ibelar:
There is a fix for this issue: #653 

Would not remove confirm option for atkAjaxec since other function might rely on it.

PR #652, opened 2019-02-15
Lookup - fixed possible error on plus button and add feature

abbadon1334:
new feature for plus function in Lookup field.
Add possibility to pass defaults in place of a string or default label

``` php

        $this->hasOne('teacher_id', [
                CourseTeacher::class,
                'required' => true,
                'caption' => __('Default lessons teacher'),
                'ui' => [
                    'form' => [
                        'Lookup',
                        'plus' => ['icon' => 'plus']
                    ]
                ],
        ]);

```
fix issue when Lookup is linked to an extended model with addCondition

``` php

class User extends \atk4\data\Model {

    public function init()
    {
        parent::init();

        $this->addField('email');
        $this->addField('password');
        $this->addField('is_teacher', ['type' => 'boolean']);
    }
}

class Teacher extends User {

    public function init()
    {
        parent::init();
        $this->addCondition('is_teacher', '=', true);
    }
}
```
however when i click the plus button i want to add a new model and i think is right to have a really new one, am i right?

### DIFF
--- a/src/Grid.php
+++ b/src/Grid.php
@@ -83,6 +83,9 @@ class Grid extends View
 
     public $defaultTemplate = 'grid.html';
 
+    // Defines which Table Decorator to use for Actions
+    protected $actionDecorator = 'Actions';
+
     public function init()
     {
         parent::init();
@@ -324,7 +327,7 @@ class Grid extends View
     public function addAction($button, $action, $confirm = false)
     {
         if (!$this->actions) {
-            $this->actions = $this->table->addColumn(null, 'Actions');
+            $this->actions = $this->table->addColumn(null, $this->actionDecorator);
         }
 
         return $this->actions->addAction($button, $action, $confirm);


romaninsh:
It looks like you've got a lot of interesting stuff here, but I couldn't find the "Plus" changes. I'd love to see this finalised, documented and merged. Well done, @abbadon1334!

src/jsCallback.php
```diff
@@ -51,7 +51,7 @@ public function jsRender()
         return (new jQuery())->atkAjaxec([
             'uri'         => $this->getJSURL(),
             'uri_options' => $this->args,
-            'confirm'     => $this->confirm,
+            //'confirm'     => $this->confirm,
```
ibelar: No need to comment. Fixes for #649 has been addressed using #653

src/FormField/Lookup.php
```diff
@@ -132,7 +132,7 @@ public function init()
         $this->template->set('place_holder', $this->placeholder);
 
         if ($this->plus) {
-            $this->action = $this->factory(['Button', is_string($this->plus) ? $this->plus : 'Add new', 'disabled' => ($this->disabled || $this->readonly)]);
+            $this->action = $this->factory(['Button', is_string($this->plus) || is_array($this->plus) ? $this->plus : 'Add new', 'disabled' => ($this->disabled || $this->readonly)]);
```
abbadon1334: this is the change that i mention to customize the plus button in lookup

src/FormField/Lookup.php
```diff
@@ -143,7 +143,7 @@ public function init()
 
         $vp->set(function ($p) {
             $f = $p->add('Form');
-            $f->setModel($this->model);
+            $f->setModel($this->model->newInstance());
```
abbadon1334: this is the other part of code that i mention in the first comment of PR.

When we use an extended model, like : 

- Model Customer extends Model ( master model )
- Model CustomerActive extends Customer ( is an empty model with only the condition state = active )

if i put in standard LookUp it works perfect, but if i add button plus functionality it will load a new model with the defined model condition applied. At this point it will open a form with the first record matching the defined condition and not a new record.

If we click the button we want to add a new Model of that kind and i think ->newInstance is correct

romaninsh:
apologies started a review but haven't submitted it previously. did now.

abbadon1334:
> apologies started a review but haven't submitted it previously. did now.

no apologies, was my first PR on github directly from phpstorm, before this i was doing cut'n'paste, the problem is that i use only to versioning code for work private, i never use it in public.

About copyright, it remains there because i have it as a default.

this PR was supposed to be only for the lookup button, but i messed up :D 

For columns i'm working here : https://github.com/abbadon1334/ui/commits/feature/tablecolumn

Much more commented than this

romaninsh:
Ok, I remember it took several tries for others before they made a first successful PR 👍 

I can move that into a branch and finish the merge when I have time, but if you can - address the issues and commit into the branch.

PR #654, opened 2019-02-19
Removed deprecated each functions from template parsing

abbadon1334:
today while i was playing with error catching and i see that there are a bunch of deprecated each functions and i remove it. I see that are covered by PHPUnit and i run tests with success, need to test more even if the use of each was not so extended.

i don't use foreach, because there are some performance issues with memory/pointers, this function is so heavy used and any minimal change in performance on a big layout will result in a huge latency on TTFB.

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -735,41 +735,60 @@ class Template implements \ArrayAccess
      */
     protected function parseTemplateRecursive(&$input, &$template)
     {
-        while (list(, $tag) = @each($input)) {
-
-            // Closing tag
-            if ($tag[0] == '/') {
-                return substr($tag, 1);
+        while (true) {
+            
+            $tag = current($input);next($input);
+            
+            if($tag === false) {
+                break;
             }
-
-            if ($tag[0] == '$') {
-                $tag = substr($tag, 1);
-                $full_tag = $this->regTag($tag);
-                $template[$full_tag] = '';  // empty value
-                $this->tags[$tag][] = &$template[$full_tag];
-
-                // eat next chunk
-                $chunk = @each($input);
-                if ($chunk[1]) {
-                    $template[] = $chunk[1];
-                }
-
-                continue;
-            }
-
-            $full_tag = $this->regTag($tag);
-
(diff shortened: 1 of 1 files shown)


src/Template.php
```diff
@@ -735,41 +735,59 @@ protected function regTag($tag)
      */
     protected function parseTemplateRecursive(&$input, &$template)
     {
-        while (list(, $tag) = @each($input)) {
```
DarkSide666: Just in case $input is not what we expect here let's add these lines here:
```
if (!is_array($input) || empty($input)) {
    return;
}
```

src/Template.php
```diff
@@ -735,41 +735,59 @@ protected function regTag($tag)
...
+                break;
+                
+                // recurse
+                default :
+                    
+                    $full_tag = $this->regTag($tag);
+                    
+                    // next would be prefix
+                    $prefix = current($input);next($input);
+                    $template[$full_tag] = $prefix ? [$prefix] : [];
```
DarkSide666: Here probably it would be better to validate against `false` and/or `null` like this:
```
$template[$full_tag] = ($prefix===false || $prefix===null) ? [] : [$prefix];
```

romaninsh:
@DarkSide666 your call on this one, you were the one to mess up with Templates last :)

abbadon1334:
i was horribly wrong about performance of current();next(), this is what you pay when you see a bunch of nerds speaking of something they think know well and make you skip benchmarking.

try to run this on your server (multiple times to have an average ) : https://gist.github.com/abbadon1334/5d343216c9ceab486fd2f10ef3cb7fe0

with this changes, we go away from deprecated error, but i think is better to put a TODO note : remove use of while/current/next from those functions.

for sure with the performance difference, on a big layout it will gain something on TTFB

DarkSide666:
Very nice benchmarking script.
Indeed it was a bit of surprise that foreach performs so well. Bad thing about foreach is double memory consumption.

merging this one in because `each()` should be deprecated anyway.

PR #655, opened 2019-02-20
Fix #651 error handling and modal loading loop

### LINKED ISSUE #651: Modal fails on js load if a php exception is thrown
If an exception is thrown in the `set()` function for a modal, the modal will fail to render completely and an error will appear in the console, the modal will stay unresponsive.

Test case
```php

$modal = $app->add(['Modal']);

$modal->set(function($m) use ($modal) {
    throw new \Exception("TEST!");
});

$button = $this->_app->add(['Button', 'Test modal']);
$button->on('click', $modal->show());
```

The following message will appear in the console:
Uncaught TypeError: Cannot create property 'id' on string '{"success":false,"message":"\n<div class [...]
pointing to the following code in the file /js/src/services/ModalService.js (67 line)
```js
onComplete: function(response, content) {
    const result = content.html(response.html);
    if (!result.length) {
        response.success = false;
        response.isServiceError = true;
        response.message = 'Modal service error: Unable to replace atk-dialog content in modal from server response. Empty Content.';
    } else {
        if ($modal.modal.settings.autofocus) {
            modalService.doAutoFocus($modal);
        }
        $modal.modal('refresh');
        //content is replace no need to do it in api
        response.id = null; // error here apparently
    }
}
```

It appears that this function does not actually try to read the response as JSON, instead opting to only read it as html.
Even though there's already handlers for this kind of behaviour, so maybe it's an oversight?
I've tried it with both a normal exception and \atk4\ui\Exception.

Using ATK4 UI 1.6.2 (slightly modified, although it's not related) / ATK4 Core 1.3.8

abbadon1334:
in php 7+ Errors and Exception use the same interface Throwable and can be catched.

1. defined a new function catchThrowable to normalize processing of Exceptions and Errors.
2. move the try/catch from shutdown_function inside run method to catch even when always_run is false and run is called directly
3. if there Debug trait is in use log the error

The problem for the extra output was clearly in catchException where $l echo the $jsonData correctly, but the script hit execution end and $this exec a run, because there was no other flags in that condition that stop him to do it.
To stop this i add at the start of the new method caughtThrowable $this->run_called = true, because in any case : we catch an error and we want to output only the response error.

Teoretically, now every error and exception will pass in caughtThrowable and show and if debugtrait active will be logged too.

Testcase : 
```php
$app->initLayout('Centered');

$modal = $app->add(['Modal']);

$modal->set(function($m) use ($modal) {
    throw new \Exception("TEST!");
});

$button = $app->add(['Button', 'Test modal exception']);
$button->on('click', $modal->show());


$modal2 = $app->add(['Modal']);

$modal2->set(function($m) use ($modal2) {
    trigger_error("error triggered");
});

$button2 = $app->add(['Button', 'Test modal error']);
$button2->on('click', $modal2->show());
```

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -169,7 +169,7 @@ class App
         // Set our exception handler
         if ($this->catch_exceptions) {
             set_exception_handler(function ($exception) {
-                return $this->caughtException($exception);
+                return $this->caughtThrowable($exception);
             });
         }
         
@@ -193,10 +193,30 @@ class App
                 }
             }
             
-            throw new \ErrorException($errstr,$errno,1,$errfile,$errline);
+            $type = 'Unknown Error';
             
-            return true;
-        }, E_ALL);
+            switch ( $errno ) {
+                case E_USER_ERROR:
+                    $type = 'Fatal Error';
+                break;
+                case E_USER_WARNING:
+                case E_WARNING:
+                    $type = 'Warning';
+                break;
+                case E_USER_NOTICE:
+                case E_NOTICE:
+                case E_STRICT:
+                    $type = 'Notice';
+                break;
+                case E_RECOVERABLE_ERROR:
+                    $type = 'Catchable';
+                break;
+            }
+            
+            throw new \ErrorException($type . ':' . $errstr,0,$errno,$errfile,$errline);
+            
+            //return true; no return to stop the execution
+        },E_WARNING);
         
         // Always run app on shutdown
         if ($this->always_run) {
@@ -209,6 +229,13 @@ class App
                 
                 if (is_string($this->_cwd_restore)) {
                     chdir($this->_cwd_restore);
+                }
+                
+                $e = error_get_last();
+                if(!is_null($e))
+                {
+                    $error = new \ErrorException($e['message'],$e['type'],E_ERROR,$e['file'],$e['line']);
+                    $this->caughtThrowable($error);
                 }
                 
                 if (!$this->run_called) {
@@ -841,6 +868,9 @@ class App
         // @TODO add a flag for level of verbosity of errors dispay to user
         $this->catch_runaway_callbacks = false;
         
+        // run_called moved here because is used on any case
+        $this->run_called = true;
+        
         // if class App was extended
         // we get the right class to call
         // if not we get wrong data like : title => Agile Toolkit - untitled
@@ -855,16 +885,10 @@ class App
         if($e instanceof \Exception)
         {
             $exception = $e;
-            
             if ($exception instanceof \atk4\dsql\Exception)
             {
                 $l->layout->template->setHTML('Content', $exception->getHTML());
-                $params = $exception->getParams();
-                
-                if (array_key_exists('query', $params))
-                {
-                    $l->layout->template->appendHTML('Content', nl2br($params['query']));
-                }
+                //$params = $exception->getParams();
             }
             elseif ($exception instanceof \atk4\core\Exception)
             {
@@ -879,7 +903,6 @@ class App
         if($e instanceof \Error)
         {
             $error = $e;
-            
             if ($error instanceof \Error)
             {
                 $l->layout->add(['Message', get_class($error).': ' . $error->getMessage().' (in '.$error->getFile().':'.$error->getLine().')', 'error']);
@@ -904,12 +927,9 @@ class App
         {
             $l->catch_runaway_callbacks = false;
             $l->run();
-            $this->run_called = true;
         }
         
         $this->callExit();
-        
-        return true;
     }
     
     /**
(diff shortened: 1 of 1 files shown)


abbadon1334:
i clearly don't know how to phpunit this correctly, but a manual test case can be this : 

``` php
        $error_types = [
            //\E_ERROR => "E_ERROR",
            //\E_WARNING => "E_WARNING",
            //\E_PARSE => "E_PARSE",
            //\E_NOTICE => "E_NOTICE",
            //\E_CORE_ERROR => "E_CORE_ERROR",
            //\E_CORE_WARNING => "E_CORE_WARNING",
            //\E_COMPILE_ERROR => "E_COMPILE_ERROR",
            //\E_COMPILE_WARNING => "E_COMPILE_WARNING",
            \E_USER_ERROR => "E_USER_ERROR",
            \E_USER_WARNING => "E_USER_WARNING",
            \E_USER_NOTICE => "E_USER_NOTICE",
            //\E_STRICT => "E_STRICT",
            //\E_RECOVERABLE_ERROR => "E_RECOVERABLE_ERROR",
            //\E_DEPRECATED => "E_DEPRECATED",
            \E_USER_DEPRECATED => "E_USER_DEPRECATED",
        ];
        
        $cols = $this->add(['Columns']);
        $colError = $cols->addColumn(8);
        $colError->add($segError = new View(['','ui' => 'ui basic buttons vertical']));
        
        foreach($error_types as $errType => $errName) {
            
            $modal = $this->add(['Modal']);
    
            $modal->set(function ($m) use ($errType,$errName) {
                trigger_error('error: ' . $errName,$errType);
            });
    
            $btn = [
                'Button',
                'Test Err.' . $errName
            ];
            
            $button = $segError->add($btn);
            $button->on('click', $modal->show());
        }
        
        $exception_types = [
            'Exception' => new \Exception('message',0),
            '\atk4\core\Exception' => new \atk4\core\Exception('atk4 core Exception',0),
            '\atk4\dsql\Exception' => new \atk4\dsql\Exception('atk4 dsql Exception',0),
            '\atk4\ui\Exception' => new \atk4\ui\Exception('atk4 core Exception',0),
            '\atk4\atk4\data\Exception' => new \atk4\data\Exception('atk4 core Exception',0)
        ];
        
        $colExc = $cols->addColumn(8);
        $colExc->add($segExc = new View(['','ui' => 'ui basic buttons vertical']));
    
        foreach($exception_types as $excName => $exc) {
        
            $modal = $this->add(['Modal']);
        
            $modal->set(function ($m) use ($excName,$exc) {
                throw $exc;
            });
        
            $btn = [
                'Button',
                'Test Exception :' . $excName
            ];
        
            $button = $segExc->add($btn);
            $button->on('click', $modal->show());
        }
    
        $array = ['test','tested'];
        each($array); // trigger notice;
```

romaninsh:
this would be quite hard to PHPUNIT. There were a bunch of un-caught situations that were messing up callbacks.

romaninsh:
Test indicate error with the file "tablefilter.php" in demos. Can you check, @abbadon1334 ?

src/App.php
```diff
@@ -147,70 +159,196 @@ public function __construct($defaults = [])
...
+                
+                case E_USER_DEPRECATED:
+                    $errName = 'E_USER_DEPRECATED';
+                    $ignoreError = $ignoreDeprecated;
+                break;  // 16384
             }
-        }
-
+            
+            if($ignoreError)
```
abbadon1334: here is the condition that generate errors on phpunit, if you change it to : 
```php
if($ignoreError || true)
```
with this condition we decide what to do with the error:
 - ignore
 - ignore and log
 - output and log

if we ignore throwing errors we go back to starting point, where we not have control over error output and when you open a modal and a server error happens browser will receive mixed json + visual html error : {success:false,message:'blabla html code'} <!doctype><html>...

i look further in log and code, to see if there is a solution or tragically some components depends on this "issue/bug".

abbadon1334: ok any suppressed in the code break the system like in jsSSE.php :
```php
if (@$_GET['event'] === 'sse') {
```

i look now if is only this

abbadon1334:
test passed and found origin of the problems just some old code that trigger some E_NOTICE errors, i decided to disable/ignore display of the error but log it only when debugtrait is active, i think we need a flag or smething that define much bettere if we are in debug or not, because a good app need logging and probably many users have debug trait even in production.

Clearly now is a pleasure, with no more modal in loop and tail -f in console that clearly log where the problem is.

abbadon1334:
i'm still a little bit busy, but in this week i will redo a full immersion on the problem and make the changes and clean some stuff in code

romaninsh:
@abbadon1334 I'm closing this as it's stale, please re-open when you have time.

PR #665, opened 2019-02-26
Defines which Table Decorator to use for Actions

gowrav-vishwakarma:


src/Grid.php
```diff
@@ -83,6 +83,9 @@ class Grid extends View
 
     public $defaultTemplate = 'grid.html';
 
+    // Defines which Table Decorator to use for Actions
+    protected $actionDecorator = 'Actions';
```
DarkSide666: shouldn't it be `Actions::class` ? :)

Issue #664, opened 2019-02-25
Lookup field empties itself when resetting


romaninsh:
Clicking on Lookup field which currently has a value, and then closing it by clicking somewhere else resets the value to the empty one.

(observed in safari)

ibelar:
Fix in #671

DarkSide666:
Should be fixed

Issue #663, opened 2019-02-25
On Custom Form Layout, Lookup FormField has an empty <label> tag


PhilippGrashoff:
Normally, input fields in Custom Form Layout do not have a <label> tag. However, the lookup formfield has an empty <label> tag in a custom form layout.

romaninsh:
Could you provide some screneshot or test-code?

PhilippGrashoff:
![label_on_lookup](https://user-images.githubusercontent.com/33204878/53366972-2d79e280-3945-11e9-8f63-82e2957122ef.png)
The 2 fields with extra indent come from hasOne(), hence Lookup FormField is used now.

ibelar:
Fix in #671

DarkSide666:
@PhilippGrashoff can you confirm that this is fixed now?

romaninsh:
re-open if not fixed.

PR #671, opened 2019-03-04
fix/Lookup Autocomplete fixes

### LINKED ISSUE #656: Width is not applied correctly to elements
Given the following code results in an optical bug where the second field does not size correctly.
Instead of having two equal fields with 50% size, we have the first field sized 50% and the second field 25% since the width is applied twice.
```
$form = $app->layout->add(new \atk4\ui\Form());
$form->addHeader('Example');
$group = $form->addGroup();
$group->addField('trainingResult', ['width'=>'eight']);
$group->addField('trainingNote', ['width'=>'eight']);
```

This is the HTML which is getting generated
```
<div class="eight wide  field">
	<label for="_5694b51f__trainingNote_input">Example</label>

	<div id="_5694b51f__trainingNote" class="ui field eight wide action input" style="">
		<label></label>
		<div id="_5694b51f__trainingNote-ac" class="ui search selection dropdown  " style="width:100%">
			<input name="trainingNote" type="hidden" id="_5694b51f__trainingNote_input" value="">
			<i class="dropdown icon"></i>
			<div class="default text"></div>
			<input class="search" value="" tabindex="0">
			<div class="menu" tabindex="-1"></div></div>
		<div id="_5694b51f__trainingNote_button" class="ui  button" style="">Add new</div>
	</div>
</div>
```

And that's how it should look like to work correctly: (removed two classes from second div)
```
<div class="eight wide  field">
	<label for="_5694b51f__trainingNote_input">Example</label>

	<div id="_5694b51f__trainingNote" class="ui field action input" style="">
		<label></label>
		<div id="_5694b51f__trainingNote-ac" class="ui search selection dropdown  " style="width:100%">
			<input name="trainingNote" type="hidden" id="_5694b51f__trainingNote_input" value="">
			<i class="dropdown icon"></i>
			<div class="default text"></div>
			<input class="search" value="" tabindex="0">
			<div class="menu" tabindex="-1"></div></div>
		<div id="_5694b51f__trainingNote_button" class="ui  button" style="">Add new</div>
	</div>
</div>
```

ibelar:
Various fix for Lookup and Autocomplete field

Fixes #656, #663, #664

- Keep previous select value;
- Proper width apply when used in group;
- Remove empty filter label

### DIFF
--- a/src/FormField/AutoComplete.php
+++ b/src/FormField/AutoComplete.php
@@ -3,6 +3,8 @@
 namespace atk4\ui\FormField;
 
 use atk4\ui\jQuery;
+use atk4\ui\jsExpression;
+use atk4\ui\jsFunction;
 
 class AutoComplete extends Input
 {
@@ -135,6 +137,9 @@ class AutoComplete extends Input
         if ($this->action) {
             $this->action->js('click', new \atk4\ui\jsModal('Adding New Record', $vp));
         }
+
+        $this->settings['forceSelection'] = false;
+
     }
 
     /**
--- a/src/FormField/Lookup.php
+++ b/src/FormField/Lookup.php
@@ -162,6 +162,8 @@ class Lookup extends Input
         if ($this->action) {
             $this->action->js('click', new \atk4\ui\jsModal('Adding New Record', $vp));
         }
+
+        $this->settings['forceSelection'] = false;
     }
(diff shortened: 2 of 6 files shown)


FabulousGee:
Fix for #656 confirmed.

PR #672, opened 2019-03-04
remove header block completely if we don't set title

DarkSide666:


DarkSide666:
Works fine if you pass title=false. It just leaves empty and invisible <div>, but who cares :)

PR #675, opened 2019-03-05
Feature/Fomantic Update

ibelar:
Update to Fomantic version 2.7.2


DarkSide666:
Good to merge, but need to do that before new UI release because JS library needs to be compiled.

romaninsh:
will release 1.6.5 now...

PR #677, opened 2019-03-20
add top and bottom padding to centered layout

PhilippGrashoff:
Centered layout didnt have any top and bottom padding/margin. On pages that exceeded browsers inner height, that looked quite crappy.

Maybe theres a nicer way using some predefined FUI class?


romaninsh:
please include screenshots.

PhilippGrashoff:
Hi,

this is what centered looks like without top&bottom padding/margin. The Heading starts at the very top of the browser window. I resized the window so there's just no scrollbar yet:
centerednopadding

With some top&bottom padding/margin theres always some space to to the browser window, see screenshot. Again I resized the browser window so there's just no scrollbar yet:
centeredwithpadding

PR #678, opened 2019-03-29
Feature/Ui view handle by Vue.js

ibelar:
### Implementation of  Vue.js with Ui view.

- Add vueService to atkjs-ui.js for handling creation of Vue component within atk4\ui;
- Add two new view component:
    - InlineEdit for saving a simple field to db;
    - ItemSearch for searching into db and reloading a view.

### Atkjs File name

- Refactoring and standardizing atkjs-ui file name.

#### Demo in vue-composent.php

Screen Shot 2019-03-29 at 2 47 12 PM


js/src/components/inline-edit.component.js
```diff
@@ -0,0 +1,71 @@
...
+ * value:  array, array of value to send to server.
+ *
+ */
+
+export default {
+  name: 'atk-inline-edit',
+  props: {
+    url: String,
+    initialValue: String,
+    id: Number,
```
romaninsh: i think ID should be a inside URL as a sticky get. The reasons are:
 - there may not be ID, we might want to edit something that has no ID, like "status" of a currently logged-in user.
 - ID is not something that component can change as it wishes.
 - If we introduce a special way to protect "get" arguments by signing them with limited-time token, this may not apply on "ID" and would require a separate protection.
 - similarly - form does not have "ID" as a separate get argument.

js/src/plugins/ajaxec.plugin.js
```diff
@@ -1,4 +1,4 @@
-import atkPlugin from 'plugins/atkPlugin';
+import atkPlugin from './atk.plugin';
```
romaninsh: why atk is not is plugins ? I'm not saying to move it, just curious.

js/src/services/vue.service.js
```diff
@@ -0,0 +1,103 @@
+import Vue from 'vue';
+import atkInlineEdit from '../components/inline-edit.component';
+import itemSearch from '../components/item-search.component';
+
+
+let atkComponents = {
+  'atk-inline-edit' : atkInlineEdit,
+  'atk-item-search' : itemSearch,
```
romaninsh: does this have to be populated for all new components? does not seem very universal.

js/src/services/vue.service.js
```diff
@@ -0,0 +1,103 @@
...
+    if(!VueService.instance){
+      this.vues = [];
+      this.eventBus = new Vue();
+      VueService.instance = this;
+    }
+    return VueService.instance;
+  }
+
+  /**
+   * Created a Vue component and add it to the vues array.
```
romaninsh: description is unclear.  Does that mean that other than those existing 2 components others would need to invoke this method to register themselves? is the name "createAtkVue" miss-leading? maybe not, just raising a question.

js/src/components/inline-edit.component.js
romaninsh: On other hand - perhaps we can pass OTHER arguments, for example if i have multiple inline fields on a page and i want them all to communicate with the same callback JS. Or ability to reference value of other field when submitting. 

Perhaps the reason for ID field is so that we can use this component inside table, which would kinda justify the "ID" field. I'd prefer to have either "args" or nothing.

js/src/services/vue.service.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Emit an event to the eventBus.
+   * Listener to eventBus can respond to emitted event.
+   *
+   * @param event
+   * @param data
+   */
+  emitEvent(event, data = {}) {
```
romaninsh: unclear what this does and how to use it. No doc or example.

src/Component/InlineEdit.php
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Whether callback should save value to db automatically or not.
+     * Default to using onChange handler.
+     * If set to true, then saving to db will be done when model get set
+     * and if model is loaded already.
+     *
+     * @var bool
+     */
+    public $autoSave = false;
```
romaninsh: Perhaps the name of the callback should be "saveCallback" ? I'm also confused if this is same as saveOnBlur..  Maybe a simplification is needed?

src/Component/InlineEdit.php
```diff
@@ -0,0 +1,179 @@
...
+     * @var bool
+     */
+    public $saveOnBlur = true;
+
+    /**
+     * Default css for the input div.
+     *
+     * @var string
+     */
+    public $inputCss = 'ui right icon input';
```
romaninsh: normally we use public $ui = 'right icon input' 

this is inconsistent with other views.

src/Component/InlineEdit.php
```diff
@@ -0,0 +1,179 @@
...
+     *
+     * @return \atk4\data\Model
+     */
+    public function setModel(\atk4\data\Model $model)
+    {
+        parent::setModel($model);
+        $this->modelField = $this->modelField ? $this->modelField : $this->model->title_field;
+        if ($this->autoSave && $this->model->loaded()) {
+            if ($this->cb->triggered()) {
+                $value = $_POST['value'] ? $_POST['value'] : null;
```
romaninsh: it is possible for a callback to pass value. See https://github.com/atk4/ui/blob/develop/docs/callbacks.rst and argument to "getURL()".

so the next line would be:

$this->cb->set(function($value) {

src/Component/InlineEdit.php
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * On error message.
+     *
+     * @param string $message
+     *
+     * @return \atk4\ui\jsToast
+     */
+    public function jsError($message)
```
romaninsh: is that consistent with $form->error() ?

src/Component/ItemSearch.php
```diff
@@ -0,0 +1,103 @@
...
+            $this->q = $this->getQuery();
+        }
+    }
+
+    /**
+     * Return query string sent by request.
+     */
+    public function getQuery()
+    {
+        return $_GET['_q'] ? $_GET['_q'] : null;
```
romaninsh: what if multiple search fields present? can they be nested? wouldn't this conflict?

src/Component/ItemSearch.php
```diff
@@ -0,0 +1,103 @@
...
+     * Set model condition base on search request.
+     *
+     * @param $m
+     *
+     * @return mixed
+     */
+    public function setModelCondition($m)
+    {
+        $q = $this->getQuery();
+        if ($q && ($_GET['__atk_reload'] ? $_GET['__atk_reload'] : null) === $this->reload->name) {
```
romaninsh: I think App should have a method for those magic get arguments. Like here: https://github.com/atk4/ui/blob/develop/src/App.php#L264

src/Component/ItemSearch.php
```diff
@@ -0,0 +1,103 @@
...
+
+        // reloadId is the view id selector name that need to be reload.
+        // this will be pass as get argument to __atk_reload.
+        if ($this->reload instanceof View) {
+            $reloadId = $this->reload->name;
+        } else {
+            $reloadId = $this->reload;
+        }
+
+        $this->js(true, (new jsVueService())->createAtkVue('#'.$this->name,
```
romaninsh: If VUE is built-in component of ATK, then this line should be simpler:

```
$this->vue('atk-item-search', ['reload'=>............... ]);
```

src/jsVueService.php
```diff
@@ -0,0 +1,74 @@
...
+    public function createVue($id, $component, $data = [])
+    {
+        return $this->service->createVue($id, $component, $data);
+    }
+
+    /**
+     * Emit an event that other component can listen too.
+     * Allow Vue instances to talk to each other.
+     *
+     * This output js: atk.vueService.emitEvent("eventName", {});
```
romaninsh: no examples given.

js/src/components/inline-edit.component.js
ibelar: Id value was not mandatory and thought it could be good to have. I removed it.

js/src/plugins/ajaxec.plugin.js
ibelar: not sure what you mean here. atk.plugin.js is the base class for all other plugins. ex: class ajaxec extends atkPlugin...

js/src/services/vue.service.js
ibelar: Vue instance need to have it's component register when you created it. This is one way of doing it and it is ok for component that atk managed. For external component, there would be another way for registering them like the createVue method in Vue service.

js/src/services/vue.service.js
ibelar: Yes. Component defined externally would have to use this method to create the Vue instance.

js/src/services/vue.service.js
ibelar: This would allow for component define in seperate Vue instance to be able to talk to each other. 
For example, I use this is saasty for updating the add-ons category when using the item-search component of the add-on market page. The button emit an event that tell the item-search component the category needed to include in it's query.
Been said, the mechanism is in place but I do not have proper use case yet to use it.

src/Component/InlineEdit.php
ibelar: Update to use field validation in model.

src/Component/InlineEdit.php
ibelar: This is not the regular css template field {$_class} or {$class} which define the view that create the Vue instance. The inputCss value is for defining the inline edit input inside the Vue template.
I have set it up as a user defined value. This is how you would change the input look of the inline edit field.

src/Component/InlineEdit.php
ibelar: Add some explanation.

src/Component/ItemSearch.php
ibelar: Updated to accept multiple item search.

src/Component/ItemSearch.php
ibelar: This is just to make sure we are setting up the model for the proper reload view.

src/jsVueService.php
ibelar: No example to show yet. Will come later.

src/Component/ItemSearch.php
ibelar: Do you mean adding a vue method inside View class?

js/src/plugins/ajaxec.plugin.js
romaninsh: gotcha!

js/src/services/vue.service.js
romaninsh: ok

romaninsh: at some point if we have a vue create guide, should mention that then ;)

js/src/services/vue.service.js
romaninsh: yes, but i mean - this need documentation.

src/Component/ItemSearch.php
romaninsh: yes.

romaninsh:
To be documented in #683

ibelar:
@romans - just push a quick update.
I changed the initial component variable name from "item" to "initData". Think this will reflect more the purpose of the variable. Prior name "item" was referring to the first item-search component we had in Saasty.

romans:
Guys, stop mentioning me. Thanks! :)

PR #690, opened 2019-04-14
afterRow is good to render sub lines related to any line

gowrav-vishwakarma:
Implementation of `afterRow` hook

Required in cases where a few more lines are required to be inserted based on the current line after the current row


gowrav-vishwakarma:
count moved

PR #698, opened 2019-05-02
ui-editable-bug

gowrav-vishwakarma:
Implementation of `New test case for ui=>editable test` 

``` php
$this->addField('no_persist_but_show_in_ui',['never_persist'=>true,'ui'=>['editable'=>true]]);
// this should not be readonly in ui
..
```


romaninsh:
OK, so  this PR is intended to break, Now just need to add a fix.

PR #700, opened 2019-05-03
features/MultiLine Form Field

ibelar:
## MultiLine field

New form input in order to allow add/edit multiple record within a form.

## Usage

```
$f = $app->add('Form');
$f->setModel($invoice, false);

$sub_layout = $f->layout->addSubLayout('Generic');
$sub_layout->setModel($invoice, ['name', 'client_id', 'address']);

$sub_layout = $f->layout->addSubLayout('Generic');

//Add multi line input to the layout.
$ml = $sub_layout->addField('ml', ['MultiLine']);
$ml->setModel($invoice, ['item','cat','qty','price', 'sub', 'tx', 'grd'], 'Items', 'invoice_id');

// add onchange handler and update grand total.
$ml->onChange(function($rows) use ($f) {
    $total = 0;
    foreach ($rows as $row => $cols) {
        foreach ($cols as $col) {
            $fieldName = key($col);
            if ($fieldName === 'grd') {
                $total = $total + substr($col[$fieldName], 4);
            }
        }
    }
 return $f->js(true, null, 'input[name="g_total"]')->val(number_format($total, 2));

}, ['qty', 'price']);
```

Screen Shot 2019-05-03 at 11 06 02 AM

## Demo file
multiline.php

## Feature

- Adding row will use Model/Field default value if set.
- Multiple row deletion;
- Calculate Expression or Callback value from Field definition;
- onChange callback for entire row record or on specify field.
- Validate according to Field definition in model.
- Validation is apply on every row when form is saved.

Screen Shot 2019-05-03 at 11 06 31 AM

### DIFF
--- a/src/FormField/MultiLine.php
+++ b/src/FormField/MultiLine.php
@@ -575,7 +575,7 @@ class MultiLine extends Generic
                                       'url'         => $this->cb->getJSURL(),
                                       'eventFields' => $this->eventFields,
                                       'hasChangeCb' => $this->changeCb ? true : false,
-                                      'options'     => $this->options
+                                      'options'     => $this->options,
                                   ],
                               ]);
     }
--- a/src/FormField/MultiLine.php
+++ b/src/FormField/MultiLine.php
@@ -462,7 +462,7 @@ class MultiLine extends Generic
      *
      * @return Model
      */
-    public function setModel($model, $fields = [], $modelRef = null, $linkField = null)
+    public function setModel(\atk4\data\Model $model, $fields = [], $modelRef = null, $linkField = null)
     {
         //remove our self from model
         if ($model->hasElement($this->short_name)) {
--- a/src/FormField/MultiLine.php
+++ b/src/FormField/MultiLine.php
@@ -1,0 +1,755 @@
+<?php
+/**
+ * Create a multiple line input field.
+ * Allow to add/edit multiple row of a data table.
+ * If model define in Multiline contains expression, these expression
+ * will be evaluate on the fly while entering data.
+ * ex: Invoice Items model
+ *     $model->addExpression('total', ['expr'=>'[qty]*[price]', 'type' => 'money', 'caption' => 'Total']);
+ * If using total field in row of record, then the expression will be evaluated when user enter data in each
+ * row.
+ *
+ * Each rows will be added to db using Multiline::saveRows() method,
+ * usually on form submit. If you have setup a form model where multiline's
+ * model is a reference to your form model, form model should be saved prior to call saveRows() method.
+ *
+ *  $f = $app->add('Form');
+ *  $f->setModel($invoice, false);
+ *  //setup fiel in layout..
+ *
+ *  //add multiline field and set model for it.
+ *  $ml = $f->addField('ml', ['Multiline']);
+ *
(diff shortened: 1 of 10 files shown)


romaninsh:
I'd like $rows to be a model linked with array persistence. Also - user should be able to change that with effect on the UI. 

@ibelar are you sending ALL the rows back to the frontend or just changed ones ?

romaninsh:
I'd like to play with this implementation a little bit before merging.

src/FormField/MultiLine.php
```diff
@@ -0,0 +1,764 @@
...
+                }
+                $value = $col[$fieldName];
+
+                try {
+                    $field = $m->getElement($fieldName);
+                    // save field value only if field was editable in form at all
+                    if (!$field->read_only) {
+                        $m[$fieldName] = $this->app->ui_persistence->typecastLoadField($field, $value);
+                    }
+                } catch (\atk4\core\Exception $e) {
```
romaninsh: we should only catch and display validation errors (they have a dedicated class).

src/FormField/MultiLine.php
```diff
@@ -0,0 +1,764 @@
...
+    /**
+     * Render callback.
+     *
+     * @throws ValidationException
+     * @throws \atk4\core\Exception
+     * @throws \atk4\data\Exception
+     */
+    private function renderCallback()
+    {
+        $action = isset($_POST['action']) ? $_POST['action'] : null;
```
romaninsh: could this be using callback's trigger value?

src/FormField/MultiLine.php
```diff
@@ -0,0 +1,764 @@
...
+     *
+     * @param $exprField
+     * @param $fieldName
+     *
+     * @return int|mixed|string
+     */
+    private function getValueForExpression($exprField, $fieldName, $model)
+    {
+        switch ($exprField->type) {
+            case 'money':
```
romaninsh: ok, FOR NOW, but would have to refactor.

src/FormField/MultiLine.php
```diff
@@ -0,0 +1,764 @@
...
+     * @throws \atk4\core\Exception
+     *
+     * @return mixed
+     */
+    private function getDummyExpression($exprField, $model)
+    {
+        $expr = $exprField->expr;
+        $matches = [];
+
+        preg_match_all('/\[[a-z0-9_]*\]|{[a-z0-9_]*}/i', $expr, $matches);
```
romaninsh: rather should rely on Model's callbacks.. OK for now. Or rather Persistence should do this.

src/FormField/MultiLine.php
ibelar: Was not sure how to check for post load validation error. Thought that typecastLoadField was supposed to do so. Let me know how this should work and I can change it.

src/FormField/MultiLine.php
ibelar: The callback url is render only once for the component but the same callback url is used for two different action.

romaninsh:
can't merge, CI fails.

ibelar:
@romaninsh - ci should passed now. Rename onChange method to onLineChange...

romaninsh:
Thanks!

PR #701, opened 2019-05-07
Feature/integrate actions

romaninsh:
Implements support for ATK Data Actions in ATK UI through Generic UI:

### Implicit use in CRUD
``` php
$files->addAction(
    'import_files',
    [
        'args'=>['path'=>['type'=>'string','required'=>true]],
        'scope'=>atk4\data\UserAction\Action::NO_RECORDS
    ]
);

$files->addAction('download', function($m){ 
  file_get_contents($m['path']);
  $m->app->terminate(); 
} );

$crud->setModel($files);  // "Download" button will be next to each file
```
image

### Explicit executor use
``` php
$app->add($executor = new \atk4\ui\ActionExecutor\ArgumentForm());

$executor->setAction($model->getAction('import_files'));

// optionally we can augment return handling
$executor->addHook(
  'afterExecute', 
  function($x, $ret) { return new \atk4\ui\jsToast('Files imported: '. $ret); }
);
```
image

## TODO

 - [ ] CRUD should show actions with scope=NO_RECORDS and scope=ALL_RECORDS in menu. Clickable.
 - [ ] CRUD should enable "selectable" at least one action with scope=MULTIPLE_RECORDS exist. If clicked before items are selected, should show error.
 - [ ] CRUD should include actions automatically along with "edit" and "delete" with scope=SINGLE_RECORD
 - [ ] CRUD should support ability to specify which actions must be presented
 - [x] We need some View that would display and execute single-item records.
 - [ ] Must support warning if action is dangerous
 - [x] Must support output from the action
 - [x] Support for default executor (will display form for missing arguments, then console with output)
 - [x] Must support custom action executors
 - [ ] Support use of custom model for defining arguments
 - [ ] Respect action permissions. If action is not permitted for the user hide(or disable) control
 - [ ] Support for model-type arguments (single-record)

## Explanation

Various UI elements will become aware of Model actions and will provide UI to invoke a corresponding PHP method. The "UserAction" class will provide meta-information on how the method should be called. 

# Use Cases

I'm collecting some of the potential use-cases here which can be solved by this PR:

## Emailing Invoice

User has a CRUD of Invoices. He can click "Email Invoice" button on any client. Modal window will appear showing a Preview of a specific invoice. Two buttons below "Confirm" and "Cancel". Preview is generated by non-ui $action->preview callback. Actual emailing happens by invoking nun-ui $action->callback. Arguments for both are identical and in this case - empty.

Executor = Preview\HTML

## Generate VAT report

User have to supply START and END date as arguments to a non-ui callback "generateVATReport". Arguments must be supplied by a user. No preview. When submitted, callback return model "VatReport" that was just generated. 

We can use Hook of the executor, to properly redirect the user:

Executor = ArgumentForm

## Big Red Button

This is a executor which displays a message along with a big red button. Will use name and description of the action. If button is clicked, action is executed.

Executor = Confirmation


ibelar (approved):
looks good!

PR #702, opened 2019-05-07
Only prepend namespace for known classes

romaninsh:
Currently `atk4\ui\App` normalize class will prepend `\atk4\ui\` when using short name of the class, like this:

``` php
$app->add('CRUD');

# factory uses \atk4\ui\CRUD
```

However, this requires us to always use full name for the class:

``` php
$app->add('\my\test\MyClass')
```

This is incompatible with:

``` php
use \my\test\MyClass;

$app->add(MyClass::class);
```

because it uses `my\test\MyClass` without opening backslash.

This PR fixes problem by checking for the presence of the class inside \atk4\ui\ before attempting to use the prefix.

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -376,8 +376,33 @@ class App
      */
     public function normalizeClassNameApp($name)
     {
-        if (class_exists('\\'.__NAMESPACE__.'\\'.$name)) {
-            return '\\'.__NAMESPACE__.'\\'.$name;
+        /**
+         * @see https://agile-core.readthedocs.io/en/develop/factory.html#FactoryTrait::normalizeClassName
+         * replacing / to \
+         */
+        $checkClass = str_replace("/","\\",$name);
+
+        // check FQCN existence without prepend \\
+        // @case $name = "\\externalNamespace\\className"
+        if(class_exists($checkClass))
+        {
+            return $checkClass;
+        }
+
+        $checkClass = "\\" . $checkClass;
+        // check FQCN existence prepending \
+        // @case $name = "externalNamespace\\className"
+        if(class_exists($checkClass))
+        {
+            return $checkClass;
+        }
+
+        $checkClass = "\\" . __NAMESPACE__ . $checkClass;
+        // check FQCN existence prepending FQNS \atk4\ui
+        // @case $name = "FormField/AutoComplete"
+        if(class_exists($checkClass))
+        {
+            return $checkClass;
         }
 
         return $name;
--- a/src/App.php
+++ b/src/App.php
@@ -379,6 +379,7 @@ class App
         if (class_exists('\\'.__NAMESPACE__.'\\'.$name)) {
             return '\\'.__NAMESPACE__.'\\'.$name;
         }
+
         return $name;
     }
 
--- a/src/App.php
+++ b/src/App.php
@@ -376,7 +376,10 @@ class App
      */
     public function normalizeClassNameApp($name)
     {
-        return '\\'.__NAMESPACE__.'\\'.$name;
+        if (class_exists('\\'.__NAMESPACE__.'\\'.$name)) {
+            return '\\'.__NAMESPACE__.'\\'.$name;
+        }
+        return $name;
     }
 
     /**
--- a/src/App.php
+++ b/src/App.php
@@ -380,28 +380,25 @@ class App
          * @see https://agile-core.readthedocs.io/en/develop/factory.html#FactoryTrait::normalizeClassName
          * replacing / to \
          */
-        $checkClass = str_replace("/","\\",$name);
+        $checkClass = str_replace('/', '\\', $name);
 
         // check FQCN existence without prepend \\
         // @case $name = "\\externalNamespace\\className"
-        if(class_exists($checkClass))
-        {
+        if (class_exists($checkClass)) {
             return $checkClass;
         }
 
-        $checkClass = "\\" . $checkClass;
+        $checkClass = '\\'.$checkClass;
         // check FQCN existence prepending \
         // @case $name = "externalNamespace\\className"
-        if(class_exists($checkClass))
-        {
+        if (class_exists($checkClass)) {
             return $checkClass;
         }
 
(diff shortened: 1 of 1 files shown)


abbadon1334:
> Is a good feature but i suggest add a @todo can be improved caching results in array?

nevermind, profiling the case it shows that is still checked by char like the other comparison class/interface functions, but cache is not needed.
Sorry but using socket i need to see any slowness and memory leak even if is relatively obviously that is managed at low level.

https://gist.github.com/88bc2fde67ee2e584559ae3ce0ccfe3d

src/App.php
```diff
@@ -376,7 +376,10 @@ public function addStyle($style)
      */
     public function normalizeClassNameApp($name)
     {
-        return '\\'.__NAMESPACE__.'\\'.$name;
+        if (class_exists('\\'.__NAMESPACE__.'\\'.$name)) {
```
abbadon1334: __NAMESPACE__ return the namespace with single "\" you need to str_replace("\","\\",_NAMESPACE__) i think.

NOTE: I think we need to talk about declaration of argument type :
- fast no control at runtime :: in comments for static analysis
- "slow" strict check at runtime :: in function declaration raise exception

related video of Rasmus Lerdorf - https://www.youtube.com/watch?v=rKXFgWP-2xQ

php 5.6 support ended on 01-01-2019 ( https://www.php.net/supported-versions.php )

abbadon1334:
```

        /* can be defined as constant */
        $ns_sepa = chr(92);

        /** test against untouched name */
        if (class_exists($name)) {
            return $name;
        }

        /** normalize namespace */
        $normalized_namespace = $ns_sepa.__NAMESPACE__;

        /** normalize name */
        $normalized_name = $ns_sepa.trim($name,$ns_sepa);

        /** test against normalize name */
        if (class_exists($normalized_name)) {
            return $normalized_name;
        }

        /** test against normalized namespace + normalize name */
        if (class_exists($normalized_namespace.$normalized_name)) {
            return $normalized_namespace.$normalized_name;
        }

        /** last chance to check it */
        /* give error even with this extra check
        $atk4_namespace = $ns_sepa.'atk4/ui';
        if (class_exists($atk4_namespace.$normalized_name)) {
            return $atk4_namespace.$normalized_name;
        }
        */

        /* it will throw exception for sure check to help sleepy dev */
        // char(47) = / aka wrong slash used as separator
        if(strpos($name,chr(47)) !== false)
        {
            throw new Exception('Wrong separator used for class, detect / in place of \ called from class ' . $name);
        }

        /** class not exists DIE */
        throw new Exception('Class with name ' . $name . ' not exists, checked ' . 
$normalized_name. ' | ' . $normalized_namespace . $normalized_name);
```

BUT even if in this way it works as intended, it redeclare classes and throw errors in some demos unittest.
Errors are due to misuse of wrong slash in place of backslash in $name.
We can make a str_replace and lose performance and solving a problem of the dev, in my opinion this is wrong, or throw an exception and make it changes the code or add a flag and make this function switchable in App.

abbadon1334: i remember somewhere there was something on this subject of usage of slash in place of backslash

https://agile-core.readthedocs.io/en/latest/factory.html#namespace

we need to find where to put the str_replace of slash to backslash, before all or at the end and make another call to the function and finally throw exception?

```
public function normalizeClassNameApp($name)
    {
        /* can be defined as constant */
        $ns_sepa = chr(92);

        /** test against untouched name */
        if (class_exists($name)) {
            return $name;
        }

        /** normalize namespace */
        $normalized_namespace = $ns_sepa.__NAMESPACE__;

        /** normalize name */
        $normalized_name = $ns_sepa.trim($name,$ns_sepa);

        /** test against normalize name */
        if (class_exists($normalized_name)) {
            return $normalized_name;
        }

        /** test against normalized namespace + normalize name */
        if (class_exists($normalized_namespace.$normalized_name)) {
            return $normalized_namespace.$normalized_name;
        }

        /** last chance to check it */
        /* give error even with this extra check
        $atk4_namespace = $ns_sepa.'atk4/ui';
        if (class_exists($atk4_namespace.$normalized_name)) {
            return $atk4_namespace.$normalized_name;
        }
        */

        /* it will throw exception for sure check to help sleepy dev */
        // char(47) = / aka wrong slash used as separator
        if(strpos($name,chr(47)) !== false)
        {
            return $this->normalizeClassNameApp(str_replace(chr(47), chr(92), $name));
        }

        /** class not exists DIE */
        throw new Exception('Class with name ' . $name . ' not exists, checked ' . $normalized_name. ' | ' . $normalized_namespace . $normalized_name);
    }
```

gowrav-vishwakarma:
Why not using it like 

```
use \my\test\MyClass;
$app->add(new MyClass);

```

abbadon1334:
> Why not using it like
> 
> ```
> use \my\test\MyClass;
> $app->add(new MyClass);
> ```
IMHO because  :

- using new you make the object too coupled
- using new you if you change the construct of that object you have to refactor in all project, opensource speaking all user of the project need to do that modification or put an argument optional which is not ok

i'm working on \atk4\core\ConfigTrait to be used like slim with callback in config to instantiate objects against interface implementation

like :
```
$return = [
    "logger.name" => "app",
    "logger.path" => "var/log/app-log.log",
    "\Psr\Logger\LoggerInterface" => function($container) {
        $logger = new \Monolog\Logger($container->getConfig('logger.name'));
        /* init logger blah blah blah */
        return $logger;
    }
];
```

romaninsh:
OK lets discuss on a hangout later.

romaninsh:
will keep using "class_exist" here even if it has some performance hit.

abbadon1334:
@romaninsh OK, i check the flow and i see that exception for class existence is throw in atk4/core/FactoryTrait

romaninsh:
Looks like a better implementation is in https://github.com/atk4/ui/pull/726. Perhaps I can split that out into a separate PR.

romaninsh:
Oh you've done that for me!! yey.

PR #705, opened 2019-05-10
feature/Alert user when leaving page with un submit form.

ibelar:
Alert user when leaving page with un submit form data.

``` php
$app->add(['Form', 'canLeave' => false]);
```

Screen Shot 2019-05-09 at 8 33 42 PM


romaninsh:
does this also apply on regular forms too ?

romaninsh:
@ibelar this breaks test-scripts.. ..

ibelar:
@romaninsh - Add change to fix test.

Issue #720, opened 2019-05-24
Grid and CRUD is not aware of having menu=>false


DarkSide666:
```
$this->add(['CRUD', 'menu' => false']); // or Grid
```

This will throw exceptions in different places because not everywhere in Grid and CRUD class code we're aware of situation when menu is not defined at all.

PR #721, opened 2019-05-24
Fix #719 and #720

### LINKED ISSUE #719: Bug in custom CRUD->itemCreate support
See lines here https://github.com/atk4/ui/blob/develop/src/CRUD.php#L135
That implementation is wrong.

DarkSide666:
Fix #719 and #720

### DIFF
--- a/src/CRUD.php
+++ b/src/CRUD.php
@@ -133,7 +133,7 @@ class CRUD extends Grid
     {
         // setting itemCreate manually is possible.
         if (!$this->itemCreate) {
-            if (!$this > menu) {
+            if (!$this-> menu) {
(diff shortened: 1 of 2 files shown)


romaninsh:
PHP Notice:  Use of undefined constant menu - assumed 'menu' in /home/travis/build/atk4/ui/src/CRUD.php on line 136

src/CRUD.php
```diff
@@ -133,6 +133,9 @@ public function initCreate()
     {
         // setting itemCreate manually is possible.
         if (!$this->itemCreate) {
+            if (!$this>menu) {
```
kyro46: missing - before >

PR #722, opened 2019-05-24
fix/File upload - jsModal Css

### LINKED ISSUE #717: Upload field dialog opens multiple times
Form upload field opens file browse dialog multiple times if clicked on field (input) element.
If you click on upload icon (on the right side of input element) then if opens browse dialog only once (as expected).

ibelar:
### Other Fix:
- Babel configuration.
    - fix core.js warning during build.
- file-upload.plugin
    - remove support for opening file dialog when input get focus. Because File dialog
    will blur and re focus input field again. This was causing file dialog to open
    multiple times. Now only open when input or button is clicked.

### Enhancement
- create-modal.plugin
    Now allow to pass css class name to create modal.
    Useful when need to create modal with full width for example.

``` php
$btn = $app->add(['Button', 'Add']);
$btn->on('click', (new jsModal('Add new', $vp))->setOption('modalCss', 'fullscreen'));
```

### DIFF
--- a/js/Release.md
+++ b/js/Release.md
@@ -4,7 +4,7 @@
 
 - Babel configuration.
     - fix core.js warning during build. 
-- create-modal.pluing
+- create-modal.plugin
     - fix: now allow to pass css class name to create modal. 
     Useful when need to create modal with full width for example.
 - file-upload.plugin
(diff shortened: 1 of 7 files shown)


PhilippGrashoff:
Thumbs up for the modal enhancement, very nice feature

PR #725, opened 2019-06-02
fix/Generic Input onChange event

ibelar:
- Allow to pass extra argument to onChange event handler.

ex: When you need to set values of the second Dropdown according to user selected value of the first Dropdown.
```
$dropdown->onChange(function($j) use ($other_dropdown) {
     return $other_dropdown->js()->dropdown('change values', $this->getValues($_POST['selected_id']?? null));
}, ['args' => ['selected_id' => $dropdown->jsInput()->val()]]);

function getValues($id)
{
   // get values
  // return this format
  return [['value' => 'value', 'text' => 'text', 'name' => 'name']];
}
```
This will make $_POST['selected_id'] available in server request callback. Then  you can setup other Dropdown values accordingly.


demos/invoice-addon.php
```diff
@@ -0,0 +1,178 @@
...
+//    'itemLink' => 'invoice_id',
+//    'paymentModel' =>  new \atk4\invoice\Model\Payment($db),
+//    'tableFields' => ['reference', 'client', 'date', 'due_date', 'total', 'balance'],
+//    'headerFields' => ['reference', 'date', 'due_date', 'client_id', 'paid_total'],
+//    'footerFields' => ['subtotal', 'tax', 'total'],
+//    'itemFields'   => ['item', 'qty', 'rate', 'amount'],
+//    'paymentRelations' => ['invoice_id' => 'id', 'client_id' => 'client_id'],
+//    'paymentEditFields' => ['method', 'paid_on', 'amount', 'details'],
+//    'paymentDisplayFields' => ['client','balance', 'paid_total','total'],
+//     ]));
```
romaninsh: is that a functional demo? everything is commented apart from a table.

src/FormField/Generic.php
```diff
@@ -134,10 +134,11 @@ public function renderView()
      *
      * @param string|jsExpression|array|callable $expr
      * @param bool                               $useDefault
```
romaninsh: remove this line.

src/FormField/Generic.php
```diff
@@ -134,10 +134,11 @@ public function renderView()
...
+     * @param array                              $default
      *
      * @throws \atk4\ui\Exception
      */
-    public function onChange($expr, $useDefault = true)
+    public function onChange($expr, $useDefault = true, $default = [])
```
romaninsh: get rid of 2nd arg

src/FormField/Generic.php
```diff
@@ -134,10 +134,11 @@ public function renderView()
...
     {
         if (is_string($expr)) {
             $expr = new \atk4\ui\jsExpression($expr);
```
romaninsh: if is_bool($default) {

PR #726, opened 2019-06-02
Some new TableColumns

abbadon1334:
Some TableColumn classes 

- TableColumn/ExternalLink : open an external link or an url with download attribute
- TableColumn/KeyValue : to show values from a enum field
- TableColumn/NoValue : show a defined value in place of an empty string
- TableColumn/Labels : convert a comma separated list of values to labels
- TableColumn/Tooltip : that add an iconTooltip (TT content from another field)
- TableColumn/ColorRating : generate a background color gradient based on the value

in the DocBlock of every class there is instructions for the usage, try to use it and see if it is useful.

After that we can create demos and docs for it.

P.S. in the PR there is even the #702 issue, and i don't know how to remove it :(

 - [ ] Documentation
 - [x] Demos


acicovic:
Awesome!

src/TableColumn/ExternalLink.php
```diff
@@ -0,0 +1,129 @@
...
+ * Use
+ *   new Link('http://yahoo.com?id={$id}');
+ * or
+ *   new Link(['order', 'id'=>'id' ]);
+ * or
+ *   new Link(['order', 'id' ]);
+ * or
+ *   new Link([['order', 'x'=>$myval], 'id' ]);.
+ */
+class ExternalLink extends Generic
```
DarkSide666: How is this class better than `TableColumn\Link` class we already have?

src/TableColumn/KeyValue.php
```diff
@@ -0,0 +1,89 @@
...
+ * 'table' => [
+ * 'KeyValue'
+ * ]
+ * ],
+ * ]);
+ *
+ *
+ * @package atk4\ui\TableColumn
+ */
+class KeyValue extends Generic
```
DarkSide666: I think TableColumn/Generic should have this functionality implemented.

src/TableColumn/Labels.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * from => label1,label2 | to => div.ui.label[label1] div.ui.label[label2]
+ *
+ * @package atk4\ui\TableColumn
+ */
+class Labels extends Generic
+{
+    public function getHTMLTags($row, $field)
+    {
+        $values = explode(',', $field->get());
```
DarkSide666: Theoretically $field->get() can already return array if model field type=array, so I would wrap this in `if (is_string($field->get()))` like this

```
$values = $field->get();
if (is_string($values)) {
    $values = explode(',', $values);
}
```

src/TableColumn/Tooltip.php
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @package atk4\ui\TableColumn
+ */
+class Tooltip extends Generic
+{
+    /**
+     *
+     * @var string $field_tooltip
+     */
+    public $icon = [];
```
DarkSide666: why empty array here if it's a string type

src/TableColumn/Tooltip.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     *
+     * @var string $tooltip_field
+     */
+    public $tooltip_field = [];
```
DarkSide666: why empty array here if it's a string type

src/TableColumn/ExternalLink.php
abbadon1334: i can merged tags in link, it was one of the first columns i have done

src/TableColumn/KeyValue.php
abbadon1334: Ok it try to merge it

src/TableColumn/Tooltip.php
abbadon1334: was a typo

src/TableColumn/Tooltip.php
abbadon1334: was a typo

romaninsh:
Merged through #743 , closing this.

PR #731, opened 2019-06-04
#730 checkbox missing use Exception

### LINKED ISSUE #730: \atk4\ui\CheckBox does not contain the required declaration for \atk4\ui\Exception
Add the following line to \src\FormField\CheckBox.php

```php
use atk4\ui\Exception;
```

Otherwise if the `->set()` function is called incorrectly an error is printed saying:

Error: Class 'atk4\ui\FormField\Exception' not found (in /srv/apache/www/virtual_git/wego2-x/lib/atk4/Ui/src/FormField/CheckBox.php:64)

abbadon1334:
fix #730

### DIFF
--- a/src/FormField/CheckBox.php
+++ b/src/FormField/CheckBox.php
@@ -3,6 +3,7 @@
 namespace atk4\ui\FormField;
 
 use atk4\ui\Form;
+use atk4\ui\Exception;
 
 /**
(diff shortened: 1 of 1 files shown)


abbadon1334:
@romaninsh i think we need to start put some static analysis tool like phpstan in our project

PR #733, opened 2019-06-05
Feature/allow dbconnect before setlayout

romaninsh:
Currently when you execute `dbConnect` before `initLayout` you get this error:

image

This is because app tries to do `$this->layout->add(persistence)`. This does not make sense, persistence is not supposed to go inside layout.

All we need for persistence is to set the `app` property, so this PR does that directly and the error is no longer popping up.


PhilippGrashoff:
Nice one thanks

PR #739, opened 2019-06-11
Fix caughtException() and normalizeClassNameApp()

### LINKED ISSUE #740: App->normalizeClassNameApp doesn't support 2nd parameter - $prefix
Currently atk4\core\FactoryTrait calls $app->normalizeClassNameApp() method and pass 2nd $prefix parameter, but \atk4\ui\App->normalizeClassNameApp() method doesn't support this argument and completely ignores $prefix.

This is bad and can be seen, for example, if you get some exception and it calls App->caughtException() which tries to initLayout('Centered').

Fix is here: https://github.com/atk4/ui/pull/739

DarkSide666:
1. `caughtException` should use static class not always core \atk4\ui\App class (in case we have extended App class).

2. `normalizeClassNameApp` should not ignore $prefix argument which it gets from FactoryTrait. fix #740

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -385,7 +385,7 @@ class App
 
         // if we have prefix, then use it - it's there for a purpose
         if ($prefix) {
-            $checkClass = $prefix.'\\'.$checkClass;
+            $checkClass = str_replace('/', '\\', $prefix).'\\'.$checkClass;
         }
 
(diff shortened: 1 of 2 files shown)


romaninsh:
closing. Proper fix is "develop2" branch.

PR #748, opened 2019-06-18
fix/Crud notifier namespace

ibelar:
- Fix notifier namespace
 - Factory method now required fully qualified namespace or prefix.

## Develop v2.0 Only


romaninsh:
Already fixed by adding 3rd argument to factory in jsSave()

Issue #749, opened 2019-06-19
Default template does not allow hints in checkboxes/toggles in form fields


pkly:
In the file located at `/template/semantic-ui/formlayout/generic.html` there's no tag for the hint to be displayed at all when the field used is a Checkbox or inherits from it.

```html
{Content}
{LabeledGroup}
<div class="{$field_class} field atk-form-group">
  <label>{$label}</label>
  <div class="{$width} {$class} fields">{$Content}</div>{$Hint}
</div>{/}
{NoLabelGroup}
<div class="{$field_class} field atk-form-group">
  <div class="{$width} {$class} fields">{$Content}</div>
</div>{/}
{InputField}
<div class="{$field_class} field">
  <label for="{$label_for}">{$label}</label>{$Input}
  {$Hint}
</div>{/}
{InputNoLabel}
<div class="{$field_class} field">{$Input}</div>{/}
{/}
{$Buttons}
```

In the `{InputNoLabel}` tag a change should be made to allow the hint to be displayed.
```html
{InputNoLabel}
<div class="{$field_class} field no-label">{$Input}{$Hint}</div>{/}
{/}
```

Along with the following changes added to css:
```css
.field.no-label > *:first-child {
display: block;
}
```

It will look like this when used: https://i.imgur.com/RlFBfCW.png

PR #750, opened 2019-06-19
Small changes to templates (form/centered layout)

pkly:
Fixes: #749 

Adds support for custom icons in headers inside of the Centered layout.

``` php
$app->initLayout('Centered');
$app->layout->image = '/admin/assets/img/my-icon.png';
```

Adds two additional tags for inserting html without re-writing the whole header section in the Centered layout.


src/Layout/Centered.php
```diff
@@ -15,6 +15,10 @@ class Centered extends Generic
 
     public $defaultTemplate = 'layout/centered.html';
 
+    // Default atk4 logo
+    public $image = 'https://github.com/atk4/ui/raw/07208a0af84109f0d6e3553e242720d8aeedb784/public/logo.png';
```
romaninsh: the $app class has a CDN string, which you can use.

pkly:
@DarkSide666 I've added the requested changes, also quickly tested them on my install, everything seems to work as expected
Also this is actually a better solution since you can just override the app and work with that instead, I didn't think of it, thanks

src/Layout/Centered.php
```diff
@@ -15,14 +15,23 @@ class Centered extends Generic
...
+    public $image = null;
     public $image_alt = 'Logo';
 
     public function init()
     {
         parent::init();
 
+        // If image is still unset load it when layout is initialized from the App
+        if ($this->image === null && $this->app && $this->app->cdn['atk-logo']) {
+            $this->image = $this->app->cdn['atk-logo'];
```
romaninsh: This->app->cdn[‘atk’].’logo.gif’

romaninsh: Dark no need to duplicate atk

romaninsh:
And yes we do overuse cdn in project quite frequently. I wonder if there is a doc for that.

src/Layout/Centered.php
DarkSide666: Oh right https://cdn.jsdelivr.net/gh/atk4/ui@1.7.1/public/logo.png will work too. I didn't thought about that.
Then this should simply be:
1. remove atk-logo from App->cdn
2. and here do like this:
```
if ($this->image === null && $this->app) {
    $this->image = $this->app->cdn['atk'] . '/logo.png';
}
```

pkly: This way is rather bad though.
For example, right now I can use the code that's implemented to just add my image directly to the App and forget about it.
The way you proposed it will never work, unless my image is named "logo.png", which is probably not ideal.

I'd ask you to reconsider, maybe change the cdn string to 'logo' instead of 'atk-logo'

pkly: @romaninsh and @DarkSide666 how about this solution? I think it's the best of both worlds https://github.com/atk4/ui/pull/750/commits/35ab1b488a17660da558aa0be829b9366c89fb15

pkly: I've also removed the logo url completely from App later, but forgot in that commit, but you can look over the next changes obviously.

DarkSide666:
Well, it looks good this way I guess. @romaninsh how you think?

PR #756, opened 2019-06-26
fix #755

### LINKED ISSUE #755: Data loss: 0 as textarea value
I have a field in my model which shows as a textarea and is of type TEXT in a MySQL backend.

When you add a value of 0 into the textarea, it stores it in the database. However, when the textarea is reloaded it is empty. If the user doesn't notice this and saves the record, data loss occurs because an empty string is saved over the 0.

This issue might also occur in text inputs but I haven't tested it.

DarkSide666:


acicovic:
This fix causes #760 issue

romaninsh:
i have reverted this fix. @DarkSide666 please re-submit the PR and @acicovic can you verify it works now?

acicovic:
Textareas work well for me, except the #755 issue of course, which this fix was fixing. Really strange issue.

PR #761, opened 2019-07-01
Revert "fix #755"

### LINKED ISSUE #755: Data loss: 0 as textarea value

romaninsh:
Reverts atk4/ui#756

### DIFF
--- a/demos/form.php
+++ b/demos/form.php
@@ -37,7 +37,6 @@ $g = $form->addGroup(['width' => 'three']);
 $g->addField('name');
 $g->addField('surname');
 $g->addField('gender', ['DropDown', 'values' => ['Female', 'Male']]);
-$form->addField('description', ['TextArea'])->set(0); // testing 0 value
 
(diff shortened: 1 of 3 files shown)


DarkSide666:
did this really helped to fix issue ??

acicovic:
@DarkSide666 yes, this solved it for me (and apparently for @romaninsh as well) but this is really strange.

acicovic:
I see that the whole commit has been reverted. In my tests, changing only textarea.php was enough to fix the issue. But I have no idea.

romaninsh:
@acicovic @DarkSide666 please prepare a fixed PR & test.

PR #768, opened 2019-07-03
feature/Form Action Executor

ibelar:
Create Form action executor for grid user action
 - enable to edit record via action;

Grid may also add Preview executor
 - Example for delete purpose where user can confirm action.

Screen Shot 2019-07-03 at 5 27 56 PM

### DIFF
--- a/src/ActionExecutor/Form.php
+++ b/src/ActionExecutor/Form.php
@@ -17,6 +17,10 @@ class Form extends Basic
 
     /**
      * Initialization.
+     * If form model is not set then will use action fields property to set model field.
+     * If action fields property is empty then will use all model fields as default.
+     *
+     * If model is already supply in form, then editable fields must match action fields property.
      */
     public function initPreview()
     {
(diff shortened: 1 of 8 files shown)


src/ActionExecutor/Form.php
```diff
@@ -0,0 +1,61 @@
+<?php
+/**
+ * Action form executor.
+ */
```
romaninsh: should we merge Argument into this ?

ibelar:
@romaninsh - Made some change in order to ease js process when using action without having to create afterExecute hook as well. Now, jsSuccess may also be a function that receive the executor an the model. So for a delete executor for example:

```
$del_executor = new atk4\ui\ActionExecutor\Preview([
    'previewType' => 'text',
    'hasHeader'   => false,
    'jsSuccess'   => function ($ex, $model) use ($g) {
        return [
            new atk4\ui\jsExpression('$(".atk-dialog-content").parent().modal("hide")'),
            new atk4\ui\jsToast('Record deleted with success!'),
            $g->table->jsRemoveRow($model->get('id')),
        ];
    },
   ]
);
```

PR #771, opened 2019-07-11
Allows to add HTML markup in addParagraph

DarkSide666:
I guess sometimes we would like to allow HTML markup in `addParagraph()`.
Important here is to not allow it by default, but if developer is sure controlled HTML is ok there, then why not?

### DIFF
--- a/src/Text.php
+++ b/src/Text.php
@@ -25,14 +25,15 @@ class Text extends View
      * Adds HTML paragraph.
      *
      * @param string $text
+     * @param bool   $allow_html
      *
      * @return $this
      */
-    public function addParagraph($text)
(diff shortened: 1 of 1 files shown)


romaninsh:
i don't think it's a good way. Normally I'll add View inside. Also - perhaps we could support markdown markup too here?

DarkSide666:
You can still set HTML directly in Text view by using `set()`... if you don't agree with adding this 2nd parameter then I will use `set()` instead :)

romaninsh:
Yes, i think the use of set() is better - or directly modifying the template.

romaninsh:
Actually you should consider:

```
$text->add(['template'=>new Template('<html here>')]);
```

PR #772, opened 2019-07-11
The default function should return as it is

gowrav-vishwakarma:
but it is returning void and that breaks demos app where in init.php you use default App.

The correct way to leave the empty function is to return what is received in params.

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -384,6 +384,7 @@ class App
      */
     public function normalizeClassNameApp($name, $prefix = '')
     {
+        return $prefix.$name;
         //return '\\'.__NAMESPACE__.'\\'.$name;
     }
 


PhilippGrashoff:
I am wondering if this function should be implemented with any functionality at all in vanilla atk. As normalizeClassName() first checks for a result from this function and exits afterwards, nearly all functionality of normalizeClassName is not executed at all.

DarkSide666 (changes requested):
Should be at least `return $prefix.'\\'.$name;` but even then I'm not sure if this is needed.
See @PhilippGrashoff comment above.

romaninsh:
no this is wrong - new implementation of the method must return "null" for the default resolution method to work.

romaninsh:
also - travis has lots of errors because of this fix.

PR #774, opened 2019-07-15
upgrade npm

romaninsh:
Upgrade npm dependencies... @ibelar is that ok?


romaninsh:
errors are solved in mainstream.

PR #775, opened 2019-07-15
refactor to use getFields()

romaninsh:
atk4\data (`develop`, v2 branch) is now expects you to use `$model->getFields()`. I have refactored by removing any uses of $model->elements directly (except when it's used with a View).

### DIFF
--- a/src/FormLayout/_Abstract.php
+++ b/src/FormLayout/_Abstract.php
@@ -119,18 +119,7 @@ abstract class _Abstract extends \atk4\ui\View
      */
     protected function getModelFields(\atk4\data\Model $model)
     {
-        $fields = [];
-        foreach ($model->elements as $f) {
-            if (!$f instanceof \atk4\data\Field) {
-                continue;
-            }
-
-            if ($f->isEditable() || $f->isVisible()) {
-                $fields[] = $f->short_name;
-            }
-        }
-
-        return $fields;
+        return array_keys($model->getFields('editable'));
     }
 
     /**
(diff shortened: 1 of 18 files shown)


src/FormLayout/_Abstract.php
```diff
@@ -119,18 +119,7 @@ public function addFields($fields)
...
-                continue;
-            }
-
-            if ($f->isEditable() || $f->isVisible()) {
-                $fields[] = $f->short_name;
-            }
-        }
-
-        return $fields;
+        return array_keys($model->getFields('editable'));
```
PhilippGrashoff: Above you replaced similar code, but with
```
return array_keys($model->getFields('not system'));
```

Here the replacement is
```
return array_keys($model->getFields('editable'));
```

Are both filters equal?

abbadon1334: is here the filters, they are not equal : 
 https://github.com/atk4/data/blob/15908690f1f580686c88e59a6045edcd6f1267ca/src/Model.php#L671

isEditable() is here : 

https://github.com/atk4/data/blob/15908690f1f580686c88e59a6045edcd6f1267ca/src/Field.php#L485

romaninsh: the "non system" code was looking for editable/visible fields. There is no filter like that. @ibelar may be looking to display all fields there, so I wasn't sure what's the best filter, but if i used "editable" then it wouldn't display R/O fields.

PR #776, opened 2019-07-15
Implement $_POST test for Form

romaninsh:
Just a bit of Form() testing making sure it's parsing POST correctly.

### DIFF
--- a/tests/FormTest.php
+++ b/tests/FormTest.php
@@ -25,7 +25,7 @@ class FormTest extends \atk4\core\PHPUnit_AgileTestCase
     public function testFormSubmit()
     {
         $f = new Form();
-        $f->app = new AppMock([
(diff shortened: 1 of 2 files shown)


tests/FormTest.php
```diff
@@ -17,4 +21,56 @@ public function testGetField()
...
+
+            // security check, unspecified field must not be changed
+            $this->assertEquals(false, $f->model['is_admin']);
+
+            $submit_called = true;
+        });
+
+        $f->render();
+        unset($f);
+
```
PhilippGrashoff: I'd empty $_POST at the end of the test. In case there will be more tests in this class below, they could produce false results from the values still set to $_POST AFAIK.

romaninsh: sure.

Although IIRC, phpunit does that too.

PR #780, opened 2019-07-20
Fix #651 - Modal fails on js load if a php exception is thrown

### LINKED ISSUE #651: Modal fails on js load if a php exception is thrown

abbadon1334:
Hope this fixes #651, but probably it will solve strange issue of no feedback in other components of ui.

what i have done : 

- I need to remove some notice suppressor in code.
- i use php 7.2 syntax, because composer now require php >= 7.2
- i removed some code that is not need anymore.
- Now any error will be raised as an exception
- Error and Exception will be handled by $app->caughtException($e)

Sample case to compare with previous version :
```php

    require 'vendor/autoload.php';

    $app = new \atk4\ui\App();

    $app->initLayout('Centered');

    $modal = $app->add(['Modal']);

    $modal->set(function($m) use ($modal) {
        throw new \Exception("TEST!");
    });

    $button = $app->add(['Button', 'Test modal exception']);
    $button->on('click', $modal->show());

    $modal2 = $app->add(['Modal']);

    $modal2->set(function($m) use ($modal2) {
        trigger_error("error triggered");
    });

    $button2 = $app->add(['Button', 'Test modal error']);
    $button2->on('click', $modal2->show());
```
Now it will output the view, if an exception is raised, it will open a modal with the description of the exception, previous version loop for ever.

I think we need more test cases, like exception/error during SSE.

ATTENTION
This change, will catch any error, this means that any bad/old habits like use of @ for suppression of notice in place of *isset* or access to undefined will throw an error, PROBABLY good code will not notice any problem.

We can added a variable to define the level of error_reporting, but now i think is better to leave it on E_ALL

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,11 +62,11 @@ before_script:
 
 after_script:
     - echo $TRAVIS_BRANCH
-      #- if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.1" ]]; then echo "Sending coverage report"; vendor/bin/test-reporter --coverage-report clover.xml; fi
-      #- if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.1" ]]; then echo "Sending codecov report"; TRAVIS_CMD="" bash <(curl -s https://codecov.io/bash) -f clover.xml; fi
+    - if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.2" ]]; then echo "Sending coverage report"; vendor/bin/test-reporter --coverage-report clover.xml; fi
+    - if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.2" ]]; then echo "Sending codecov report"; TRAVIS_CMD="" bash <(curl -s https://codecov.io/bash) -f clover.xml; fi
 
 script:
-    - if [[ ${TRAVIS_PHP_VERSION:0:3} != "7.1" ]]; then NC="--no-coverage"; fi
+    - if [[ ${TRAVIS_PHP_VERSION:0:3} != "7.2" ]]; then NC="--no-coverage"; fi
     - ./vendor/phpunit/phpunit/phpunit $NC #- wget -O - http://localhost:8888/demos/button2.php 
     - bash tools/test.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,8 @@ before_script:
     - wget  https://selenium-release.storage.googleapis.com/${SELENIUM_MAJ_VER}/${SELENIUM_JAR}
     - java -jar ${SELENIUM_JAR} > /dev/null 2>&1 &
 
-    - wget -N http://chromedriver.storage.googleapis.com/2.44/chromedriver_linux64.zip -P ~/
+    # Setup chromedriver
+    - wget http://chromedriver.storage.googleapis.com/${CHROME_DRIVER_VER}/chromedriver_linux64.zip -P ~/
     - unzip ~/chromedriver_linux64.zip -d ~/
     - rm ~/chromedriver_linux64.zip
     - sudo mv -f ~/chromedriver /usr/local/share/
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,12 @@ services:
     - mysql
     - xvfb
 
+env:
+  global:
+    - SELENIUM_MAJ_VER=3.141
+    - SELENIUM_VER=${SELENIUM_MAJ_VER}.59
+    - SELENIUM_JAR=selenium-server-standalone-${SELENIUM_VER}.jar
+    - CHROME_DRIVER_VER=75.0.3770.90
 
 addons:
   apt:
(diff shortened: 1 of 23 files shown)


abbadon1334:
@romaninsh i think there is problem on travis build process

abbadon1334:
Solved the issue with Travis and while i was watching one of my friends on IronMan Zurich competition, i solved some issue and refactor some of SSE, because in my next project i need to have jsSSE stable.

.travis.yml
```diff
@@ -24,8 +25,8 @@ before_install:
 
 before_script:
 
-    - "sh -e /etc/init.d/xvfb start"
-    - "export DISPLAY=:99.0"
+     #- "sh -e /etc/init.d/xvfb start"
+     #- "export DISPLAY=:99.0"
     - "wget http://selenium-release.storage.googleapis.com/2.43/selenium-server-standalone-2.43.1.jar"
```
PhilippGrashoff: Old version of selenium, newest one is 3.141.59 :)
https://www.seleniumhq.org/download/

src/TableColumn/Generic.php
```diff
@@ -106,7 +106,8 @@ public function setHeaderPopup($popup, $icon = 'caret square down')
         $popup->popOptions = array_merge($popup->popOptions, ['on' =>'click', 'position' => 'bottom right', 'movePopup' => false]);
         $popup->stopClickEvent = true;
 
-        if (@$_GET['__atk_reload']) {
+        $need_reload = $_GET['__atk_reload'] ?? false;
```
PhilippGrashoff: cant this be done like 
`if(($_GET['__atk_reload'] ?? false)) {` ?
Would save the extra var $need_reload which doesnt seem to be used anywhere else.

src/TableColumn/FilterModel/Generic.php
```diff
@@ -102,7 +102,8 @@ public function afterInit()
             // create a name for our filter model to save as session data.
             $this->name = 'filter_model_'.$this->lookupField->short_name;
 
-            if (@$_GET['atk_clear_filter']) {
+            $atk_clear_filter = $_GET['atk_clear_filter'] ?? false;
```
PhilippGrashoff: cant this be done like 
`if(($_GET['atk_clear_filter'] ?? false)) {` ?
Would save the extra var $atk_clear_filter which doesnt seem to be used anywhere else.

abbadon1334:
all this work after solving exception issues, was original intended to pass some time while raising the coverage, i gained only +6% i think i gained more.

On the other hand i discovered some old code to be converted.

i see that jsSSE.php is always used and SSE.php is never used, but on the lastone there is a beginning of reconnection/keepalive implementation that is absent in jsSSE, probably because the browser do this by itself?

.travis.yml
```diff
@@ -53,8 +62,8 @@ before_script:
 
 after_script:
     - echo $TRAVIS_BRANCH
-    - if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.1" ]]; then echo "Sending coverage report"; vendor/bin/test-reporter --coverage-report clover.xml; fi
-    - if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.1" ]]; then echo "Sending codecov report"; TRAVIS_CMD="" bash <(curl -s https://codecov.io/bash) -f clover.xml; fi
+      #- if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.1" ]]; then echo "Sending coverage report"; vendor/bin/test-reporter --coverage-report clover.xml; fi
+      #- if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.1" ]]; then echo "Sending codecov report"; TRAVIS_CMD="" bash <(curl -s https://codecov.io/bash) -f clover.xml; fi
```
romaninsh: no more coverage ?

romaninsh:
oh and btw - codecov is experienncing outage now, so can merge anyways.

romaninsh:
Any idea where this output is from? 

https://travis-ci.org/atk4/ui/jobs/562371710#L581

abbadon1334:
> Any idea where this output is from?
> 
> https://travis-ci.org/atk4/ui/jobs/562371710#L581

yes it's the jsSSE, i tried silencing it, but whatever i tried different PHPUnit or Test fails and at the end i leave it like this.

.travis.yml
abbadon1334: yep i removed because i'm a sleepy-coder-idiot ;) i do the same on atk4/api.
I think in both repos, that it was a sort of exclusion, which was wrong.
Now it works, but due to a missing starting point for the PR there is no comparison coverage.
But... i raised it a little ;)

Issue #781, opened 2019-07-22
DropDown data-value=0 is treated as null


acicovic:
When we fill a DropDown with data, its data-value property is populated with each item's ID. Although rare, the value "0" is sometimes a legitimate choice for a database ID column value.

When this happens, ATK UI will consider 0 as null during its dropdown validation, and display a "must not be null" validation error if 0 is selected. However, the user has made a valid selection and zero is not the same thing as null.

P.S: This also reminds of #755 which is also an issue of how a zero value is handled as null. So zero handling should be perhaps reconsidered in ATK and fix it wherever it doesn't work correctly.

acicovic:
Another case when this issue is a blocker, is when a value of 0 is declared in a model. For example
```
// Code in model addFields()

['gender_id',
    'required' => true,
    'values' => [0 => 'Female', 1 => 'Male'],
]
```
Even if you don't specify the numeric values. selection "Female" will still be blocked because it will be evaluated to 0.

PhilippGrashoff:
This was there a while ago, I fixed it, but recent changes obviously brought it back...
#618 
So if this is fixed again, we really need a test for it. My bad I didnt write a test also when writing the fix

romaninsh:
Leaving this to you then ;) make a test this time.

PhilippGrashoff:
@acicovic without diving into this as I am not on newest versions yet: Does setting 
`type` => `integer` help with the DropDown?

acicovic:
Setting `'type' => 'integer'` results in no difference.

PhilippGrashoff:
I see Imants is already on it, thanks @DarkSide666 !

romaninsh:
Test-case here: https://github.com/atk4/ui/pull/783/files#diff-f2ca0641b3256c64a192f93071ce4ba0R162, @acicovic please check!

acicovic:
It doesn't work for me and brings back #760.

acicovic:
#760 is fixed. @DarkSide666 as this is essentially a data issue, is it tracked somewhere in the data repository? If yes let's add a reference to that issue here.

DarkSide666:
It should be somewhere in this huge model field class refactor branch https://github.com/atk4/data/pull/375, but that PR is still not 100% ready for merge (only for version 2.0).

mvorisek:
@acicovic still valid? can you provide steps to replicate on demos?

acicovic:
@mvorisek give me some days and I'll get back to you.

[...]

Issue #760, opened 2019-06-29
Form HTML output in textareas


acicovic:
On the latest UI dev update, HTML code is inserted in textareas instead of being rendered as FUI controls. For example that's the content that might be in a textarea:
```
</div>

</div>
<div class=" field">
  <label for="atk_admin_studentform_generic_generic_4_personality_type_id_input">Personnalité</label>
<div id="atk_admin_studentform_generic_generic_4_personality_type_id" class="ui fluid search selection dropdown  " style="" >
```

acicovic:
I have no idea how and why, but this works now. Closing this and will re-open if I notice it again.

Thanks!

acicovic:
This happens only when the form is in create mode. If a record has been loaded all fields display correctly.

acicovic:
The cause is #756

DarkSide666:
Generated HTML in /demos/form.php looks good to me:

```
<div class=" field">
  <label for="atk_admin_tabs_tabssubview_form_2_generic_description_input">Description</label><div id="atk_admin_tabs_tabssubview_form_2_generic_description" class="ui  input" style="">
<textarea name="description" type="text" rows="2" placeholder="" id="atk_admin_tabs_tabssubview_form_2_generic_description_input">0</textarea>
</div>

  
</div>
```

acicovic:
I didn't have the time to test this further to find the cause but reverting TextArea.php solves the specific issue for me.

acicovic:
Note that I'm speaking about latest develop version, not 1.7.1

romaninsh:
I can confirm this too.

romaninsh:
i checked the PR and I can't immediatelly think of a reason why this could happen. Looks like tag not closed.

acicovic:
This PR has been reverted so I'm closing the issue

DarkSide666:
Related issues: #755, #781
I still can't think of reason why this change could possibly change HTML markup. So I don't think reverting this was good choice. Probably issue is somewhere else - maybe getTag method or something?

acicovic:
Reverting this did solve the issue but I wonder about that as well.

acicovic:
Important finding: this issue occurs for textareas which try to fill from a null value in database. Are you able to reproduce?

DarkSide666:
Will be fixed by https://github.com/atk4/ui/pull/783/commits/8ada11c8f068dade02a19aedaec78b0315159d96

DarkSide666:
[...]

PR #783, opened 2019-07-23
Feature/fix textarea

DarkSide666:
Related: #755, #760, #781 

Needs testing and try to reproduce HTML markup bug we faced before. I couldn't find how to reproduce that.

Tests are in /demos/form.php


romaninsh:
nevermind i'll add test..

tests/FormTest.php
```diff
@@ -43,35 +85,96 @@ public function testFormSubmit()
...
+            // dropdown validates to make sure option is proper
+            $this->assertFieldError('opt1', 'not one of the allowed values');
+
+            // user didn't select any option here
+            $this->assertFieldNoErrors('opt2');
+
+            // dropdown insists for value to be there
+            $this->assertFieldError('opt3', 'Must not be null');
+
+            // value with '0' is valid selection
```
romaninsh: @DarkSide666 - here is a test for dropdown value with key '0' which currently is considered invalid incorrect selection.

romaninsh:
@DarkSide666 i have added a test-suite, but it still not working with `'values'=>['0'=>'Female']`

acicovic:
Unfortunately the HTML markup bug (#760) reoccurs for me with this code. @romaninsh is everything OK on your side?

romaninsh:
@acicovic - i haven't tested against the project actually, have been a bit busy. @DarkSide666 can you see if saasty admin works OK (it uses mastercrud heavily)

acicovic:
For me this resolves #755 and #760. Doesn't fix #781 but this is not related to textarea per se, so I propose to merge this and we can tackle #781 afterwards.

DarkSide666:
fix #755 
fix #760

PR #784, opened 2019-07-23
Feature/Add Form support for containsMany field using MultiLine

ibelar:
### todo

- [x] update atk4/data/Reference/ContainsMany with new ui and system = false property
  As of now, field using containsMany need to be explicitly set using system property to false in order to appear in form.
`$this->containsMany('Addresses', [Address::class, 'system' => false]); `,
- [x] update for use with unloaded model
 This code in Form is causing issue: ` $model = $f->reference->refModel();` because when trying to create containsMany reference model on a new record, ContainsMany::getDefaultPersistence($model) in atk4/data throws exceptions: 'model should be loaded.'
```
if ($f->type === 'array') {
            $limit = ($f->reference instanceof ContainsMany) ? 0 : 1;
            $model = $f->reference->refModel();
            $fallback_seed = ['MultiLine', 'model' => $model, 'rowLimit' => $limit, 'caption' => $model->getModelCaption()];
        }
```
- [x] decide if we keep __atkml value in containsMany record - __atkml is needed for multiline component in order to manage error properly.
- [x] decide if we include Semantic-UI vue in atkjs package. As of now, if we want to have Multiline works in CRUD or in a form inside a modal, we have to explicitly load semantic-ui Vue package.
```
$app->useSuiVue();
$app->add(['CRUD'])->setModel($m);
```

### Update

- add support for limiting number of record. 
  ex: `$f->addField('test', ['MultiLine', 'model'=> Model::class, 'rowLimit' => 2])`
  Therefore containsOne will automatically use MultiLine using limit of 1 record.

- add support for Dropdown field in Multiline

Screen Shot 2019-07-24 at 5 39 36 PM

- add model caption (see image below)

- add support for Dropdown options via 'ui' => 'multiline' property
`        $this->hasOne('country_id', [new Country(), 'ui' => ['multiline' => ['dropdown' => ['search' => true]]]]);`

### usage:

```
class Client extends atk4\data\Model
{
    public $table = 'client';
    public $caption = 'Client';

    public function init()
    {
        parent::init();
        $this->addField('name');

        //use system = false until atk4\data is update.
        $this->containsMany('Addresses', [Address::class, 'system' => false]); 
    }
}

class Address extends atk4\data\Model
{
    public $caption = 'Address';

    public function init()
    {
        parent::init();
        $this->addField('street', ['required' => true]);
        $this->addField('city');
        $this->addField('country');
        $this->addField('postal_code');
    }
}

$f = $app->add('Form');
$f->setModel(new Client($db));
```
Screen Shot 2019-07-25 at 9 31 22 AM


src/Form.php
```diff
@@ -447,7 +450,11 @@ public function decoratorFactory(\atk4\data\Field $f, $seed = [])
 
         $fallback_seed = ['Line'];
 
-        if ($f->type != 'boolean') {
+        if ($f->type === 'array') {
+            $limit = ($f->reference instanceof ContainsMany) ? 0 : 1;
```
PhilippGrashoff: Setting the row limit to 1 is meant for ContainsOne, not ContainsMany if I am not mistaken

ibelar: > Setting the row limit to 1 is meant for ContainsOne, not ContainsMany if I am not mistaken

Because containsMany is a child of containsOne, I am testing for containsMany to set limit to 0 if true, i.e. no limit.

PhilippGrashoff:
Hi there,

first, thanks for all this work!

Here is a screen of a real life usage:
multiline_containsman

2 times ContainsMany, and two Multiline Form components added automatically.

2 Questions:
1) This feature does not use the "normal" ATK way to render the input fields. The second containsMany Model in the pic should have 2 Dropdowns:
```
        $this->addFields([
            ['weekday',            'type' => 'enum',    'caption' => 'Wochentag',                          'ui' => ['form' => ['DropDown', 'values' => STD_GERMAN_WEEKDAYS]]],
            ['value',              'type' => 'money',   'caption' => '+- Rabatt',      'default' => 0,     'ui' => ['form' => ['Line']]],
            ['scope',              'type' => 'string',  'caption' => 'Anwenden auf',   'default' => 'ALL', 'ui' => ['form' => ['DropDown', 'values' => ['ALL' => 'Vollzahler und Reduzierte', 'price_payfull' => 'Nur Vollzahler', 'price_payreduced' => 'Nur Reduzierte']]]],
        ]);
```

As my lack ok JS skills makes my brain hurt trying to understand all the JS code, I skip understanding and continue asking: Would it be much work having the standard way of ATK defining and rendering the inputs? Having one way throughout the project.


2) I notice that when changing a field value inside the Multiline Form, the `+` Icon shows a loading icon, as if it was saving. But at least here, it does not do anything. A normal Form save also does not save any data inside the Multiline.
Do I have to add anything to the Form->onSubmit() callback?

Thanks in advance, tell me where I can step in and help (not JS for the sake of us all :) )

src/FormField/MultiLine.php
```diff
@@ -517,13 +546,60 @@ public function setModel(\atk4\data\Model $model, $fields = [], $modelRef = null
                 'isEditable'  => $field->isEditable(),
                 'isHidden'    => $field->isHidden(),
                 'isVisible'   => $field->isVisible(),
-                'width'       => $width,
+                'values'      => $enumValues,
```
PhilippGrashoff: If I get it correctly, for a DropDown in 'values' should be the Dropdown options. This at the moment only works for Fields with `enum` or `values` property set (or references like hasOne). If its set in Fields 'ui' property (e.g. `'ui' => ['form' => ['DropDown', 'values' => [1 => 'Yes', 2 => 'Even more Yes']]]`, its not respected.

src/FormField/MultiLine.php
```diff
@@ -496,18 +520,23 @@ public function setModel(\atk4\data\Model $model, $fields = [], $modelRef = null
...
             }
-            $type = $field->type ? $field->type : 'string';
+
+            $type = $this->getFieldType($field);
+            if ($type === 'enum') {
+                $enumValues = $this->getFieldEnumValues($field);
+            }
 
             if (isset($field->ui['form'])) {
                 $type = $field->ui['form'][0];
```
PhilippGrashoff: If I get it correctly, `type` tells the JS how to render the input (Line, DropDown, TextArea etc). As of now the values of `type` are a mix of atk4 Data Field Types (string, integer, text etc). and Atk UI FormField types (DropDown, Line etc).
As its about rendering UI inputs, maybe we should be consistent with Atk UI FormField Types and always pass only these values (Using Line as Default)?

PhilippGrashoff:
As I am not sure I understood everything the way it was meant, I will post changes here as a comment: I made a more concise version of value and type getting for the input fields, where things defined in field's `ui` property take precedence (as its about rendering UI fields). Have a look:

```
foreach ($this->rowFields as $fieldName) {
            $field = $m->getField($fieldName);

            if (!$field instanceof \atk4\data\Field) {
                continue;
            }

            $options = $field->ui['multiline'] ?? null;

            $this->fieldDefs[] = [
                'field'       => $field->short_name,
                'type'        => $this->getFieldType($field),
                'caption'     => $field->getCaption(),
                'default'     => $field->default,
                'isEditable'  => $field->isEditable(),
                'isHidden'    => $field->isHidden(),
                'isVisible'   => $field->isVisible(),
                'values'      => $this->getFieldEnumValues($field),
                'fieldOptions'=> $options,
            ];
        }

        return $m;
    }

    /**
     * Return field type to use in multiline component.
     *
     * @param $field
     *
     * @return string
     */
    public function getFieldType($field)
    {
        $type = 'string';
        if (isset($field->ui['form'])) {
            $type = $field->ui['form'][0];
        } else if ($field->type) {
            $type = $field->type;
        } elseif ($field->enum || $field->reference || $field->values) {
            $type = 'enum';
        }

        return $type;
    }

    /**
     * Get value for enum field type.
     *
     * @param $field
     *
     * @return array|false|null
     */
    public function getFieldEnumValues($field)
    {
        $values = null;
        if ($field->enum) {
            $values = array_combine($field->enum, $field->enum);
        }
        else if(isset($field->ui['form']['values'])) {
            $values = $field->ui['form']['values'];
        }
        else if($field->values) {
            $values = $field->values;
        }
        elseif ($field->reference) {
            $m = $field->reference->refModel()->setLimit($this->enumLimit);

            $values = [];
            foreach ($m->export([$m->id_field, $m->title_field]) as $item) {
                $values[$item[$m->id_field]] = $item[$m->title_field];
            }
        }

        return $values;
    }
```

ibelar:
@PhilippGrashoff  - Change proposal for determining proper field type.

Multiline component are based on [Semantic UI Vue](https://semantic-ui-vue.github.io/#/).

Semantic UI Vue uses Vue components to render html input. A regular input component (sui-input) uses Vue properties (props) to render itself and setup it's html type attribute. (type="text"). This allow to pass for other html type attribute value like number, email, password etc. When it comes to Dropdown, Checkbox or Rating inputs they all use their own vue component, like sui-dropdown for Dropdown or sui-checkbox for Checkbox which are based on Semantic UI module. 

The ATK Multiline component need to know which proper SUI Vue component to render, i.e. sui-input or sui-checkbox or sui-dropdown (eventually other component like sui-rating or even or own vue component). Furthermore, when sui-input is used, it needs to know which html attribute type to applied for that input, like text, number or password. Finally, you can also supply component options via props for any of the components. For example, beside type for sui-input, you can supply other options like icon, disabled or even values for sui-dropdown.

Here is the proposal. When it come to set input type for Multiline, I think we should use a multiline option as the ui property and pass what needs to be render by the ATK multiline component. 

It could be set at the model level:
```
$this->addField('name', 'ui' => 
    [
    'multiline' => [
        'input', 
        [
            'type' => 'text',  //the html type attributes
            'options' => $name_options // icon, disabled, icon position etc.
       ]
    ]
);

$this->addField('password', 'ui' => 
    [
        'multiline' => [
            'input',
            ['type' => 'password']  //the html type attributes
    ]
);
```

This would allow to pass any supported html type attribute to each multiline input and would be independent of the form ui rendering.

For dropdown it could look like this:

```
$this->addField('choices', ['values' => ['street' => 'Street', 'blv' => 'Blv', 'road' => 'Road']], 'ui' => 
    [
        'multiline' => [
            'dropdown',
            ['options' => $dropdown_options] // sui-dropdown options
    ]
);
```

Of course, when multiline is not set, then default are applied according to model field type.

Let me know your thoughts.

src/FormField/MultiLine.php
ibelar: Agree with you that this is somehow confusing. See comment below for better approach.

PhilippGrashoff:
Hi,

now this sounds sensible to me :) I can work on the server side stuff.

So, regarding `type`:
1) Must be one of
- input
- dropdown
- checkbox
- textarea
- are there others?

2) To determine type, the following logic is used:
- if `$field->ui['multiline'] `is set, if so, use this
- if `$field->ui['form']` is set, use this (Translation might be needed, see below)
- else use standard for `field->type `(Translation, see below)
- if field->type isnt set, use 'input'

3) For this to work Atk UI FormField classes as well as ATK Data Field types need to be 'translated' into the field types available in 1. Part of this is done already I see.

I will start working on this, if you like we can skype, too.

ibelar:
## Sample usage with latest changes.

```
class Client extends atk4\data\Model
{
    public $table = 'client';
    public $caption = 'Client';

    public function init()
    {
        parent::init();

        $data = [];

        $this->addField('name');
        $this->containsMany('Accounts', [Account::class, 'system' => false]);
}

class Account extends atk4\data\Model
{
    public $caption = 'Account';

    public function init()
    {
        parent::init();

        $this->addField('email', ['required' => true, 'ui' => ['multiline' => ['input', ['icon' => 'envelope', 'type' => 'email']]]]);
        $this->addField('password', ['required' => true, 'ui' => ['multiline' => ['input', ['icon' => 'key', 'type' => 'password']]]]);
        $this->addField('site', ['required' => true]);
        $this->addField('type', ['default' => 'user', 'values' => ['user' => 'Regular User', 'admin' => 'System Admin'], 'ui' => ['multiline' => ['width' => 'four']]]);
    }
}

$m = new Client($db);
$m->loadAny();
$f = $app->add('Form');
$f->setModel($m);
```

Screen Shot 2019-08-16 at 10 03 37 AM

romaninsh:
What's the status of this?

ibelar:
> What's the status of this?

@romaninsh - All is ready on UI side. Waiting for atk4/data to be changed before we can merged.

acicovic:
So what must be done next? Is this tracked somewhere in an issue or PR in the data repository?

DarkSide666:
Please review/merge https://github.com/atk4/data/pull/459 and then data issue should be fixed.

PhilippGrashoff:
As data changes are merged, we could merge this one?

PR #789, opened 2019-07-31
Fix #786 - Exception display in app and modals

### LINKED ISSUE #786: Problematic exceptions display
Hi,

In latest UI development release, exceptions are displayed in the bottom of the page instead of the top like it used to. So we have to scroll down to see the message which is encumbering.

In addition, exceptions displayed in a modal are not displayed correctly as they output HTML code and have no formatting.

acicovic:
Currently ATK UI uses static() in caughtException() in order to display the exception within the app's environment. However, this can lead to complications like those:

And in modals:

This PR addresses the issue by creating a new App() to display the exception and also adds the line and exception title in the page's title which can be practical when juggling with exceptions.

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -241,10 +241,11 @@ class App
     {
         $this->catch_runaway_callbacks = false;
 
-        // must be static to support extended App
-        // if not the App will use default value
-        // e.g. Title = 'Agile UI - Untitled Application'
-        $l = new static();
+        // Use new App() instead of static() to prevent broken exception
+        // message display due to conflict with existing layout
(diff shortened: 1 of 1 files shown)


acicovic:
If this stays as-is, we can also probably remove `$l->catch_runaway_callbacks = false;` in lines 276/277

abbadon1334:
> If this stays as-is, we can also probably remove `$l->catch_runaway_callbacks = false;` in lines 276/277

can be right, but probably if it was done conditionally there is a deeper motive.

PR #790, opened 2019-08-02
Remove calls to exit functions and cleanups

abbadon1334:
i removed calls to php exit function, that create many problem and not allow to include the atk in a any other framework that use PSR-7 with request -> response flow.

what i have done : 
- i removed all "generalized" try catch, any Error or Exception not strictly related to the flow must be throw upper to the App->catchException.
- add some throw to methods.
- simplify methods when i see some old syntax.
- i try to commit on every change to have comments on the work in progress.
- PHPUnit on my PC pass all tests
- Manual testing to be sure all works correctly

I had only one problem in : https://github.com/abbadon1334/ui/blob/18ce128a68ad1785f0d33e09f584bc1d77992de8/src/Form.php#L592

Where in Form.php we return "Direct output" as an error, and i normalized it to Exception.


src/App.php
```diff
@@ -234,11 +232,18 @@ public function callExit($from_shutdown = false)
...
      *
      * @return bool
      */
     protected function caughtException(Throwable $exception)
     {
+        // if the intent is to exiut from application
+        // catch Exit Exception and return
+        if ($exception instanceof ExitApplicationException) {
+            return true;
+        }
```
romaninsh: clever approach to avoid error.

But it's different to what we dicsussed and i think this can cause problem if you run app with "catch_exception = false".

Also i think it will create some insanity if used with logger (which normally log all unhandled exceptions)

Best to implement the property App::call_exit = true , which can be modified to "false" if exception is required.

abbadon1334:
i will check back tomorrow for coverage, now i need to go.

abbadon1334:
i try to added to request/response check for demos and coverage, the new actions from demos/actions.php

Local it works perfectly, on travis if i call the url POST for confirm in download popup action or the url POST for submit the form in edit popup for record 1, i get an error of record not found, which is wrong because i call first the action to importFromFileSystem which will populate the database.

I don't understand what is wrong on travis.

Probably i will divide this PR in smaller PR

src/App.php
```diff
@@ -234,52 +236,53 @@ public function callExit($from_shutdown = false)
...
 
-        // Use new App() instead of static() to prevent broken exception
-        // message display due to conflict with existing layout
-        $l = new self();
-        $l->title = 'L'.$exception->getLine().': '.$exception->getMessage();
-        $l->catch_runaway_callbacks = false; // Allow exceptions within modals
-        $l->initLayout('Centered');
+        // just replace layout to avoid any extended App->_construct problems
+        // it will maintain everything as in the original app StickyGet, logger, Events
+        $this->html = null;
```
abbadon1334: in place of creating a new app for exception, why not destroy the render tree making $app->html = null
and reuse the actual app, all $defaults will be respected.

src/App.php
```diff
@@ -320,12 +330,18 @@ public function outputDebug($str)
...
      * @throws \atk4\core\Exception
+     * @throws ExitApplicationException
      */
     public function terminate($output = null)
     {
         if ($output !== null) {
-            echo $output;
+            if ($this->isJsonRequest()) {
+                $this->outputResponseJSON($output);
+            } else {
```
abbadon1334: any response from atk4 is text/html, i made 2 function to respond with JSON or HTML headers
i have also made a outputResponse with custom headers for any special needs (SEE?)

src/App.php
```diff
@@ -434,29 +450,49 @@ public function add($seed, $region = null)
...
-        if (isset($_GET['__atk_callback']) && $this->catch_runaway_callbacks) {
-            $this->terminate('!! Callback requested, but never reached. You may be missing some arguments in '.$_SERVER['REQUEST_URI']);
+            if (isset($_GET['__atk_callback']) && $this->catch_runaway_callbacks) {
+                $this->terminate(
+                    '!! Callback requested, but never reached. You may be missing some arguments in '.$_SERVER['REQUEST_URI']
+                );
+            }
+            echo $this->html->template->render();
+        } catch (ExitApplicationException $e) {
+            $is_exit_exception = true;
```
abbadon1334: wrap all App->run i a try catch that catch only exit application exception and call exit app if raised

src/Form.php
```diff
@@ -595,7 +595,7 @@ public function ajaxSubmit()
                 ob_start();
                 $this->loadPOST();
                 $response = $this->hook('submit');
-                $output = ob_get_contents();
+                $output = ob_get_clean();
```
abbadon1334: i raise exception to the upper level to be caught by the global exception handler

src/Form.php
```diff
@@ -616,18 +616,13 @@ public function ajaxSubmit()
                     return new jsExpression('console.log([])', ['Form submission is not handled']);
                 }
             } catch (\atk4\data\ValidationException $val) {
+                ob_get_clean(); // not close output buffer on exceptions
```
abbadon1334: output buffer
if the code in try/catch not raise exception it will exit from return $response with buffer cleaned.
if an exception is raised the old finally will not be executed leaving an output buffer open.
i remove the finally ( even more because the other catches has a return.
i add a ob_get_clean which can be ob_clean when exception is raised and changed ob_get_contents to ob_get_clean directly in try.

src/View.php
```diff
@@ -749,37 +770,22 @@ public function render($force_echo = true)
...
-                $l->template->setHTML('Content', $exception->getHTML());
-            } elseif ($exception instanceof \Error) {
-                $l->add(new self(['ui' => 'message', get_class($exception).': '.
-                                                            $exception->getMessage().' (in '.$exception->getFile().':'.$exception->getLine().')',
-                    'error', ]));
-                $l->add(new Text())->set(nl2br($exception->getTraceAsString()));
-            } else {
-                $l->add(new self(['ui' => 'message', get_class($exception).': '.$exception->getMessage(), 'error']));
-            }
+        $this->renderAll();
```
abbadon1334: raise exception to upper level
we already have exception handler that format correctly the exception, if an exception happen here will broke the normal flow of the whole application.

src/View.php
```diff
@@ -1236,20 +1245,13 @@ public function _getStickyArgs($triggerBy)
      * @param string $name
      * @param string $newValue
      *
-     * @return string
+     * @return null|string
      */
-    public function stickyGet($name, $newValue = null)
+    public function stickyGet($name, $newValue = null) : ?string
```
abbadon1334: just simplify this

tests/BuiltInWebServerAbstract.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+namespace atk4\ui\tests;
+
+use GuzzleHttp\Client;
+use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\Process\Process;
+
+abstract class BuiltInWebServerAbstract extends TestCase
```
abbadon1334: add alternative base Test class to test real GET POST request using built in server

tools/test.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 echo "Setting up coverage logging";
-mkdir coverage
+mkdir -p coverage
```
abbadon1334: add -p to mkdir to avoid errors on call to mkdir 
`-p, --parents     no error if existing, make parent directories as needed`

tests/BuiltInWebServerAbstract.php
```diff
@@ -0,0 +1,81 @@
...
+    protected static $app_def_call_exit = true;
+
+    /** @var bool set the app->caught_exception in demo */
+    protected static $app_def_caught_exception = true;
+
+    protected static $webserver_root = 'demos/';
+
+    public static function setUpBeforeClass()
+    {
+        if (!file_exists(getcwd().'/coverage/')) {
```
abbadon1334: to get coverage on tests run with built in server i use the same idea that we already use with selenium, include file coverage.php and collect .cov files

PR #796, opened 2019-08-26
add "Is Not" for TableColumn::FilterModel::TypeString

funyx:
add "Is Not" option for TableColumn::FilterModel::TypeString

https://puu.sh/E9Rqy/5e2cdd5908.png


skondakov:
Now it all looks good to me. Thanks!

Issue #802, opened 2019-09-20
Newest Dev version: Array fields cause Error in Form


PhilippGrashoff:
Hi there,
the latest merged pr #784 causes some fields that are defined as array to crash the form. In my example the field was defined like this in model:
```
$this->addFields([
            ['created_records',         'type' => 'array',      'serialize' => 'json', 'default' => []],
```

That caused an Error in Form linr 455:
```
Error: Call to a member function refModel() on null
```

Reading the code:
```
 if ($f->type === 'array') {
            $limit = ($f->reference instanceof ContainsMany) ? 0 : 1;
            $model = $f->reference->refModel();
            $fallback_seed = ['MultiLine', 'model' => $model, 'rowLimit' => $limit, 'caption' => $model->getModelCaption()];
        }
```

So its assumed that whenever a field is defined as array, it has a reference set, which, of course, must not neccessarily be.

I fixed my issue and set the field to system=true so it wont appear in Form at all, but thats only a workaround.

PhilippGrashoff:
fixed by #804

Issue #752, opened 2019-06-21
add ['type'=>'array'] support


romaninsh:
Would show multiple lines on the form:

<img width="1095" alt="Screenshot 2019-06-21 at 15 07 16" src="https://user-images.githubusercontent.com/453929/59928255-43506d80-9436-11e9-9978-c37f163a8a27.png">

mvorisek:
closing in favor of https://github.com/atk4/ui/issues/1860

PR #804, opened 2019-09-20
fix/#802 Array fields error in form

ibelar:
Fix for #802 

Make sure reference is set when creating a Multiline in form using type array.


PhilippGrashoff (approved):
LGTM. As someday there is another array field support planned (https://github.com/atk4/ui/issues/752), some more has to happen here, but thats for another PR adressing #752

acicovic:
@PhilippGrashoff you didn't merge this but I see no other reviewers. You think it should be reviewed by someone else before merging?

PR #806, opened 2019-09-27
Fix #805

### LINKED ISSUE #805: Error in FilterColumn when using Text Type in model
When you use 'type'=>'text' in the model, the addFilterColumn in Grid is failing, because a definition for FilterModel/TypeText is missing.

karakal:
#Fix of `Error in addFilterColumn` (see ticket #805 ).

### DIFF
--- a/src/TableColumn/FilterModel/TypeText.php
+++ b/src/TableColumn/FilterModel/TypeText.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace atk4\ui\TableColumn\FilterModel;
+
+class TypeText extends TypeString
+{
+
+}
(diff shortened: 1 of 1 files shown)


acicovic:
@abbadon1334 that was my thought also, but after testing I see that it doesn't behave exactly like TypeString as the column expands at maximum width. So the exception is indeed fixed but there should be an additional fix to make it work like string. I suspect the factoryType() function within FilterModel\Generic. Will take a look at it if I have time. @karakal if you have the time to spot and fix this, go ahead.

acicovic:
OK, so this is by design. ATK assigns an `atk-cell-expanded` class in TableColumn\Text.php.

PR #808, opened 2019-10-02
Fix unclickable Autocomplete/Lookup fields

acicovic:
In Autocomplete and Lookup controls, a large area of the control is unclickable, showing a hand cursor. This makes those controls effectively unusable.

image

The issue is reproducible in the current demo (for example https://ui.agiletoolkit.org/demos/autocomplete.php - hovering over Country1/2/3 towards the right will produce the hand cursor and unclickable control).

This PR fixes the issue for both controls.


acicovic:
@ibelar updated pug files.

PR #819, opened 2019-10-21
Add Locale class

abbadon1334:
Add Locale class to find default localization definitions


abbadon1334:
let's merge this and add another for start translating prefixed with WIP, few more busy days and i will start it

PR #832, opened 2019-10-30
Feature/column user actions

romaninsh:
Now grid's addAction supports a single-argument form and you can pass name of the action or UserAction object from the model.

This will add a button triggering the action using our brand new action executor!

```
$grid->addAction('test');
```


romaninsh:
@ibelar - can you check why tests fail? and after we can merge and focus on CRUD. Veeeery close now!

ibelar:
@romaninsh - test is fix and ready to merge.

PR #834, opened 2019-11-07
DateTime is object and working on same object is replacing both value…

gowrav-vishwakarma:
This sets $d1 and $d2 both to "YYYY-mm-dd 23:59:59" ...

Filter for Exact does not works as both dates are set to 23:59:59

### DIFF
--- a/src/TableColumn/FilterModel/TypeDatetime.php
+++ b/src/TableColumn/FilterModel/TypeDatetime.php
@@ -92,7 +92,7 @@ class TypeDatetime extends Generic
                     break;
                 case '!=':
                 case '=':
-                    $d1 = $this->getDatetime($filter['value'])->setTime(0, 0, 0);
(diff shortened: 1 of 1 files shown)


src/TableColumn/FilterModel/TypeDatetime.php
```diff
@@ -92,7 +92,7 @@ public function setConditionForModel($m)
                     break;
                 case '!=':
                 case '=':
-                    $d1 = $this->getDatetime($filter['value'])->setTime(0, 0, 0);
+                    $d1 = clone $this->getDatetime($filter['value'])->setTime(0, 0, 0);
                     $d2 = $this->getDatetime($filter['value'])->setTime(23, 59, 59);
                     if ($d2 >= $d1) {
```
PhilippGrashoff: Reading through this part if the code, to me it seems this condition is always true. I cant figure how the else part could possibly be reached. As the same parameter is passed to getDateTime(), $d2 will always be greater then $d1.

PR #848, opened 2019-11-26
Add Github workflows (actions) 

romaninsh:


abbadon1334:
Errors in demo/actions.php come from the following error :

one of the error comes from declaring $fields as array but type hint as array or null : https://github.com/atk4/data/blob/6c93501cda8b59acb8cd9dbad0f6d3e7d4f92097/src/UserAction/Generic.php#L71

and later check it for boolean : https://github.com/atk4/data/blob/6c93501cda8b59acb8cd9dbad0f6d3e7d4f92097/src/UserAction/Generic.php#L114

@romaninsh I was not in the Actions development, can you check it?

PR #847, opened 2019-11-26
Fix few bugs and add comments

DarkSide666:
Fix few bugs and add comments

PR #851, opened 2019-12-01
Fix/skip adding to grid menu if no menu

georgehristov:
Do not add `NO_RECORD` actions to menu if Grid menu is disabled


ibelar:
Thks @georgehristov for catching it - I will just blocked it for now because it was address in #847 with many other fixes.

ibelar:
issue was address in #847

PR #852, opened 2019-12-01
Improvement/include action caption in modal title

georgehristov:
improvement


georgehristov:
Added trim around the modal title

PR #857, opened 2019-12-03
Added namespaces to quickstart and file structure example

bedengler:
I adapted and enhanced the things that have been missing for myself when I started using atk. So I first had to dig into namespaces and second I had to experiment with the file structure until I came up with one that is suited for small and big projects and offers a good overview and structure.

Useful for beginners that have first contact with atk and aren't used to PSR-4 namespaces and have just low level skills in dealing with Composer.

### DIFF
--- a/docs/filestructure.rst
+++ b/docs/filestructure.rst
@@ -21,45 +21,52 @@ Feel free to experiment with it and find the ideal file structure for your proje
 
 * public_html
 
-    * vendor (contains all needed composer modules - don't touch them)
+    * images
 
-        * atk4
+    * index.php
 
-            * ...
+    * init.php
 
-        * autoload.php
+    * admin.php (if needed)
+
+* projectfolder (could be named "app" for example)
+
+    * Forms
+
+        * ExampleForm1.php
+
+        * ExampleForm2.php
+
+        * UserDetailForm.php
+
+    * Models
+
+        * ExampleClass1.php
+
+        * ExampleClass2.php
+
+        * LoadAllUsers.php
+
+    * Views
+
+        * View1.php
+
+        * View2.php
+
+        * GridUserList.php
+
+* vendor (contains all needed composer modules - don't touch them)
+
+    * atk4
 
         * ...
 
-    * projectfolder
+    * autoload.php
 
-        * forms
+    * ...
 
-            * exampleForm1.php
+* composer.json
 
-            * exampleForm2.php
-
-        * models
-
-            * exampleClass1.php
-
-            * exampleClass2.php
-
-        * views
-
-            * View1.php
-
-            * View2.php
-
-        * images
-
-        * index.php
-
-        * init.php
-
-        * admin.php (if needed)
-
-    * composer.json
 
 
 Composer configuration
@@ -74,7 +81,7 @@ Your composer.json could look like this::
     },
     "autoload": {
       "psr-4": {
-          "myProject\\": "projectfolder/"
+          "MyProject\\": "projectfolder/"
       }
     }
   }
@@ -89,7 +96,7 @@ Read below how to load the autoload.php into your project.
 The "require" section within composer.json loads plublically available composer packages from the server.
 
 The "autoload" section within composer.json loads your individual project files (that are saved locally on your computer).
-"myProject" defines the namespace you are using in your classes later on.
+"MyProject" defines the namespace you are using in your classes later on.
 
 
 Why "public_html"?
@@ -183,17 +190,17 @@ Following the PSR-4 specifiations all class names and file names have to corresp
 
 If we want to create a class called "myFirstClass" we have to save it to a file called "myFirstClass.php".
 
-Let's do our first class. Please create a new file in the directory "projectfolder/views" and call it "View1.php".
+Let's do our first class. Please create a new file in the directory "projectfolder/Views" and call it "View1.php".
 
 Now comes a tricky part: you have to define a namespace within your class file that corresponds with the namespace you have
 defined in your composer.json.
-Do you remember? - If no, take a look at the beginning of this document. We defined there "myProject" as our namespace for
+Do you remember? - If no, take a look at the beginning of this document. We defined there "MyProject" as our namespace for
 the directory "projectfolder".
 
 Open the created file "View1.php" in your editor and add the following lines::
 
   <?php
-  namespace myProject\views;
+  namespace MyProject\Views;
 
   class View1 extends \atk4\data\View {
       function init() {
@@ -203,8 +210,8 @@ Open the created file "View1.php" in your editor and add the following lines::
       }
   }
 
-"namespace myProject\\views;" defines the namespace to use. It reflects the folder structure of the app.
-The file located in "projectfolder/views/View1.php" becomes "myProject\\views\\View1" in the namespace.
+"namespace MyProject\\Views;" defines the namespace to use. It reflects the folder structure of the app.
+The file located in "projectfolder/Views/View1.php" becomes "MyProject\\Views\\View1" in the namespace.
 
 For each of your classes create a separate file. As long as you follow the name conventions all your class
 files will be autoloaded by Composer.
@@ -219,9 +226,9 @@ To use our class in our app, we have to include it into our app. This can be don
 
 Please add the following lines into your index.php::
 
-  $app->add(new myProject\views\View1());
+  $app->add(new MyProject\Views\View1());
 
-or if you have added at the beginning of your index.php "use myProject\\views\\View1;" you can write::
+or if you have added at the beginning of your index.php "use MyProject\\Views\\View1;" you can write::
 
   $app->add(new View1());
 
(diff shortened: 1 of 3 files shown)


docs/filestructure.rst
```diff
@@ -0,0 +1,228 @@
...
+This file structure is a recommendation and no must. It is a best practice example.
+Feel free to experiment with it and find the ideal file structure for your project.
+
+* config
+
+    * db.php
+
+* public_html
+
+    * vendor (contains all needed composer modules - don't touch them)
```
abbadon1334: vendor must be in root like config and never in public folder

docs/filestructure.rst
```diff
@@ -0,0 +1,228 @@
...
+
+        * atk4
+
+            * ...
+
+        * autoload.php
+
+        * ...
+
+    * projectfolder
```
abbadon1334: project the same as vendor all in root

docs/filestructure.rst
```diff
@@ -0,0 +1,228 @@
...
+
+        * forms
+
+            * exampleForm1.php
+
+            * exampleForm2.php
+
+        * models
+
+            * exampleClass1.php
```
abbadon1334: PSR-1+PSR-4 Class must be **StudlyCaps**
PSR-1+PSR-4 Method must be in **camelCase**

docs/filestructure.rst
```diff
@@ -0,0 +1,228 @@
...
+      "psr-4": {
+          "myProject\\": "projectfolder/"
+      }
+    }
+  }
+
+
+What does that mean?
+--------------------
+As soon as you start a "composer update" or "composer dump-autoload" in the public_html directory, all needed atk4 files
```
abbadon1334: vendor is out of public folder

docs/filestructure.rst
```diff
@@ -0,0 +1,228 @@
...
+
+            * View1.php
+
+            * View2.php
+
+        * images
+
+        * index.php
+
+        * init.php
```
abbadon1334: init.php can be called bootstrap.php and can load autoload.php + load config and set the chdir(__DIR__)

docs/filestructure.rst
```diff
@@ -0,0 +1,228 @@
...
+
+  class View1 extends \atk4\data\View {
+      function init() {
+          parent::init();
+
+          $text = $this->app->add(['Text', 'here goes some text']);
+      }
+  }
+
+"namespace myProject\\views;" defines the namespace to use. It reflects the folder structure of the app.
```
abbadon1334: PSR-1+PSR4 wrong naming

docs/filestructure.rst
```diff
@@ -0,0 +1,228 @@
...
+The file located in "projectfolder/views/View1.php" becomes "myProject\\views\\View1" in the namespace.
+
+For each of your classes create a separate file. As long as you follow the name conventions all your class
+files will be autoloaded by Composer.
+
+.. warning:: Keep in mind that as soon as you have created one or more new file(s) within the projectfolder you have to run "composer dump-autoload"!!! Otherwise the newly generated file(s) and classes will not be autoloaded and are therefore unavailable in your application.
```
abbadon1334: This are things related more to composer than atk, i don't know if we need to explain such things.

docs/filestructure.rst
```diff
@@ -0,0 +1,228 @@
...
+
+
+Load your class in index.php
+============================
+
+To use our class in our app, we have to include it into our app. This can be done either through index.php or admin.php.
+
+Please add the following lines into your index.php::
+
+  $app->add(new myProject\views\View1());
```
abbadon1334: you see if you don't follow naming rules namespaces and classes look strange and not normalized

docs/filestructure.rst
```diff
@@ -0,0 +1,228 @@
...
+
+or if you have added at the beginning of your index.php "use myProject\\views\\View1;" you can write::
+
+  $app->add(new View1());
```
abbadon1334: i think we need to find a case much real than naming View1, something like GridUserList can be better?

docs/filestructure.rst
abbadon1334: projectfolder probably can be changed to src or App, to be in line with others frameworks

docs/filestructure.rst
bedengler: Reason why it's in public_html is because I'm loading JS and CSS frameworks as well. These have to be accessible through the domain. Therefore vendor has to be in public folder in my case. If you just use atk4, you could move vendor outside of the public_html folder that's true. But I've tried to bring in a real world example...as the other examples mostly have a limited scope and are good for small projects with just a few repositories. 
But as your app grows, you need a more flexible structure imho.

Sure, you could have 2 Composer files, one in public and one hidden from public. But if a repository depends e.g. on PHPSpreadsheet it would be downloaded twice, one time in public vendor and one time in hidden vendor.
This is imho unnecessary...but I'm open for discussion of better practice ;)

docs/filestructure.rst
bedengler: I think naming should be as simple and as noob understandable as possible.
If something's called "app" I have to research what app means. 

I've investigated many frameworks and I'm annoyed by namings you have to research first. That's why I've chosen this simple approach...

Concerning vendor in root see above...

docs/filestructure.rst
abbadon1334: IMHO JS and CSS can be copied in *public/assets* folder in place of exposing all the vendor directory, but leaving apart the actual problem of needing of rebuild css and js, i think is better to leave the default assets that are defined in $app->cdn which are loaded by default

docs/filestructure.rst
bedengler: Good point! Will correct that, thanks...

docs/filestructure.rst
abbadon1334: IMHO i think if you define a folder app and in composer.json you put autoload psr-4 pointing there, all the logic must be there.

docs/filestructure.rst
bedengler: See my reply above...

docs/filestructure.rst
bedengler: I called it init.php, because it's called so in the demos of atk4...and again: init.php is in my opinion clearer than bootstrap.php (which also could mean the framework).

docs/filestructure.rst
bedengler: Even if you know that it's an information that can help people who are just starting...

docs/filestructure.rst
bedengler: It can be an addition, but I wanted to keep it as simple as possible to make people understand the way it works basically.

docs/filestructure.rst
bedengler: Default assets means there's no flexibility. What if you have a client that wants flexibility?
Just copying Composer assets to other directories clutters up everything. You can't maintain them properly and they can't be autoloaded. 
As an example: I'd like to use "select2" which replaces conventional select boxes.

docs/filestructure.rst
bedengler: Basically, but as said you first have to research in composer.json. So 1 step more ;)

docs/filestructure.rst
abbadon1334: IMHO if you want to use select2, YOU need to tweak the standard configuration and there is no point on adding complexity in a getting started project.

- app
- config
- - db.php
- models
- public
- - index.php
- vendor
- init.php
- composer.json

if you start to put some customization, can be disorienting


- app
- - User
- - - Grid.php
- - - Form.php
- - Profile.php
- - Login.php
- - Logout.php
- config
- - db.php
- models
- - User.php
- public
- - index.php
- - admin
- - - index.php
- vendor
- init.php
- composer.json

With admin can be, but already here there are many ways of defining a structure, for that i want to discuss in a meeting, because we need to give a path to follow, if not every project have is own structure and when you open a project you don't know where to look to find objects.

IMHO Not easy, but must be done.

bedengler: Yes you are right. 
That's why I have corrected the file structure in my latest contribution and moved vendor to the root.
In your example, why should we have 2 directories? One for app and one for models? Keep it simple!

It's good to have a basic structure that makes a good fundament for small AND big projects.

Here's another approach: 
* config
    * db.php
    * init.php

* public_html
    * images
    * index.php
    * admin.php (if needed)

* projectfolder (could be named "app" for example)
    * Forms
        * ExampleForm1.php
        * ExampleForm2.php
        * UserDetailForm.php

    * Models
        * ExampleClass1.php
        * ExampleClass2.php
        * LoadAllUsers.php

    * Views
        * View1.php
        * View2.php
        * GridUserList.php

* vendor (contains all needed composer modules - don't touch them)
    * atk4
        * ...
    * autoload.php
    * ...

* composer.json

docs/filestructure.rst
bedengler: corrected...

romaninsh:
@abbadon1334 waiting for your review, but I think change is good!!

PR #869, opened 2019-12-26
fomantic update 2.8.3

romaninsh:
bumping version of Fomantic

See https://github.com/atk4/ui/pull/979 instead


romaninsh:
Looks like this is breaking our forms?

acicovic:
Hi, won't be able to review this until mid-January unfortunately.

PR #873, opened 2019-12-26
Add workflow for deploying "ui.agiletoolkit.org" automatically.

romaninsh:
Will create ui.agiletoolkit.org automatically.


mvorisek:
@romaninsh Does it work?

PR #884, opened 2020-01-18
set model on initialization

georgehristov:
If model is included in the `Grid` seed array the `setModel` method will me executed.
This allows for injecting the model directly using the seed definition and not call `setModel` explicitly.


romaninsh:
If we move setModel() higher on a global level, we would need to make sure that all the views actually can function properly if setModel() is to be executed from constructor. 

The next thing would be then to move everything else into seed and we end up with Yii ;)

Might not be that bad of a thing.

We could add:

``` php
if ($this->model) {
    $this->setModel($this->model);
}
```

inside View::init() and I hope that won't disrupt things too much. 2.0 probably would be a much better place to   do it rather than 2.4 later ;) 


@georgehristov  - can you open PR to address this on View level as Imants suggests?

georgehristov:
If we are to generalize I may suggest to consider and go even further calling a magic setProperty method for all seed properties if method available. E.g. 'model' -> 'setModel', 'title' -> 'setTitle', etc?

PR #888, opened 2020-01-21
Fix missing EOLs in text files

mvorisek:
Fix CS, important as many editors add the final EOL on save and it is very hard to work with the current text files.


mvorisek:
Tests seem to pass, there is an issue with the tests itself: https://github.com/atk4/ui/issues/889

mvorisek:
@DarkSide666 End NL newly removed by https://github.com/atk4/ui/commit/5adec249dafe70198f83220e2835491e4d172cc1 - can you update you editor to always put NL on the end of any text file - this is industry standard - and enforce this by:

- https://docs.styleci.io/fixers#single_blank_line_at_eof
- https://docs.styleci.io/fixers#unix_line_endings

PR #892, opened 2020-01-22
usage of getFields() and fix Card

DarkSide666:


src/Card.php
```diff
@@ -209,7 +209,7 @@ public function setModel(Model $m, $fields = null)
         }
 
         if (!$fields) {
-            $fields = array_keys($this->model->getFields('visible'));
+            $fields = array_keys($this->model->getFields(['editable', 'visible']));
```
ibelar: @DarkSide666 - Why do we need to restrict display of field value to editable fields in Card?  What if a non editable field needs to be display? Will that prevent it to show in Card?

DarkSide666: It now will show not only `visible` fields but also `editable` fields (but will of course not allow to edit them because it's just a readonly card view).
Otherwise currently it only shows `visible` fields, but if we have some field which we don't want to show in Grid as column (not visible), but show in form (is editable), then in Card view these fields was not visible and i think in case of card view it's wrong (for example for mastercrud which uses card view).

ibelar: Ok make sense.

PR #894, opened 2020-01-26
Fix/Enable Callback argument to stick at application

ibelar:
Sometimes, callback argument need to stick application-wide, not just in View.
This is especially true for ActionExecutor using application modal.

This PR allows us to create a URL argument that stick to application.


ibelar:
This is creating reload problem in Saasty. Found another solution in Saasty directly.

ibelar:
Still need App wide callback argument under certain circumstance.

PR #896, opened 2020-01-27
Fix/UserAction Validation error

ibelar:
Properly catch Data Validation error in UserAction executor.


DarkSide666:
Not sure what it does and why.

ibelar:
> Not sure what it does and why.

This is to give better information on error by including the action name in error message. 
The fix allows for validation error to get handle by form and display field error.

PR #900, opened 2020-01-31
Fix/Table Popup - Dropdown

### LINKED ISSUE #554: Column filter going on second row
If Grid column content is shorter than column name, filter icon go on the next row
screen shot 2018-09-03 at 14 28 04 1

### LINKED ISSUE #897: Column filtering does not work in Tabs
When a grid is in a tab, the column filtering does not work. It works just fine when not in a tab.
This is using the 2.0.x versions of agiletoolkit.

ibelar:
Fixes:
- Allow table filter to appear in Tabs fix #897 
- Better display of filtering icon fix #554 

Enhancement:
- Display column header name in a div using CSS class named "atk-table-column-header"
- Display sort icon on right side of column header
- Display column filtering icon under the column name
- allow to change filtering icon via CSS only 

Screen Shot 2020-01-31 at 11 44 41 AM

### DIFF
--- a/demos/tablecolumnmenu.php
+++ b/demos/tablecolumnmenu.php
@@ -6,7 +6,9 @@ require 'database.php';
 
 $app->add(['Header', 'Table column may contains popup or dropdown menu.']);
 
-$table = $app->add(['Table', 'celled' => true]);
+// Better Popup positionning when Popup are inside a container.
+$container = $app->add(['ui' => 'vertical segment']);
+$table = $container->add(['Table', 'celled' => true]);
 $table->setModel(new SomeData(), false);
 
 //will add popup to this column.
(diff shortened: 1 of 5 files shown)


mvorisek:
@ibelar Please see issue #932, this PR is merged, but it breaks tests as in the `tablefilter.php` page the filters were effectively removed:

Before this PR:
image

After this PR:
image

ibelar:
@mvorisek - this pr rely on new css style. Please make sure you include it. You may have to build it from agileui.less file and load it locally until the next release.

mvorisek:
> @mvorisek - this pr rely on new css style. Please make sure you include it. You may have to build it from agileui.less file and load it locally until the next release.

That is exactly the point why I opened issue #940 yesterday. I came to it just by a look at your PR (as there was no css change), never realized that this actually solves this issue. Thanks.

mvorisek:
@ibelar tested, with rebuilded css/js filters work, but the filter arrow/icon is now always at a new line - related with issue #554 which I reopened - can you fix it?

PR #901, opened 2020-02-02
Update string quotes to make the code compatible with Netbeans

mvorisek:
Change string quotes from double quotes to single quotes as the double quotes with current complex content is not parsed by Netbeans correctly. No functional change.

Before:
image

After:
image

Also reporting to the Netbeans - https://issues.apache.org/jira/browse/NETBEANS-3782 . Please fix here too for cleaner code and to make autocomplete/refactoring working before it is fixed in Netbeans.

PS: I also checked other files matching `^[^"\n]*("[^"\n]*"[^"\n]*)*"[^"\n]*\$[^"\n]*"` regex but this seems to be the only issue.

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -368,7 +368,7 @@ class App
                 }
                 if ($keys) {
                     $ids = implode(',', $keys);
-                    $remove_function = "$('.ui.dimmer.modals.page').find('${ids}').remove();";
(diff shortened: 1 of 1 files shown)


mvorisek:
@DarkSide666 Thanks, other files seems to be fine, or they are at least parsed to some extend (all PHP files now seems to be fully ok, only some `.jade` and some template files has error flag, but the structure for them is visible (not affected directly by this issue))

Issue #902, opened 2020-02-02
Critical: Checkbox keeps set even if unset/unchecked


mvorisek:
Test code:
```
<?php

require __DIR__ . '/vendor/autoload.php';

$app = new \atk4\ui\App('My App');
$app->initLayout('Admin');

/* @var $form \atk4\ui\Form */
$form = $app->add('Form');
$form->addField('purge_missing', ['CheckBox']);
$form->buttonSave->set('Import');

$form->onSubmit(function(\atk4\ui\Form $form) {
    throw new \Exception($form->model['purge_missing']);
});

```

If the form is just send, the `$form->model['purge_missing']` evaluates to `0`. Fine.

But if the checkbox is checked, form is sent, and then unchecked and submitted again, the value is read as `1`!

Please reproduce and fix asap.

ibelar:
@mvorisek - Sorry could not reproduce the issue with your code above.  Submitting form multiple time always send proper data, whether checkbox input is check or unchecked.

mvorisek:
I have updated the example - the issue occur when an exception is rendered on the server side (and for some reasons then the `purge_missing` value after check/uncheck is sent to the server as `true`).

![image](https://user-images.githubusercontent.com/2228672/73700272-63435180-46e6-11ea-8c5a-4b1f795f7d1b.png)

@ibelar can you please retest the updated example?

ibelar:
@mvorisek - Thanks for the update. 
Fix in  #909

mvorisek:
@ibelar Great job!

PR #909, opened 2020-02-04
Fix/Clear Api Data with all server response

ibelar:
Fix: #902

### DIFF
--- a/js/Release.md
+++ b/js/Release.md
@@ -1,5 +1,12 @@
 ## Release note
 
+### version 1.12.2 (2020-02-04)
+
+- Fix https://github.com/atk4/ui/issues/902 
+   Throwing error on server was not clearing api.data object, thus keeping
+   old value with new server request. This fix make sure api.data gets clear
+   on all server response, successful or not.
+
 ### version 1.12.1 (2020-01-14)
 
 - update api service to generate modals in json request.
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "atkjs-ui",
-  "version": "1.12.1",
+  "version": "1.12.2",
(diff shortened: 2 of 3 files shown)


mvorisek:
The issue will still be presented if 2nd request will be send before the 1st one is returned.

The serialized data should be cleared no later before the request is dispatched (and this is expected to be done in one shot without any context switch).

ibelar:
> The issue will still be presented if 2nd request will be send before the 1st one is returned.

Not really. Fomantic Api is implemented as jQuery plugin which means that each form will have there own data instance object attached to them. successTest will properly clear them for each request. The problem we had before is that when server response was an error,  api.data was not clear for that instance.

mvorisek:
@ibelar And if for some reasons we serialize the same form before the request is returned?

ibelar:
> @ibelar And if for some reasons we serialize the same form before the request is returned?

Can you explain a little more? Not sure what is your use case here.

mvorisek:
@ibelar Currently there is an issue, when the server reponse is not successful, the data are not cleared. Based on the fix I see the issue of bad data will remain if the form is resent before the 1st response has returned.

ibelar:
> @ibelar Currently there is an issue, when the server reponse is not successful, the data are not cleared. Based on the fix I see the issue of bad data will remain if the form is resent before the 1st response has returned.

If the same form is resend prior to receiving its first server response, then fomantic api module will abort the first request. It will also properly clear api.data.

On the other end, if sending multiple forms simultaneously then each form will create its own api request and have their own api.data attach. This fix will properly clear each one of them.

mvorisek:
> If the same form is resend prior to receiving its first server response, then fomantic api module will abort the first request. It will also properly clear api.data.

If so, then this PR is perfect!

PR #913, opened 2020-02-05
Prevent dropdown open on autofocus

### LINKED ISSUE #512: DropDown: dropdown drops down instantly if it's first field in form
If DropDown field is first field in form, then it drops down automatically because it gets focus on form load.
This is annoying.

Possible solution:
- maybe dropdown field should not have `tabindex` ?
- or somehow do `blur` if dropdown field is first field in form.

mvorisek:
Currently, when dropdown is the first control in modal windows, it opens by default and can cover a lot of controls/space.

fixes https://github.com/atk4/ui/issues/512

### DIFF
--- a/src/Modal.php
+++ b/src/Modal.php
@@ -133,7 +133,7 @@ class Modal extends View
             $js_chain->data(['args' => $args]);
         }
 
-        return $js_chain->modal('show');
+        return $js_chain->modal(['autofocus' => false])->blur()->modal('show'); // show modal, but prevent autofocus, as it can open dropdown and cover a lot of controls
     }
 
     /**


mvorisek:
The auto focus of the modal should work by default, only it should be "silent" - i.e. if the first focusable control is for example dropdown, the focus should be placed, but the dropdown should not be opened.

Can someone advise a better solution that the current one?

DarkSide666:
This definitely is a problem, but not sure about best solution.
Simplest solution is to avoid using dropdown field as first field in form.
Another solution is to always add some hidden/useless field as first field in all forms.
Then also you can mess with javascript and do something with it (as you did in pr).

mvorisek:
This is a solution I was using for the past one year.

The best solution should be to update the JS code of each Semantic UI Dropdown/Autocomplete/Lookup component to detect the autofocus (i.e. focus from load / modal show / browser tab change) - so probably supress any autoopen and only detect manual refocus via mouse / keyboard (should be easy) and then trigger the open manually only on these events. I am not an expert on front-end nor JS nor Semantic UI, help wanted.

romaninsh:
Our solution to this so far has been not to put drop-down as a first field. Using auto-focus is huge when you use keyboard and loosing this would suck.

mvorisek:
@DarkSide666 How to solve this? What about 1x1 invisible (without borders and 1% transparent) input field where the focus will be placed and this field will be automatically removed once the focus will be moved.

mvorisek:
Closing, should be solved to keep the focus

PR #916, opened 2020-02-05
Fix enum with custom keys support

mvorisek:


DarkSide666 (changes requested):
I don't think this is correct.

`enum` is for values only format
`values` is for value=>caption format

romaninsh:
Agree with  @DarkSide666 .. Use ->values property.

mvorisek:
> 
> 
> I don't think this is correct.
> 
> `enum` is for values only format
> `values` is for value=>caption format

Makes sense, I was using int keys with enum values as a value, but it makes sense to use enum solely with one of them (key or value) like now. Sorry about this PR.

PR #929, opened 2020-02-07
Use ::addTo() method in View::add()

### LINKED ISSUE #907: Huge improve in autocomplete / refactoring / usability for programmers - return specific type instead of generic one
A lot of code constructs look like:
```
$form = $app->add('Form');
```

in better case there is an extra type hint like:
```
use atk4\ui\Form;

/* @var $form Form */
$form = $app->add('Form');
```

This is a feature request to add `__construct` option `addTo` (and maybe some generic option for things like `addField`) like:
```
use atk4\ui\Form;

$form = new Form(['addTo' => $app]);
```
which should be equivalent of `$form = $app->add('Form');`.

This will hugely improve utocomplete / refactoring / usability for programmers.

All usages and examples should be converted to the new syntax.

`$app->add('Form')` syntax should be dropped in the new 2.x branch and only `$app->add($obj)` should be supported. We should aim step by step for 100% refactorability.

mvorisek:
Fixes #907 , no BC break

Example usage:
```
$crud = \atk4\ui\CRUD::createAndAdd($app); // phpdoc return type = atk4\ui\CRUD
$crud->displayFields = ['name']; // type-hinting work
```

code is equivalent to:
```
$crud = $app->add(\atk4\ui\CRUD::class); // phpdoc return type = atk4\ui\View
$crud->displayFields = ['name']; // type-hinting does not work
```

### DIFF
--- a/demos/accordion-nested.php
+++ b/demos/accordion-nested.php
@@ -5,7 +5,7 @@ require_once __DIR__ . '/init.php';
 /*
 $app->add(['Button', 'View Form input split in Accordion section', 'small right floated basic blue', 'iconRight' => 'right arrow'])
     ->link(['accordion-in-form']);
-$app->add(['ui' => 'clearing divider']);
+$app->add(['View', 'ui' => 'clearing divider']);
 */
 
 $app->add(['Header', 'Nested accordions']);
--- a/demos/accordion.php
+++ b/demos/accordion.php
@@ -4,12 +4,12 @@ require_once __DIR__ . '/init.php';
 
 $app->add(['Button', 'Nested accordions', 'small right floated basic blue', 'iconRight' => 'right arrow'])
     ->link(['accordion-nested']);
-$app->add(['ui' => 'clearing divider']);
+$app->add(['View', 'ui' => 'clearing divider']);
 
 $app->add(['Header', 'Accordion\'s section can be control programmatically.']);
 
 // toggle menu
-$bar = $app->add(['ui' => 'buttons']);
+$bar = $app->add(['View', 'ui' => 'buttons']);
 $b1 = $bar->add(['Button', 'Toggle Section #1']);
 $b2 = $bar->add(['Button', 'Toggle Section #2']);
 $b3 = $bar->add(['Button', 'Toggle Section #3']);
--- a/demos/breadcrumb.php
+++ b/demos/breadcrumb.php
@@ -9,7 +9,7 @@ $crumb = $app->add('BreadCrumb');
 $crumb->addCrumb('UI Demo', ['index']);
 $crumb->addCrumb('BreadCrumb Demo', ['breadcrumb']);
 
-$app->add(['ui'=>'divider']);
+$app->add(['View', 'ui'=>'divider']);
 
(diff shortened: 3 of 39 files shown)


PhilippGrashoff:
Huge thumbs up for changing the code this way. It actually never really bothered me until I started thinking about your proposal. Will do so much good in combination with a powerful IDE like PHPStorm.

romaninsh:
I've created this: https://github.com/atk4/core/pull/123/files

abbadon1334:
@mvorisek 
> ```php
> $crud = $app->add(\atk4\ui\CRUD::class); // phpdoc return type = atk4\ui\View
> $crud->displayFields = ['name']; // type-hinting does not work
> ```
FYI With this syntax the return type can be rewrite in PhpStorm using .phpstorm.meta.php file
https://blog.jetbrains.com/phpstorm/2019/02/new-phpstorm-meta-php-features/
In this case syntax must be always the same, because from what i see array and simple arguments can't be mixed, IMHO the freedom of writing code in php and atk "conflicts" with some strict rules of defining things in editor, which is legit :(

Defining an interface i think solve only partially the issue, IMHO the only way to solve this would be adding a method for every possible sub-component with the corresponding return type, but this means a big addiction of one line methods, :-1:  i follow the quote "More code, more troubles".

At now all major editor work perfectly like this : 
```php
/** @var Crud $crud */
$crud = $app->add(\atk4\ui\CRUD::class);
$crud->displayFields = ['name']; // type-hinting will work
```

I don't know what can be the best solution.

mvorisek:
@abbadon1334 The magic is to HINT type `static` which is already supported by Netbeans and other IDEs. RETURN of type `static` is expected to be added to PHP 8.0.

HINT = phpdoc
RETURN = type checked by PHP

There is no possibility to return particular type without a static method or generics which is not supported by PHP nor phpdoc.

abbadon1334 (changes requested):
need at least a unit test for static method, but i like the static method and codeintel for the return value. GJ @mvorisek

romaninsh:
ATK Core PR has been merged, so I'm closing this. @mvorisek - thanks for this great suggestions. Lets keep this as an alternative code.

https://github.com/atk4/core/pull/123

mvorisek:
@romaninsh I have this in my todo list, but currently I am out of time. The merged code is not fully ok - for example it hints `self` instead of `static` which almost destroy 50% of its magic ;-)

This PR is actually very hand crafted to support type validation - see https://github.com/atk4/ui/pull/929/files#diff-94ba2466eb979e1b8ba97278c547793aR47 - and I think it is more than required as it can catch a lot of errors and I managed to make it fully compatible with existing code (any `->add` method can be rewriten to it). Wdyt about adding the `$seedExclClassName` parameter directly to the new add trait in `atk4/core`?

I can propose the code, but probably no sonner than 2. Mar.

mvorisek:
Work in progress

mvorisek:
Related work:
- https://github.com/atk4/ui/issues/930
- https://github.com/atk4/ui/issues/989

PR #934, opened 2020-02-07
Fix CS to PSR-12

mvorisek:
Add CS config and fix PSR-12 errors.

StyleCI needs to be reconfigured toPSR-12 CS.


romaninsh:
@mvorisek leave this to me,. I'll set this up properly.

mvorisek:
Closing as expected to be fixed by PR #980

Issue #911, opened 2020-02-04
CRUD - delete not working


mvorisek:
Consider this model:
```
class User extends BaseModel
{
    /** @var string */
    public $table = 'user';

    public function init()
    {
        parent::init();

        $this->addField('email', ['required' => true]);
    }
```

Currently, when:
- User model is present/saved and:
- delete button is pressed and delete is confirmed

then:

the CRUD shows edit dialog, instead of sending a request to server to delete the model immediatelly.

romaninsh:
Fixed?

mvorisek:
Fixed, closing.

Issue #922, opened 2020-02-05
CRUD record delete - record is not removed from the table until reload


mvorisek:
If record is deleted by an action, it needs to be removed from CRUD.

The same applies if record is changed and does not longer satisfy the model conditions.

romaninsh:
@mvorisek  was this fixed?

mvorisek:
Fixed, closing.

PR #936, opened 2020-02-07
Fix/Crud delete action and UserAction 

ibelar:
- Fix Crud delete action when editFields are set.
- Fix UserAction when actions->fields is set to true and using other step then Fields.

- Add Crud::addFields property for setting field in add action via Crud seed.


ibelar:
Fixes: #911, #922

src/CRUD.php
```diff
@@ -266,6 +268,17 @@ protected function getExecutor(Generic $action)
             return $this->factory($action->ui['executor']);
         }
 
+        // prioritize CRUD addFields over action->fields for Model add action.
+        if ($action->short_name === 'add' && $this->addFields) {
+            $action->fields = $this->addFields;
+        }
+
+        // prioritize CRUD editFields over action->fields for Model edit action.
+        if ($action->short_name === 'edit' && $this->editFields) {
```
mvorisek: EditFields should be used in case of addField are not defined for add.

Also here and above the condition should check for null as fields may be an empty array.

ibelar: EditFields and addFields are just a quick way to setup add et edit action fields in CRUD seed, without having to modify the Model action itself, so they should be separated.

Having empty array here is fine since it will rely on $action->fields property instead. 
Remember, this is for specific add and edit Model action so passing an empty array would result in UserAction executor "field" step to directly save record without user input.

Issue #932, opened 2020-02-07
Tests are failing on develop


mvorisek:
see https://travis-ci.org/atk4/ui/builds/647286179

help wanted

mvorisek:
Issue introduced in PR #900

DarkSide666:
should be fixed by #942

mvorisek:
#942 fixed the original issue, but there is still one more test to fix: https://travis-ci.org/atk4/ui/jobs/649098810#L919

And do not merge any failing PR then :)

DarkSide666:
Travis tests are now fixed.
That atk4/ui-demo is unfinished integration with codefresh.io.

mvorisek:
> Travis tests are now fixed.
Perfect!

Closing this issue and opened an separate one for the `atk4/ui-demo`.

DarkSide666:
Closing in favor of #946

PR #942, opened 2020-02-09
Fix/Filter Popup test

ibelar:
- update page url for Table Filter popup test file.


ibelar:
Fix #932 - Thanks @mvorisek for catching this.

mvorisek:
@ibelar Is the removal of the column filters intended? Is seems wrong to me as the demo for it is named `tablefilter.php` and now there are no filters, please see https://github.com/atk4/ui/pull/900#issuecomment-583786623

PR #945, opened 2020-02-11
fix/Duplicate Url encoding

### LINKED ISSUE #924: Critical: atkReloadView does encode URL values twice
Example:
```
        $button = $this->app->add('Button')->set('click me');
        $button->on('click', $button->js()->atkReloadView(
            ['uri' => $button->url(), 'uri_options' => ['move' => 'next']]
        ));
```

How to reproduce?
- navigate to URL: `https://xxx/yyy.php?crud_model=AppAtk4%5CModel`
- click the button and notice, that the new URL of that request is URL encoded twice

Currently this URL is requested:
```
https://xxx/yyy.php?crud_model=AppAtk4%255CModel
```
vs. (expected)
```
https://xxx/yyy.php?crud_model=AppAtk4%5CModel
```

Btw: where is the `atkReloadView` JS method defined? It seems that there is only `atkjs-ui.min.js`.

ibelar:
Add fix in reload-view and ajaxec plugin.

fix #924

### DIFF
--- a/js/src/plugins/ajaxec.plugin.js
+++ b/js/src/plugins/ajaxec.plugin.js
@@ -32,7 +32,7 @@ export default class ajaxec extends atkPlugin {
         const data = this.settings.uri_options ? this.settings.uri_options : {};
 
         // retrieve param from url.
-        let urlParam = $.atkGetQueryParam(this.settings.uri);
+        let urlParam = $.atkGetQueryParam(decodeURIComponent(this.settings.uri));
 
         // get store object.
         let store = atk.dataService.getStoreData(this.settings.storeName);
--- a/js/src/plugins/ajaxec.plugin.js
+++ b/js/src/plugins/ajaxec.plugin.js
@@ -32,7 +32,7 @@ export default class ajaxec extends atkPlugin {
         const data = this.settings.uri_options ? this.settings.uri_options : {};
 
         // retrieve param from url.
-        let urlParam = $.atkGetQueryParam(decodeURIComponent(this.settings.uri));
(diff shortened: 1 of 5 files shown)


js/src/plugins/ajaxec.plugin.js
```diff
@@ -32,7 +32,7 @@ export default class ajaxec extends atkPlugin {
         const data = this.settings.uri_options ? this.settings.uri_options : {};
 
         // retrieve param from url.
-        let urlParam = $.atkGetQueryParam(this.settings.uri);
+        let urlParam = $.atkGetQueryParam(decodeURIComponent(this.settings.uri));
```
mvorisek: Will this work if `this.settings.uri` = `http://x/y%3Fq`?

ibelar: Why do you think it will not work?

mvorisek: `this.settings.uri` seems to be that it can contain URL in standard format like `http://x/y?q` or `http://x/y%3Fq` (URL with path part = `/y%3Fq`, i.e. totaly different URL).

If so, then decoding `http://x/y%3Fq` to `http://x/y?q` is wrong.

ibelar: I see. This will cover it. https://github.com/atk4/ui/pull/945/commits/4e4eaca3e4475d6389840659fde75cbae1d246f8

mvorisek: @ibelar Tested, very good, thanks!

PR #950, opened 2020-02-13
REFACTOR: use array casting and arguments spread

georgehristov:


src/Form.php
```diff
@@ -363,11 +363,7 @@ public function addField($name, $decorator = null, $field = null)
     public function addFields($fields)
     {
         foreach ($fields as $field) {
-            if (is_array($field)) {
-                $this->addField(...$field);
-            } else {
-                $this->addField($field);
-            }
+            $this->addField(...(array) $field);
```
DarkSide666: what if `$field===null`? But ok, in this case it would be very uncommon to use `null value :)

PR #951, opened 2020-02-13
[FEATURE] introduce autocomplete dependency

georgehristov:
- introduce multiple selection
- introduce dependency handling
- introduce `renderRowFunction`
- cleanup code
- add demos
- update docs

### DIFF
--- a/demos/autocomplete.php
+++ b/demos/autocomplete.php
@@ -98,3 +98,79 @@ $lookup = $form->addField('country_b', [
 $lookup->addFilter('letter1');
 
 $form->buttonSave->set('Add Countries');
+
+$app->add(['Header', 'Auto-complete dependency']);
+
+$form = $app->add(new \atk4\ui\Form(['segment']));
+$form->add(['Label', 'Input information here', 'top attached'], 'AboveFields');
+
+$form->addField('starts_with', [
+    'DropDown',
+    'values'       => [
+        'a' => 'Letter A',
+        'b' => 'Letter B',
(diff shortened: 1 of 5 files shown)


romaninsh:
Regarding [feature] in the title there is better way:

Modifying this would allow to create universal label throughout all ATK repositories:


https://github.com/atk4/infrastructure/blob/master/projects/agiletoolkit.org/github/agiletoolkit.org/addon/addon-repo.tf#L53

see also https://www.terraform.io/docs/providers/github/r/issue_label.html

And then labels can be used here:
https://github.com/atk4/ui/blob/develop/.github/release-drafter.yml#L3

and they would be properly grouped here: https://github.com/atk4/ui/releases

romaninsh (approved):

DarkSide666 (approved):
Looks good to me.
One thing I don't understand is how this goes together with Lookup and DropDown classes. I mean - AutoComplete was meant to be kind of deprecated class in favor of Lookup I think.
@ibelar what you think about this?

mvorisek (approved):
@georgehristov, perfect work!

ibelar:
Excellent job! I believe that dependency implement here is easier to use than actual Lookup input.

mvorisek:
> I believe that dependency implement here is easier to use than actual Lookup input.

What do you mean exactly?

mvorisek:
> I believe that dependency implement here is easier to use than actual Lookup input.

@ibelar What do you mean exactly?

Issue #903, opened 2020-02-03
Fix outdated doc for CRUD


mvorisek:
Fields like `fieldsRead` are no longer preseted in the latest release - https://agile-ui.readthedocs.io/en/latest/crud.html#atk4\ui\CRUD::$fieldsRead

ibelar:
Fix in #953

PR #953, opened 2020-02-15
Feature/Add properties to force js execution for actions

ibelar:
Add new properties to force Table reloading or Table row remove using certain action. 

```
/** @var array Action name container that will reload Table after executing */
public $reloadTableActions = [];
```
```
/** @var array Action name container that will remove the corresponding table row after executing */
public $removeRowActions = [];
```

For example, if the model action returns an unload model but you would still like to see reload the table, just add it to the $reloadTableActions array property to force Crud table to reload.

## Update Crud docs for v2.0

### DIFF
--- a/docs/crud.rst
+++ b/docs/crud.rst
@@ -46,109 +46,61 @@ parameters to CRUD.
 Disabling Actions
 =================
 
-.. php:attr:: canCreate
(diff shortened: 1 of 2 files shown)


ibelar:
Fix #903

PR #954, opened 2020-02-16
[CHORE] use shorthand syntax

georgehristov:


src/Table.php
```diff
@@ -273,7 +270,7 @@ public function setFilterColumn($cols = null)
         // set filter to all column when null.
         if (!$cols) {
             foreach ($this->model->getFields() as $key => $field) {
-                if (isset($this->columns[$key]) && $this->columns[$key]) {
+                if (! empty($this->columns[$key])) {
```
PhilippGrashoff: AFAIK this will create a PHP notice if `$this->columns[$key]` does not exist? 
How about
`if($this->columns[$key] ?? null) {`

DarkSide666: No this is correct.
_A variable is considered empty if it does not exist or if its value equals FALSE. empty() does not generate a warning if the variable does not exist. That means empty() is essentially the concise equivalent to !isset($var) || $var == false._
So empty() will not throw warning (and probably also no notice) if array element doesn't exist.

PR #959, opened 2020-02-17
[FEATURE] enable use of existing menu for Grid

georgehristov:
Grid will use the Menu passed to `$menu` property directly if it is already an object

As example is the use of common Menu for the page. Then it can be assigned to accept items from the Grid instead of Grid creating its own menu.

Without feature:
before

With feature:
after


DarkSide666 (approved):
LGTM

PR #960, opened 2020-02-18
[FIX] table column actions icon

### LINKED ISSUE #554: Column filter going on second row

georgehristov:
fixes #554

a4d8720 is the actual fix
5b958f7 refactoring of variable names and using shorthand syntax for clarity

Print screen of the solution in action:
solution

### DIFF
--- a/src/TableColumn/Generic.php
+++ b/src/TableColumn/Generic.php
@@ -79,7 +79,7 @@ class Generic
      *
      * @return mixed
      */
-    public function addPopup($popup = null, $icon = 'table-filter-off icon')
+    public function addPopup($popup = null, $icon = 'table-filter-off')
     {
         $id = $this->name.'_ac';
 
@@ -114,7 +114,7 @@ class Generic
 
         $this->headerActionTag = ['div',  ['class'=>'atk-table-dropdown'],
             [
-                ['i', ['id' => $id, 'class' => $class.' icon']],
+                ['i', ['id' => $id, 'class' => $class.' icon'], ''],
             ],
         ];
     }
@@ -128,7 +128,7 @@ class Generic
     {
         $this->headerActionTag = ['div',  ['class'=>'atk-table-dropdown'],
             [
-                ['i', ['id' => $this->name.'_ac', 'class' => $icon.' icon']],
+                ['i', ['id' => $this->name.'_ac', 'class' => $icon.' icon']. ''],
             ],
         ];
     }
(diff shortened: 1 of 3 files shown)


ibelar:
@georgehristov - Thanks for taking care of this. However, this ends up with the same problem as before, i.e. icon moving under the label name when there is no enough space. 

Screen Shot 2020-02-18 at 8 22 27 AM

It would be great if you could find a way to take care of this. For the rest, it looks very good!

georgehristov:
@ibelar Thanks. The problem was that even if there is space the icon ended up below the caption.
Will look into it to keep them together always...

src/TableColumn/Generic.php
```diff
@@ -294,74 +286,69 @@ public function getTag($position, $value, $attr = [])
...
-        } else {
-            $tag = $this->getTag(
-                'head',
-                [$captionHtmlTag],
-                $attr
-            );
-        }
-
-        return $tag;
+        return $this->getTag('head', [['div', compact('class'), $caption]], $attr);
```
georgehristov: I figured out how to do that but I get a problem with generating tags.

When generating the tag for the header all works well on atk4 side on the line linked to this comment.

The generated tag (var_dump) is:
```
<th data-column="accessed_at" id="atk_layoutview_fileaccesshistory_crud_view_table_generic_th" data-sort="accessed_at">
	<div class="atk-table-column-header">
		<div class="atk-table-dropdown">
			<i id="atk_layoutview_fileaccesshistory_crud_view_table_generic_ac" class="table-filter-off icon">
		</div>Accessed At</div>
</th>
```

Yet in the browser the column title is wrapped with a copy of the <i> tag?

```
<th data-column="accessed_at" id="atk_layoutview_fileaccesshistory_crud_view_table_generic_th" data-sort="-accessed_at">
	<div class="atk-table-column-header sorted ascending">
		<div class="atk-table-dropdown">
			<i id="atk_layoutview_fileaccesshistory_crud_view_table_generic_ac" class="table-filter-off icon"/>
		</div>
		<i id="atk_layoutview_fileaccesshistory_crud_view_table_generic_ac" class="table-filter-off icon">Accessed At</i>
	</div>
</th>
```

All works fine is text is "declared" first in the atk tag array

Any idea?

georgehristov: Found the issue - it is related to the fact that `<i>` tag is not closed when generated by atk

georgehristov:
Now it should work properly

georgehristov:
It was a typo causing both, please recheck

ibelar:
Thanks, @georgehristov - Looks very good now!

georgehristov:
@DarkSide666 thanks for noticing this. I introduced a fix

PR #961, opened 2020-02-18
move typecasting to Field and allow DropDown towork with model field …

DarkSide666:
* Moves typecasting to Field class. Form should not care about that.
* Now allow FormField\DropDown to also work nicely with model field with type=array not just string (CSV)

For example, model field like this:
```
        $this->addField('expression_list', [
            'type'      => 'array',
            'required'  => true,
            'serialize' => 'json',
            'ui' => [
                'form' => [
                    \atk4\ui\FormField\DropDown::class,
                    'isMultiple'=>true,
                    'model'=>$this->ref('service_id')->ref('Expressions'),
                ],
                'table' => ['Labels', 'values'=>$v],
            ],
        ]);
```

### DIFF
--- a/docs/field.rst
+++ b/docs/field.rst
@@ -317,15 +317,18 @@ DropDown uses Fomantic UI Dropdown (https://fomantic-ui.com/modules/dropdown.htm
 
 Usage with a Model
 ------------------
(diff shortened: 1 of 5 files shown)


romaninsh:
no docs? :D bad coverage?

DarkSide666:
docs added, merging in

Issue #966, opened 2020-02-22
[IMPROVEMENT] renaming hook addHook method


georgehristov:
I find the hook methods (`hook` and `addHook`) a bit confusing for newcomers.

How about renaming the `addHook` to `onHook` (keeping BC) to make it more intuitive?
IMHO this would improve the DX.

If agreed with the change you can assign me to that issue.

georgehristov:
Maybe I am a bit obsessed with semantics but starting to think that `onHook` should even be simply `on`
WDYT?

PhilippGrashoff:
Dont like it, as "on" does not give any hint what it does. "onHook" states clearly that it has something to do with a hook.
There also is "on" in ATK UI, definitely that could confuse new ATK users.

georgehristov:
Good, then we can close the issue

PR #968, opened 2020-02-24
[REFACTOR] rename addHook to onHook

georgehristov:
related to https://github.com/atk4/ui/issues/966


romaninsh:
@georgehristov - could you please re-trigger the CI/CD? I tried rerunning but something is wrong. The rename was applied everywhere, so we should make it consistent here (in UI) too.

georgehristov:
@romaninsh I have triggered the CI but was there a change in StyleCI policy?

romaninsh:
Thanks for this!

PR #980, opened 2020-03-12
Bump PSR and reactivate styleci

romaninsh:
just get styleci going again.


mvorisek:
Add CS fixer dev-dependency + rule file like in the core repo - https://github.com/atk4/core/blob/develop/.php_cs.dist

Preferably for all atk4 repos and with one/same rule file.

romaninsh:
I'll close this PR and will open new one due to conflicts. Easier this way. Just need to sort out #991.

mvorisek:
@romaninsh Why not to force-push this PR branch?

PR #984, opened 2020-03-13
Add return type hint

mvorisek:


PhilippGrashoff:
Hi, if it always returns a FormField\Generic, why not use type hinting like
`public function getField($name): Generic` ? I think thats way better than only a comment.

mvorisek:
@PhilippGrashoff  I agree, I will amend this PR, I was not sure if this is ok as atk4 has very loose type checking.

PhilippGrashoff:
As far as I understood thats one thing thats constantly improved :)

DarkSide666:
And no more `@return FormField\Generic` typehint? :(

mvorisek:
> And no more `@return FormField\Generic` typehint? :(

Yes, this is new industry standard practice to not write phpdoc without any extra meaningfull info.

The same for typed parameters without description. (same again, no extra info, the phpdoc only occupies space)

This is also default behaviour of CS fixer with default rules and probably also advised by PSR-12.

PR #985, opened 2020-03-14
Fix missing EOL in text files

mvorisek:
End EOL is industry standard and in some enviroments like our it causes warnings.

Better fix of https://github.com/atk4/ui/commit/5adec249dafe70198f83220e2835491e4d172cc1


mvorisek:
Help wanted. Why are the TravisCI tests - https://travis-ci.org/github/atk4/ui/jobs/662349532#L573 - failing and how to fix them?

DarkSide666:
This is one template file which should not have EOL because it will change HTML markup and as result tests will fail.
I guess final EOL line is standard in php and other source code files, but it's not mandatory in html templates.

mvorisek:
I understand the the end EOL will be added to render, but the industry standard practice is to remove one end EOL of the data if presented and this should be the solution for this.

But before this general fix the issue should be identified as multiple whitespaces should not break HTML processing - all other already normalized files are fine https://github.com/atk4/ui/commit/143e953eb5d14e943aa71c9f85e8dc88d8399438

mvorisek:
The issue is that:
```
<div id="atk_admin_tabs_tabssubview_form_generic_button" class="ui primary labeled icon button" style="" tabindex="0"><i id="atk_admin_tabs_tabssubview_form_generic_button_icon" class="  mail icon" style="" ></i>Subscribe</div>
```

is with fixed HTML file rendered as:
```
<div id="atk_admin_tabs_tabssubview_form_generic_button" class="ui primary labeled icon button" style="" tabindex="0"><i id="atk_admin_tabs_tabssubview_form_generic_button_icon" class="  mail icon" style="" ></i>
Subscribe</div>
```

It there a simple option to fix this search expression:
https://github.com/atk4/ui/blob/592503b9f2a420e1921323295c3d234a670b2996/features/form.feature#L9
to match ` ?Subscribe`?

mvorisek:
Template loader fixed. End NL is trimmed if presented. If end NL is desired, simply add 2 - the last one will be always trimmed.

DarkSide666:
1. using preg_replace will probably slightly slow down template parsing
2. I guess now this will generate all HTML in one long, unreadable line? If so, then you should add these 2 newlines at the end of most of template files too.

mvorisek:
> 1. using preg_replace will probably slightly slow down template parsing

`preg_*` functions are amazingly fast, as the compiled regexes are cached - I was evaluating them few months ago - and to my suprise one `preg_replace` is usually faster than two string manipulations with very basic functions like `trim`, `substring`, ...

> 2. I guess now this will generate all HTML in one long, unreadable line? If so, then you should add these 2 newlines at the end of most of template files too.

Yes and no :) - End NLs were added on Jan 22 - https://github.com/atk4/ui/commit/143e953eb5d14e943aa71c9f85e8dc88d8399438 - so this fix basically restored the old render results prior this date. But feel free to add an extra NL if needed.

DarkSide666:
ok, nevermind then :)

PR #997, opened 2020-03-17
fix #995, simplify session object in demo

### LINKED ISSUE #995: Call of add() on Session demo class
What should this add() do? The add method is not defined..
https://github.com/atk4/ui/blob/2c51c5fe81fc109efb0b1815ba0e9a93482bb59c/demos/Session.php#L121

DarkSide666:
fix #995


DarkSide666:
Yeah, this was some old demo code I guess. At that time we probably didn't have this as separate trait yet.

PR #1012, opened 2020-03-25
[Feature] Dropdown Cascade Form Field

ibelar:
Implementation of DropDownCascade Form field:

Will load it's list value according to the value set in another Dropdown. 

For example, showing a list of products based on selected Category and subcategory.

``` php
$f = \atk4\ui\Form::addTo($app);

$f->addField('category_id', [\atk4\ui\FormField\DropDown::class, 'model' => new Category($db)]);
$f->addField('sub_category_id', [\atk4\ui\FormField\DropDownCascade::class, 'cascadeFrom' => 'category_id', 'reference' => 'SubCategories']);
$f->addField('product_id', [\atk4\ui\FormField\DropDownCascade::class, 'cascadeFrom' => 'sub_category_id', 'reference' => 'Products']);

```

Screen Shot 2020-03-25 at 2 18 21 PM

### DIFF
--- a/demos/cascade-dropdown.php
+++ b/demos/cascade-dropdown.php
@@ -29,7 +29,6 @@ class SubCategory extends \atk4\data\Model
 
         $this->hasOne('category_id', new Category());
         $this->hasMany('Products', new Product());
-
     }
 }
 
--- a/demos/cascade-dropdown.php
+++ b/demos/cascade-dropdown.php
@@ -1,0 +1,58 @@
+<?php
+require_once __DIR__ . '/init.php';
+require_once __DIR__ . '/database.php';
+
+/*********** MODEL ***************/
+
+class Category extends \atk4\data\Model
+{
+    public $table = 'category';
+
+    public function init()
+    {
+        parent::init();
+        $this->addField('name');
+
+        $this->hasMany('SubCategories', new SubCategory());
+        $this->hasMany('Products', new Product());
+    }
+}
+
+class SubCategory extends \atk4\data\Model
+{
+    public $table = 'sub_category';
+
+    public function init()
+    {
+        parent::init();
+        $this->addField('name');
+
+        $this->hasOne('category_id', new Category());
+        $this->hasMany('Products', new Product());
+
+    }
+}
+
+class Product extends \atk4\data\Model
+{
+    public $table = 'product';
+
+    public function init()
+    {
+        parent::init();
+        $this->addField('name');
+        $this->addField('brand');
+        $this->hasOne('category_id', [new Category()])->addTitle();
+        $this->hasOne('sub_category_id', [new SubCategory()])->addTitle();
+    }
+}
+
+$f = \atk4\ui\Form::addTo($app);
+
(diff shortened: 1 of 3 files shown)


georgehristov:
@ibelar is this not a duplicate of dependency functionality?

ibelar:
@georgehristov - Not exactly.

This one is applicable to DropDown input and create a cascading dropdown list, or  a series of dependent DropDownList controls .  The values(list) of DropdownCascade input is based on the selected value of it dependant cascadeFrom input. Changing cascadeFrom input value will reload DropDownCascade list, based on this new value.

georgehristov:
@ibelar this was exactly what I was aiming at when implementing the dependency functionality.
You have the possibility to apply dependency (cascading) on multiple inputs

ibelar:
@georgehristov - Understand that and your implementation offer even more possibilities. But here, even if the end result can be the same, the implementation and user experience are different. 

This one is base on DropDown control, not AutoComplete, and each DropDown is tightly coupled with each other.  It also relies on hasMany references between models. You select value from one, then the next cascading one is automatically filled with related values and so on.

For example, when selecting a Category, it will automatically fill the Sub Category dropdown with the corresponding value, then selecting a sub-category will automatically fill the Product dropdown list.
When changing a Category, then cascading dropdowns are clear automatically i.e. sub-category and product are reset and fill with new corresponding value.

georgehristov:
@ibelar thanks for clarification. Of course expanding available features is great! Just wanted to point this out to avoid double work but if experience and use scenario is different then this is not the case.

mvorisek:
Wdyt about implementing this in `Dropdown` directly?

ibelar:
> Wdyt about implementing this in `Dropdown` directly?

@mvorisek - I have thought about it also but then I remember AutoComplete and Lookup. Things would have been simpler if we had tried implementing filtering by extending class, I think.  

So if someone has a different, or better, way of implementing Cascading between dropdown, then it would be easier to switch. Thus, DropDownCascade become a specific specialization of DropDown.

mvorisek:
@ibelar I have created an issue #1020.

I think the proper solution is:
1. Implement everything related to dropdown in `Dropdown` /wo dynamic data, i.e. without API.
2. Wrap `Dropdown` in `Lookup` and implement here dynamic data, i.e. data from API.

romaninsh:
LGTM. Although I do think that the functionality should have probably been in DropDown.

romaninsh (approved):

mkrecek234:
I am using the DropdownCascade implementation which works when I select (in my case a 3-staged) dropdown set of fields cascaded. The chosen IDs are properly stored in the DB. However when re-loading the form, the stored values of each dropdown are not properly shown as pre-selected values (but the DropdownCascade fields are empty).

`	    $gr->addField('make_id', [DropDown::class, 'model' => new CarMake($this->app->db)])->set(3);
	    $gr->addField('model_id', [DropDownCascade::class, 'cascadeFrom' => 'make_id', 'reference' => 'CarModels']);
	    $gr->addField('motor_id', [DropDownCascade::class, 'cascadeFrom' => 'model_id', 'reference' => 'Motors']);`

PR #1016, opened 2020-03-27
fix adding of sub-menus

### LINKED ISSUE #1015: Using static::addTo results in unexpected behavior when called from child classes
As example in `Menu` methods `addMenu`, `addGroup`, `addMenuRight` use `static::addTo`.
I had defined a child class `NavMenu` extending `Menu`. 
When calling the `addMenu` method within the new class the expected behavior is that a `Menu` is added (as original behavior) and not `NavMenu` (as happens now).

Using `self` instead of `static` in `Menu` methods does not solve the issue. Only using `Menu::addTo` results in expected behavior. That is because `static` is used in `addTo` method as well.

All classes using `addTo` must be checked for this.

georgehristov:
fixes #1015

### DIFF
--- a/src/Menu.php
+++ b/src/Menu.php
@@ -91,7 +91,7 @@ class Menu extends View
      */
     public function addMenu($name)
     {
-        $subMenu = static::addTo($this, ['defaultTemplate' => 'submenu.html', 'ui' => 'dropdown', 'in_dropdown' => true]);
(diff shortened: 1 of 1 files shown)


mvorisek:
This PR restore the original behaviour changed in https://github.com/atk4/ui/commit/1744324984af52caacf980c384e031bf46f5a194#diff-d1b2b70dc450f42ab1aed7b47ed90cfeR94

Thanks @georgehristov

Issue #1010, opened 2020-03-23
CRUD - PHP notice / bad index when submitted


mvorisek:
When CRUD is submitted, the following PHP notice is shown:
```
<b>Notice</b>:  Undefined offset: 0 in <b>.../vendor/atk4/ui/src/jsCallback.php</b> on line <b>144</b><br />
```

It seems that:
a) the `$response` is an array with one value but instead of with index `0`, the index is `1`
b) adding extra test if the index `0` exists seems to fix this issue

Questions:
a) What does the `0` index mean?
b) what is the purpose of the extra code when index `0` is presented?

How to fix this?

ibelar:
This problem is associated with form submit having more than one hook to deal with and come from Hook refactoring.

https://github.com/atk4/core/blob/188895e63e2a30458a142e2b5d5e3f4751f69ffb/src/HookTrait.php#L175

For example:
```php
$form = \atk4\ui\Form::addTo($tab);
$form->addField('email');
$form->addField('checkbox', ['CheckBox']);
$form->addField('checkbox_', ['CheckBox']);

$form->onSubmit(function ($form) {
    // implement subscribe here
    $m = new \atk4\ui\Message('Error executing ', 'red');
    $m->init();
    $m->text->content = ('test');
    return $m;
});
````
When form is submitted, each CheckBox has added a 'loadPost' hook. Thus, when hook are executing, the response from `$response = $this->hook('submit');` is not correctly indexed. Because submit hook is the third one to execute, it will have an index of 2.

On the other side, jsCallback return is expected to receive a response with a 0 base index in order to check if a View object is return by form::onSubmit and act accordingly.

There are different way to tackle this problem but my question to @mvorisek is: 
Do we need to return an index response array in HookTrait::hook method?
````
 $return[$index] = call_user_func_array(
                            $data[0],
                            array_merge(
                                [$this],
                                $args,
                                $data[1]
                            )
                        );
````

if so, then we should check the response in Form::ajaxSubmit() method and re-index response properly.
something like:
````
                $response = $this->hook('submit');
                if (is_array($response)) {
                    $response = array_values($response);
                }
````

mvorisek:
I have already analysed the impact of the hook refactoring already, but thanks, need help with fixing it - see https://github.com/atk4/ui/issues/986 . It makes sense to return the hook values indexes by the hook index that produced it.

[...]

PR #1017, opened 2020-03-27
[Fix] PHP Notice Bad Index on jsCallback

ibelar:
Refactor jsCallback::getAjaxec method.
- Returning a View is not dependant of index[0] anymore when in array.

Fix: #1010


src/jsCallback.php
```diff
@@ -122,11 +122,13 @@ public function set($callback, $args = [])
...
      *
-     * @param array|jsExpressionable $ajaxec  Array of jsExpressionable
-     * @param string                 $msg     General message, typically won't be displayed
-     * @param bool                   $success Was request successful or not
+     * @param array|jsExpressionable $ajaxec Array of jsExpressionable
+     * @param string $msg General message, typically won't be displayed
+     * @param bool $success Was request successful or not
      *
-     * @return [type] [description]
+     * @return void [type] [description]
```
mvorisek: `[type] [description]` is probably unintended.

If nothing is expected to be returned, do not add phpdoc at all and add return type `: void` to the method header directly.

src/jsCallback.php
```diff
@@ -188,4 +175,46 @@ public function getURL($mode = 'callback')
...
+
+        return $action;
+    }
+
+    /**
+     * Render View into modal.
+     *
+     * @param View $response
+     *
+     * @return jsExpression
```
mvorisek: Remove phpdoc and force type in the code directly.

src/jsCallback.php
```diff
@@ -188,4 +175,46 @@ public function getURL($mode = 'callback')
...
+     * @throws \atk4\core\Exception
+     */
+    private function _jsRenderIntoModal($response)
+    {
+        $html = '<div class="ui modal"> <i class="close icon"></i>  <div class="content atk-content"> ';
```
mvorisek: Quite hardcoded, is possible to render this using standard `View` components?

src/jsCallback.php
ibelar: yes, will remove it.

src/jsCallback.php
ibelar: in fact, why not passing Modal directly. Then developper would have total control of Modal to pass in. I will make some change here.

mvorisek:
@ibelar Please rebase on the latest `develop`.

src/jsCallback.php
mvorisek: Very good!

src/jsCallback.php
mvorisek: I fixed the merge conflicts and fixed this as well.

mvorisek:
@ibelar Behat tests / TravisCI needs to be updated for commit https://github.com/atk4/ui/pull/1017/commits/ae8c38c4c1d99413fed44cf17ab9c5df7a0f878d

ibelar:
@mvorisek - Selenium/Behat test are fixed.

PR #1018, opened 2020-03-27
[refactor] AutoComplete into Lookup

### LINKED ISSUE #983: Lookup vs Autocomplete
What is the extra functionality of `Autocomplete` form field vs. `Lookup`?

A lot of the code is copy pasted between these two classes - is it ok to deprecate `Autocomplete` or share the code by implementing `Lookup` by extending `Autocomplete`?

Compare these two files:
- https://github.com/atk4/ui/blob/develop/src/FormField/Lookup.php
- https://github.com/atk4/ui/blob/develop/src/FormField/AutoComplete.php

ibelar:
I think it is time to stop maintaining two versions of the same input type field and keep what @georgehristov has build for dependency setup in AutoComplete. 

This Pr keep the Lookup class name only, which was prefered in the past and AutoComplete class was to be removed eventually. 
BUT keep latest AutoComplete implementation with dependency use.

- Remove old Lookup class and template;
- Rename AutoComplete class into Lookup;
- Rename proper template file;
- Replace usage of AutoComplete for Lookup in demo files;
- Keep AutoComplete for backward compatibility by extending Lookup only.

Fixes #983


src/FormField/AutoComplete.php
```diff
@@ -1,457 +1,11 @@
 <?php
+/**
+ * @obsolete Backward compatibility.
```
mvorisek: You should probably use `deprecated` instead of `obsolete`, `obsolete` is not officially supported in phpdoc - see https://docs.phpdoc.org/latest/references/phpdoc/index.html

mvorisek:
@ibelar Please rebase on the latest `develop`.

src/FormField/Lookup.php
```diff
@@ -478,57 +412,46 @@ public function renderView()
...
     }
+
+    /**
+     * Convert value to expected comma separated list before setting it.
+     *
+     * {@inheritdoc}
+     *
+     * @see \atk4\ui\FormField\Generic::set()
+     */
+    public function set($value = null, $junk = null)
```
mvorisek: What is the use-case of this method? If we have `$this->values`, why do we need this?

src/FormField/Lookup.php
```diff
@@ -99,69 +117,32 @@ class Lookup extends Input
...
      */
-    public $filterEmpty = 'All';
-
-    public $filterChain = null;
+    public $renderRowFunction;
 
     /**
-     * Array containing filters.
+     * Whether or not to accept multiple value.
+     *   Multiple values are sent using a string with comma as value delimiter.
```
mvorisek: Is it possible to support string here? - if string is set, then use this as a delimiter.

PR #1041, opened 2020-03-31
Dump Behat page data on failed step

### LINKED ISSUE #1022: Behat - dump page source if test has failed
Important to quickly identify failed tests like this one: https://travis-ci.org/github/atk4/ui/jobs/668304912#L1017

mvorisek:
Fixes #1022

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -1,6 +1,8 @@
 <?php
 
 use Behat\Behat\Context\Context;
+use Behat\Behat\Hook\Scope\AfterStepScope;
+use Behat\Testwork\Tester\Result\TestResult;
 use Behat\MinkExtension\Context\RawMinkContext;
 
 /**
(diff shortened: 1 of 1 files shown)


mvorisek:
PR complete, more than needed as the current `develop` tests are broken (issue #1043).

Is it ok to add `j0k3r/php-imgur-api-client` to `dev` composer dependencies only for this PR? Or do you have a better idea where to upload a creen of a failed test?

Screenshots are important for quick debugging and even more important for AJAX tests, where the main page source may be not enought.

DarkSide666:
I think it's OK to use `j0k3r/php-imgur-api-client` in `require-dev` section if that helps us to identify issues.

romaninsh:
I wonder if there is a way to attach screenshot to the issue through a comment.

romaninsh:
On the PR - it only adds the method. At least some documentation is necessary on how others can use it.

features/bootstrap/FeatureContext.php
```diff
@@ -26,6 +28,30 @@ public function getSession($name = null)
         return $this->getMink()->getSession($name);
     }
 
+    /**
+     * Dump current page data when step failed to allow easy debug on TravisCI.
+     *
+     * @AfterStep
+     */
+    public function dumpPageAfterFailedStep(AfterStepScope $event)
```
romaninsh, line 37: do we need a place to store screenshot? I can provision a public S3 bucket for that.

romaninsh (changes requested):

romaninsh:
I can provide an S3 bucket where image(s) could be uploaded.

features/bootstrap/FeatureContext.php
mvorisek: > 
> 
> On the PR - it only adds the method. At least some documentation is necessary on how others can use it.

It has on step hook, so it is added automatically activated for every failed step.

mvorisek: > 
> 
> do we need a place to store screenshot? I can provision a public S3 bucket for that.

I was planning to use imgur.com, but S3 is good - can you please add it directly to this PR? And please - do not forget to handle automatic expires after some time + always put there a link to delete - important, when some sensitive data will be dumpled/saved by a mistake.

S3 can be also used for dumping the page source (currently printed to stdout in this PR).

mvorisek:
replaced with https://github.com/atk4/ui/pull/1281

Issue #989, opened 2020-03-16
All UI elements should extend View


mvorisek:
Currently some classes like `Callback` are not extended from `View`.

This break things a litte bit as strict type checking is not possible. Is there any significant reason or can all UI elements than can be added to the render tree be based on the `View` class?

Afftected (top-most only) classes:
- `atk4/ui/Callback`
- `atk4/ui/Template`
- `atk4\ui\TableColumn\Generic`

Once fixed, change no class (unsafe) checking to check if the new objects extends at least `self`/`View` - see https://github.com/atk4/ui/pull/929/files#diff-248b7e363055cfdcca5a52260252ef41R425

romaninsh:
I think this is non issue. All listed class have a good reason not to extend View. 

Callback is non visual and does not facilitate itself as an entity in html tree. 

Template is too abstract and is an utility class

TableColumn is also an utility for rendering cells and is part of Table therefore cannot exists without it. Field is also part do Form but they can be individually used and therefore extend View

mvorisek:
> I think this is non issue. All listed class have a good reason not to extend View.

For me, it is, as long we want a consistent code - see for ex. this PR: https://github.com/atk4/ui/pull/1047/files

> `Callback` is non visual and does not facilitate itself as an entity in html tree.
> Template is too abstract and is an utility class

Yes, it seems `Template` should not/can not extend `View`.

This issue needs to be further analysed. We need at least these interfaces:
- `IRenderableItem` (for `App::terminateHTML()`, `render()`, `getHTML()`)
- `ITreeItem` (for `App::add()`)

Please provide feedback for this. Basically every common method should be called on checked type (class or interface).

romaninsh:
`instanceof \atk4\ui\View` is a good ol' way to check.

I do not see how ATK would benefit from this suggestion, but I can see lot of issues with this change.

mvorisek:
But you can not then enforce this type as request type for any render tree item. Makes sense so?

PR #1047, opened 2020-03-31
Fix refactorability for "owner"

mvorisek:


mvorisek:
I can not be currently fixed at one place - see #989

src/TableColumn/Generic.php
```diff
@@ -10,6 +10,8 @@
 
 /**
  * Implements Column helper for table.
+ *
+ * @property \atk4\ui\Table $owner
```
mvorisek, line 14: Only a typehint, is it right that this base class will be the owner in 95% of the use cases?

DarkSide666 (approved):
Yes sure if these typehints helps anyone.

PR #1048, opened 2020-03-31
Fix demo

### LINKED ISSUE #987: /demos/table.php - Exception Value should be scalar
Related code:
https://github.com/atk4/ui/blob/ad62e4488da24996016c6f12678d22d9799453b6/demos/table.php#L40

But `\DateTime` value should be supported.

DarkSide666:
1. set Faker as requirement because of demo.
2. fix demo page

fix #987

### DIFF
--- a/demos/table.php
+++ b/demos/table.php
@@ -34,7 +34,7 @@ $table->onHook('getHTMLTags', function ($table, $row) {
     }
     if ($color) {
         return [
-            'name' => $table->app->getTag('div', ['class' => 'ui ribbon '.$color.' label'], $row['name']),
+            'name' => $table->app->getTag('div', ['class' => 'ui ribbon ' . $color . ' label'], $row['name']),
         ];
     }
 });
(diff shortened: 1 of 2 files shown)


ibelar (approved):

demos/table.php
```diff
@@ -26,9 +26,15 @@
 $table->addColumn('logo_url', [new \atk4\ui\TableColumn\Image()], ['caption'=>'Our Logo']);
 
 $table->onHook('getHTMLTags', function ($table, $row) {
-    if ($row->id == 1) {
+    switch ($row->id) {
```
mvorisek, line 29: `$color = [1 => 'yellow', 2 => 'grey', 3 => 'brown'][$row->id] ?? ''` would be much simpler

demos/table.php
```diff
@@ -37,10 +43,10 @@
 
 // ========================================================
 $my_array = [
-    ['name' => 'Vinny', 'surname' => 'Sihra', 'birthdate' => new \DateTime('1973-02-03')],
-    ['name' => 'Zoe', 'surname' => 'Shatwell', 'birthdate' => new \DateTime('1958-08-21')],
-    ['name' => 'Darcy', 'surname' => 'Wild', 'birthdate' => new \DateTime('1968-11-01')],
-    ['name' => 'Brett', 'surname' => 'Bird', 'birthdate' => new \DateTime('1988-12-20')],
+    ['name' => 'Vinny', 'surname' => 'Sihra', 'birthdate' => '1973-02-03'],
```
mvorisek, line 46: From long-term view not a solution to #987 :)

demos/table.php
DarkSide666: well, yes it would :)

Issue #1043, opened 2020-03-31
TravisCI test are broken


mvorisek:
See https://travis-ci.org/github/atk4/ui/builds/669012682

mvorisek:
Issue fixed

PR #1049, opened 2020-03-31
[fix] - Downgrade Fomantic-UI release to 2.7.4

### LINKED ISSUE #1044: Tabs with wrong default active tab
In the most recent version, the tab view chooses automatically the last tab as the default active tab. Prior behaviour was the first tab which makes more sense.

ibelar:
Going back to a atk4/tab friendly Fomantic UI release.

fixes : #1044, #1043

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -45,6 +45,14 @@ class FeatureContext extends RawMinkContext implements Context
     }
 
     /**
+     * @When wait for callback
+     */
+    public function waitForCallback()
+    {
+        $this->jqueryWait(20000);
+    }
+
+    /**
      * @When Wait until loading stops
      */
     public function untilLoadingStops()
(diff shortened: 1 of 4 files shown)


mvorisek (dismissed):
Are we under pressure to solve things by downgrading?

I would highly prefer to keep the latest stable version and solve the real issues. It is also important when others (and we count to them) are using atk with 3rd party frameworks/components that also use Fomantic UI - only by fixing the issues we can ensure compatibility with the latest stable version that should be supported by `develop` branch.

DarkSide666:
Not sure what exactly you mean @mvorisek .

Latest release of Fomantic contain bug. We're not sure when they will solve it, so we have to roll back or Fomantic issue will become our UI issue of our stable branch.

DarkSide666 (approved):
I think we need this while Fomantic don't fix their bug and make new, more stable release.

mvorisek:
> Latest release of Fomantic contain bug.

@DarkSide666 Can we fix it from atk side?

mvorisek:
> Sorry but we do not have any control on Fomantic UI release. Bug is in there package.. We need to downgrade their release version for our Tabs to work.

Please link that bug to the issue #1044.

Can we fix this bug with Javascript? Is there any other major issue that prevent us to keep the new version?

DarkSide666:
@mvorisek Is there something we really, really need in that new Fomantic release? If no, then I don't see a problem to use 2.7.4. If there are a lot of cool features which we could use in atk UI, then ... at this point still - use 2.7.4 because it works :)
Maybe it's possible to make some JS workaround, but why would @ibelar spend time on that?

mvorisek:
@DarkSide666 almost one year of fixes https://github.com/fomantic/Fomantic-UI/releases :) The important ones are for dropdown and modal components. So the latest FUI version can save us all a lot of time not figuring out bugs that are already fixed.

DarkSide666:
Ok. Let @ibelar check if he can work around that tab issue somehow.

ibelar:
Sorry guys, can't do much on this one but wait for Fomantic-UI to fix this. The problem is from the new Tab feature introduced in 2.7.8 when they are initialized. They are aware of the situation. (see https://github.com/fomantic/Fomantic-UI/issues/1255) Not exactly the same issue but caused by the same code. I have also discussed the problem with Lubber from Fomantic UI and he is aware of our situation.

My suggestion is to revert to release 2.7.4 - and those who need a higher release can change it in their app manually. This way our tab and behat test our back to normal.

lubber-de:
I proposed a quick solution in Fomantic by adding a new option which, if disabled, will skip the autodetection of an active tab, so the old behavior will be used.
https://github.com/fomantic/Fomantic-UI/issues/1255#issuecomment-636092402

PR #1051, opened 2020-03-31
Use content type from the source + cleanup of URL methods

### LINKED ISSUE #978: Make App->isJsonRequest() public
Hi,

at the moment this function is protected. However, we might want to know from inside a View if the request is a full page reload or a JSON request. Hence I think this function should be public. Also the name without a leading _ supposes the function is public.

What do you think? any counter-arguments?

### LINKED ISSUE #1040: Content-type is not reset to text/html when an exception is thrown
image

Steps to reproduce:
1. open network tab in developer tools in browser
2. click on any `Lookup` - you should see `200` and `content-type: application/json`, fine
3. now emit (break the code manually) an exception right before the output in the `Lookup` API virtual page
4. repeat step 2. and you should see `500` but still `content-type: application/json` which is wrong when a html error page is outputed

Important for easier debug - when mime is not (re)set to html, error pages are very hard to read.

### LINKED ISSUE #1050: Do not assume anything from "X-Requested-With: XMLHttpRequest" request header
Two totally unrelated things, for ex. a binary image may be requested by AJAX.

Presence of `X-Requested-With: XMLHttpRequest` request header should have no effect on the output.

Affected code:
https://github.com/atk4/ui/blob/3e65d4e3ec0036f1016928179c6c018dd8a03b61/src/App.php#L314-L318

Very important to prevent bad results and make debugging more robust.

### LINKED ISSUE #1059: Modal - rename "mode" to something self-explaining
Code to update:
https://github.com/atk4/ui/blob/b97a304037bcbb748a77aa1a708a64a905968279/src/jsModal.php#L18
and
https://github.com/atk4/ui/blob/b97a304037bcbb748a77aa1a708a64a905968279/js/src/plugins/create-modal.plugin.js#L18

I propose `data_type` which is also use in Fomantic UI API (`dataType`), but feel free to comment better name.

mvorisek:
fixes #1050
fixes #1040
fixes #1059
fixes #978

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -696,12 +696,23 @@ class App
     }
 
     /**
+     * @var bool Set to true to override isJsRequest() when "X-Requested-With" header is equal to "xmlhttprequest".
+     */
+    public $legacyJsRequestDetection = false;
+
+    /**
      * Request was made using jsURL().
      *
      * @return bool
      */
     public function isJsRequest()
     {
+        if ($this->legacyJsRequestDetection) {
+            if (strtolower($_SERVER['HTTP_X_REQUESTED_WITH'] ?? '') === 'xmlhttprequest' && !isset($_GET['__atk_tab'])) {
+                return true;
+            }
+        }
+
         return isset($_GET['__atk_json']) && $_GET['__atk_json'] !== '0';
     }
 
--- a/src/App.php
+++ b/src/App.php
@@ -307,17 +307,7 @@ class App
             return false;
         }
 
-        if (isset($_GET['json'])) {
-            return true;
-        }
-
-        $ajax = false;
-        if (!empty($_SERVER['HTTP_X_REQUESTED_WITH'])
-           && strtolower($_SERVER['HTTP_X_REQUESTED_WITH']) == 'xmlhttprequest') {
-            $ajax = true;
-        }
-
-        return $ajax;
+        return isset($_GET['json']);
     }
 
     /**
--- a/src/App.php
+++ b/src/App.php
@@ -596,6 +596,8 @@ class App
      */
     public function jsURL($page = [], $needRequestUri = false, $extra_args = [])
     {
+        $extra_args = array_merge($extra_args, ['__atk_json' => 1], $extra_args); // append to the end but allow override
+
         return $this->url($page, $needRequestUri, $extra_args);
     }
 
@@ -665,6 +667,8 @@ class App
                 $result[$arg] = $val;
             }
         }
+
+        unset($result['__atk_json']);
 
         // put URL together
         $args = http_build_query($result);
--- a/src/App.php
+++ b/src/App.php
@@ -613,7 +613,7 @@ class App
     public function url($page = [], $needRequestUri = false, $extra_args = [])
     {
         if ($needRequestUri) {
-            return $_SERVER['REQUEST_URI'];
+            $page = $_SERVER['REQUEST_URI'];
         }
 
         $sticky = $this->sticky_get_arguments;
@@ -629,14 +629,17 @@ class App
             }
         }
 
-        // if page passed as string, then simply use it
+        $pagePath = '';
         if (is_string($page)) {
-            return $page;
-        }
-
-        // use current page by default
-        if (!isset($page[0])) {
-            $page[0] = $this->page;
+            $page_arr = explode('?', $page, 2);
+            $pagePath = $page_arr[0];
+            parse_str($page_arr[1] ?? '', $page);
+        } else {
+            $pagePath = $page[0] ?? $this->page; // use current page by default
+            unset($page[0]);
+            if ($pagePath) {
+                $pagePath .= $this->url_building_ext;
+            }
         }
 
         //add sticky arguments
@@ -657,10 +660,6 @@ class App
 
         // add arguments
         foreach ($page as $arg => $val) {
-            if ($arg === 0) {
-                continue;
-            }
-
             if ($val === null || $val === false) {
                 unset($result[$arg]);
             } else {
@@ -668,11 +667,13 @@ class App
             }
         }
 
-        unset($result['__atk_json']);
+        if (!isset($extra_args['__atk_json'])) {
+            unset($result['__atk_json']);
+        }
 
         // put URL together
         $args = http_build_query($result);
-        $url = ($page[0] ? $page[0] . $this->url_building_ext : '') . ($args ? '?' . $args : '');
+        $url = $pagePath . ($args ? '?' . $args : '');
 
         return $url;
     }
--- a/src/App.php
+++ b/src/App.php
@@ -311,16 +311,16 @@ class App
      * other classes.
      *
      * @param string|array $output Array type is supported only for JSON response
-     * @param string|null  $contentType Null for HTML/JSON based on App::isJsRequest()
+     * @param string       $contentType
      *
      * @throws \atk4\core\Exception
      * @throws ExitApplicationException
      */
-    public function terminate($output = null, string $contentType = null): void
+    public function terminate($output = null, string $contentType = 'text/html'): void
     {
-        $type = preg_replace('~;.*~', '', strtolower($contentType)) ?: null; // type in LC without charset
+        $type = preg_replace('~;.*~', '', strtolower($contentType)); // type in LC without charset
 
-        if ($type === 'application/json' || ($this->isJsRequest() && $type === null)) {
+        if ($type === 'application/json') {
             if (is_scalar($output) || $output === null) {
                 $decode = json_decode($output, true);
                 if (json_last_error() === JSON_ERROR_NONE) {
@@ -331,7 +331,7 @@ class App
                 $output['modals'] = $this->getRenderedModals();
             }
             $this->outputResponseJSON($output);
-        } elseif (isset($_GET['__atk_tab']) && ($type === 'text/html' || $type === null)) {
+        } elseif (isset($_GET['__atk_tab']) && $type === 'text/html') {
             // ugly hack for TABS
             // because fomantic ui tab only deal with html and not JSON
             // we need to hack output to include app modal.
--- a/src/App.php
+++ b/src/App.php
@@ -279,7 +279,8 @@ class App
         // remove header
         $this->layout->template->tryDel('Header');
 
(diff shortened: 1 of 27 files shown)


mvorisek:
@ibelar The detection based on `X-Requested-With: XMLHttpRequest` was introduced by you in https://github.com/atk4/ui/commit/cc72579639220ae2b583d893e16fda240bd5b1ea .

Can you please review this PR? Feel free to add changes directly when needed.

ibelar (changes requested):
What do you propose then? 
We still need a way to detect if the request is a json request. Just checking for url parameter 'json' is not enough. 
For example all jsCallback, like in form submit, will return the wrong type if you remove this check.

Furthermore, this does not affect the output but rather check if the request is of json type.

mvorisek:
The `isJsonRequest` method is used for output control like here:
https://github.com/atk4/ui/blob/3e65d4e3ec0036f1016928179c6c018dd8a03b61/src/App.php#L522
so it affects the output.

I propose using what is needed, but strictly based on GET or POST params.

ibelar:
@mvorisek  - Get and post param will not be enough - unless refactoring a lot of UI js stuff. So, unless you have a better way to detect if the request is of Json type, then we can close this.

What are you trying to achieve by removing this check? For which reason?

mvorisek:
> What are you trying to achieve by removing this check? For which reason?

For me this concept is very fragile and limits the adoption of atk for wider applications.

I got caught myself wih this - I copied some failing URL to browser from the browser network console and it worked!

We should honor concept of industry standard APIs and avoid solutions like this.

ibelar:
@mvorisek - Understand. Please link the issue properly has to where it is causing problem and we will see what can be done. I am afraid the change you are proposing will have other impacts.

mvorisek:
@ibelar Everywhere, really, we need to remove this detection from the code.

mvorisek:
> Get and post param will not be enough

@ibelar, what are the individual use cases where it is not enough?

PhilippGrashoff:
just a side thought: "json" is a very common string. I can easily think of scenarios where $_GET['json'] might be set outside of the original atk usage. Other special ATK URL params start with double underscores. e.g. `__atk_callback`. Maybe we should change "json" to `__atk_json` so accidently having this GET parameter set outside our meaning is much less likely?

or is "json" set somewhere out of our reach, e.g. by FUI?

mvorisek:
@PhilippGrashoff Totally agree!, step by step :), I will submit a PR for it shortly

mvorisek:
> We still need a way to detect if the request is a json request. Just checking for url parameter 'json' is not enough.
> For example all jsCallback, like in form submit, will return the wrong type if you remove this check.
> 

I have carefully review this.

The current code in `App->isJsonRequest()` for json detection is actually hack in hack :)
https://github.com/atk4/ui/blob/b97a304037bcbb748a77aa1a708a64a905968279/src/App.php#L304-L321
as when the AJAX header detection is used, is can not be used for `Tab` as json type is not supported.

To your `jsCallback` comment:
If the response should be of type json is actually already supported with `jsURL()` and for `jsCallback` it is alread used:
https://github.com/atk4/ui/blob/b97a304037bcbb748a77aa1a708a64a905968279/src/jsCallback.php#L68

As `Tab` use `url()` instead of `jsURL()`:
https://github.com/atk4/ui/blob/b97a304037bcbb748a77aa1a708a64a905968279/src/Tab.php#L25
thus the response will not be of type json as required.

I belive this solution is the only real one.

src/App.php
```diff
@@ -295,31 +295,6 @@ public function caughtException(Throwable $exception)
         return true;
     }
 
-    /**
-     * Most of the ajax request will require sending exception in json
-     * instead of html, except for tab.
-     *
-     * @return bool
-     */
-    protected function isJsonRequest()
```
PhilippGrashoff, line 304: Even though this function would result in a one-liner, I like having this function. Typing `$app->isJsonRequest()` seems better than having to put `isset($_GET['__atk_json'])` everywhere. Think of IDE hinting :)

PhilippGrashoff: see also #978, WDYT?

mvorisek: I removed it intentionally as it might not be valid all the times, i.e. when the current URL was not made by `getJSURL`.

But the more I worked on this PR, the more I understand the tight integration of HTML/JS URLs/requests in `App`. So to make everyone happy I will restore it, rename it to `isJsRequest` and make it public. Win-win?

ibelar:
@mvorisek - I appreciate the effort you are putting in. But please make sure this is still valid.

````
$ajax = false;	
        if (!empty($_SERVER['HTTP_X_REQUESTED_WITH'])	
           && strtolower($_SERVER['HTTP_X_REQUESTED_WITH']) == 'xmlhttprequest') {	
            $ajax = true;	
        }
````

As of now, this is the way to check if a request is coming from atkjs-ui.js. This js bundle use Fomantic API call in a lot of places: 
- ajaxec, 
- reload-view, 
- js-search, 
- upload-service
- scroll,
-  vue-service

All of these calls will return bad content type if not check this way and there is no time to refactor atk-js as this time. Furthermore, our own Saasty project, build on atk4/ui rely on components, via an external js bundle, that use Fomantic API call directly. There is no time for refactoring there also.

Besides, I am still unsure about what you are trying to resolve with this PR. Maybe there are another way to look at it. Could you please provide a reproducible issue for it?

**Comment by mvorisek** [comment]
@ibelar Please provide examples with code and link directly the code parts from git where AJAX requests are made and `getJSURL()` is not used or impossible to use.

I am not a frontend guy, but using detection based on `HTTP_X_REQUESTED_WITH` in hope that the request need to be returned in JSON for the whole framework is absolutely againt all rules of web standards.

If your examples will not be easily fixable, I can add some special property/switch to the App so the edge cases can be covered and when not needed, it can be deactivated.

**Comment by PhilippGrashoff** [comment]
after googling HTTP_X_REQUESTED_WITH, it seems to be a de-facto standard used by all major JS frameworks. No official RFC for it, but it seems its used widely.

**Comment by mvorisek** [comment]
@PhilippGrashoff Yes, all browsers send it, but backend/API should not work differently on it - is that what you have googled?

**Comment by ibelar** [comment]
@mvorisek - Well - all API call in js file mentioned above to start with. Then all API call we used in saasty which depend on atk4/ui. There is also a place in Form where the API call is rendered directly.
https://github.com/atk4/ui/blob/b97a304037bcbb748a77aa1a708a64a905968279/src/Form.php#L624

Also, everyone using direct API call from within their app will be impacted if they do not append the proper URL parameter to it. 

I am not against refactoring, just saying that we cannot just remove the check to `HTTP_X_REQUESTED_WITH` and think all will be fine. Plus there is no urgent need to fix this issue, so I would suggest that we take the time to do it.

My first guess would be to check each $app-terminate call. This is really where we control the kind of output necessary and we could generate proper action from there, probably without even need to add the extra JSON parameter.

**Comment by PhilippGrashoff** [comment]
well, I googled that it seems that major PHP frameworks like Laravel use it, too. I agree with you that using some non-standard header that might not be sent along with the request ain't a perfect solution, but it seems to be working quite well across the WWW.

@ibelar WDYT about renaming "json" to "__atk_json"? Could that be done easily without interfering somewhere?

**Comment by mvorisek** [comment]
https://github.com/atk4/ui/blob/b97a304037bcbb748a77aa1a708a64a905968279/src/Form.php#L624
can you provide a full example - you mean for ex. `jQuery.ajax()`? This is of course unware of any URL requirements, but as long as you send the request to URL generated by `getJSURL()`, it fully works as of this PR.

**Comment by mvorisek** [comment]
> My first guess would be to check each $app-terminate call. This is really where we control the kind of output necessary and we could generate proper action from there, probably without even need to add the extra JSON parameter.

This is very good point.

**Comment by mvorisek** [comment]
@ibelar Can you please provide feedback? I think I have isolated the issues. The main issue was that there was no typing of the response internally (i.e. no type for `terminate()`) and everything was guessed. And the `X-Requested-With` header seems to be ok for uncaught exceptions, but strictly for that only (as response type is unknown before render is complete).

File: src/VirtualPage.php [hunk 2]
```diff
@@ -118,26 +118,26 @@ public function getHTML()
...
 
                 // special treatment for popup
-                if ($type == 'popup') {
+                if ($type === 'popup') {
                     $this->app->html->template->set('title', $this->app->title);
                     $this->app->html->template->setHTML('Content', parent::getHTML());
                     $this->app->html->template->appendHTML('HEAD', $this->getJS());
 
-                    $this->app->terminate($this->app->html->template->render());
+                    $this->app->terminateHTML($this->app->html->template);
```
ibelar, line 126: this is causing an issue. You should probably sending the rendered template here.
```
$this->app->terminateHTML($this->app->html->template->render());
````
see demos/modal.php on 'Virtual Page Logic' section.

File: src/VirtualPage.php [hunk 3]
```diff
@@ -154,7 +154,7 @@ public function getHTML()
             $this->app->html->template->setHTML('Content', $this->app->layout->getHTML());
             $this->app->html->template->appendHTML('HEAD', $this->app->layout->getJS());
 
-            $this->app->terminate($this->app->html->template->render());
+            $this->app->terminateHTML($this->app->html->template);
```
ibelar, line 157: same here!

ibelar (changes requested):
Beside VirtualPage rendering, all look very good! Excellent job!

One more thing. if you can bump up the release in js/package.json, say 1.12.6 and explain change in release.md file would be good. If not, no worry, I can do it later.

src/VirtualPage.php
mvorisek: Good catch! That is exactly why I already opened issue #989. Fix is not that easy (`render()` methods need to be unified etc.,), so I fixed it in `App::terminateHTML()` directly - https://github.com/atk4/ui/pull/1051/commits/2a9de48d14c8938669d59c8ac90b032026fa6099

src/VirtualPage.php
mvorisek: same here, fixed!

ibelar (approved):

mvorisek:
> One more thing. if you can bump up the release in js/package.json, say 1.12.6 and explain change in release.md file would be good.

done

PR #1053, opened 2020-03-31
Fix set null for non-nullable fields

### LINKED ISSUE #1052: Lookup - set(null) can fail
See https://github.com/atk4/data/issues/542

Issue is triggered when the new `dependency` feature is used.

mvorisek:
Fixes #1052


DarkSide666 (approved):
Looks like workaround, but I guess in this case it's ok to bypass field value validation this way.

PR #1055, opened 2020-04-01
Remove @param phpdoc without type nor comment

mvorisek:


DarkSide666 (commented):
I think these should be fixed not removed :)

mvorisek:
@DarkSide666 Fixing them is optimal, but these currently do not provide any value and most of them are even wrong (no type at all or type `null`)

DarkSide666:
Well, you're right about if no type is set for @param then it's kind of useless, BUT:

1. type `null` could be legit - means that parameter is not used in this method, but in parent method it maybe is used (and can't be simply removed)

2. Looks like you have done some regex replacement in this PR because in some places you remove 2 of 4 @param lines and that looks even worse than wrong ones :)
For example, https://github.com/atk4/ui/pull/1055/files#diff-b62ed7d502ea3e363136f3005a786ebaL340 (btw removing also legit `Generic` param).

mvorisek:
@DarkSide666 Looked up problematic lines by several regexes, but then I editted them all manually.

https://github.com/atk4/ui/pull/1055/files#diff-b62ed7d502ea3e363136f3005a786ebaL341-L342
- The first line is redundant (as the type is enforced in method header) and usually removed by CS fixers
- The 2nd line is example of `null` type - your point about no usage of these pamameters is interesting. I double checked the PR and this is not the case - the nulls in phpdocs were really put there wrongly, probably somehow from the default values in the method declaration.

What I have also noticed that there are a lot of checks if arg is of some type like `callable` with some more or less bad Exception/message. Is it ok to remove them against type in the method signature (which will throw `TypeError`)?

Can you review this PR?

DarkSide666:
> What I have also noticed that there are a lot of checks if arg is of some type like callable with some more or less bad Exception/message. Is it ok to remove them against type in the method signature (which will throw TypeError)?

If only option for arg is to be of specific type then i think it's OK to set this type in method signature. Although need to keep backward compatibility as much as possible!

But in many cases we accept multiple types of arguments, like if arg ir array, then ... if arg is callable, then ... In such case i believe method signature can't be used and only way to describe such args is in phpdocs @param comment.

DarkSide666 (changes requested):
I still don't see what improvement this PR does.
It removes some incorrect comments and that's all. If it would fix these comments or method signatures where possible then it's another story...

mvorisek:
@DarkSide666 It removed code that provides zero value or even provide negative value, when the phpdocs hints are wrong.

romaninsh:
hm. how was this merged without approvals?

mvorisek:
> 
> 
> hm. how was this merged without approvals?

I merged it as the changes fix basically a typos / invalid phpdocs. Removing unneed phpdocs is defacto standard and it is also done by `CS fixer` by default - see `no_superfluous_phpdoc_tags` rule in standard `@Symfony`, `@PhpCsFixer` rulesets.

PR #1064, opened 2020-04-03
Remove App::outputDebug() method

mvorisek:
Methods like this should be avoided. There is no escaping.

If needed, use `echo "DEBUG: $str<br />";` directly which is much more clear what id does and how dangerous it is.


DarkSide666 (commented):
I agree. Would be good to wait for @romaninsh opinion too.

DarkSide666 (approved):

romaninsh (approved):
I'm not sure what was this for originally.

PR #1065, opened 2020-04-03
Improve HTTP headers output

### LINKED ISSUE #1074: Remove SSE class
SSE class seems to be obsolete and it is not even tested. What is the purpose of it?

There is very similar jsSSE class which is used in demos and tested.

If SSE has no special puspose, remove it.

mvorisek:
Complete rework of HTTP headers. HTTP headers can be easily set in `terminate()` / `output()` methods as well at any stage using `App->setResponseHeader()` method.

- Headers are property checked if they can be outputed (i.e. if no content was sent yet).

- Compatible with SSE.

- HTTP response/status code can be newly and very easily set as well.

- Fixes TODOs like https://github.com/atk4/ui/blob/bd95949954fbb86766b3c3d78d1e95c5ea57ca8c/src/App.php#L999
- Fixes/reverts usage of `flush()` introduced by me a few days ago: https://github.com/atk4/ui/pull/1051/files#diff-58980ac23a3a9a0b92abd9805eed1e78R227
  it is not compatible with built-in PHP server, probably a PHP bug - when `flush()` is called, `headers_sent()` returns `true` even if no content was outputed.
- Fixes #1074


src/App.php
```diff
@@ -987,47 +1072,71 @@ function () {
...
-    private function outputResponseHTML(string $data): void
+    private function outputResponseJSON($data, array $headers = []): void
     {
-        $this->outputResponse(['Content-Type: text/html' => true], $data);
+        if (!is_string($data)) {
+            $data = $this->encodeJson($data);
+        }
+
+        $this->outputResponse(
+            $data,
```
abbadon1334, line 1137: can be replaced like this, no? `!is_string($data) ? $this->encodeJson($data) : $data`

abbadon1334 (approved):
Only that comment, the rest looks really good.
Thanks for all the work @mvorisek

src/App.php
mvorisek: Why to assing $data = $data?

js/Release.md
```diff
@@ -1,5 +1,9 @@
 ## Release note
 
+### version 1.12.6 (2020-04-08)
```
DarkSide666: wrong version number. should be 1.12.7

DarkSide666 (changes requested):

src/App.php
```diff
@@ -302,32 +306,87 @@ public function outputDebug($str)
...
+            }
+
+            $res[strtolower(trim($k))] = trim($v);
+        }
+
+        return $res;
+    }
+
+    /**
+     * @return $this
```
DarkSide666, line 331: no comment what this method does.

src/App.php
```diff
@@ -302,32 +306,87 @@ public function outputDebug($str)
...
+     */
+    public function setResponseStatusCode(int $statusCode): self
+    {
+        $this->setResponseHeader(self::HEADER_STATUS_CODE, $statusCode);
+
+        return $this;
+    }
+
+    /**
+     * @return $this
```
DarkSide666, line 341: no comment what this method does.

src/App.php
```diff
@@ -763,12 +830,9 @@ public function requireCSS($url)
      * @throws \atk4\core\Exception
      * @throws ExitApplicationException
      */
-    public function redirect($page)
+    public function redirect($page, bool $permanent = false): void
     {
-        header('Location: ' . $this->url($page));
-
-        $this->run_called = true; // prevent shutdown function from triggering.
```
DarkSide666, line 770: i remember this `run_called` property was used somewhere in legacy project but probably it's OK to not use it there.

src/App.php
mvorisek: I am very strictly against any duplication/copying of the same - writing there "set Response Status Code" from the method header "setResponseStatusCode" is dangerous when the method is refactored, it takes only real-estate in the code and does provide zero value. So no single advantage.

What is the purpose of it?

src/App.php
mvorisek: Yes, and the new terminateHTML calls terminate which sets it as previously, only at one place. The same again, the more you duplicate, the more you have to worry about :D

js/Release.md
mvorisek: Good catch, fixed.

src/App.php
mvorisek: same as above

src/App.php
DarkSide666: About duplicating you're definitely right. I mean - code duplicating.
But comments are useful when you check code and to easily understand what each method does. Normally it would be very good doc where all methods are described, but toolkit don't have such extensive documentation, so quite often I find myself reading toolkit code instead to understand how it works :)

DarkSide666 (approved):
LGTM

PR #1067, opened 2020-04-04
Cleanup coverage testing

mvorisek:
Not sure why there was a `sleep` command - if some files are flushed not immediatelly (like from webserver, we should cleanly shutdown the process first).

Before:
image

After:
image

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -50,7 +50,7 @@ jobs:
           mysql -uroot -ppassword -h mysql atk4 < demos/atk4.sql
 
           mv demos/db.github.php demos/db.php
-          (cd js; npm install; npm run build)
+          (cd js; npm install --loglevel=error; npm run build)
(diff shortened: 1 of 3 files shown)


mvorisek:
The `sleep` is there intentionally - coverage decreased even if it should not - please advise how to cleanly restart the webserver before coverage analysis - this should fix the issue.

abbadon1334:
if i remember correctly behat needs some time to initialize

mvorisek:
> 
> 
> if i remember correctly behat needs some time to initialize

Behat is called before the `sleep`.

abbadon1334:
> > if i remember correctly behat needs some time to initialize
> 
> Behat is called before the `sleep`.

if i remember correctly, behat starts and cannot be called immediatly for that there was a sleep after start.

abbadon1334 (approved):
i like this solution

romaninsh (commented):
Thanks!

PR #1070, opened 2020-04-05
[fix] UserAction Modal Id

ibelar:
Modal create by UserAction can be added via callback request. Therefore, we can not rely on regular element number id mechanism for generating their id. 

This will make sure HTML modal id are set properly.


DarkSide666 (approved):
LGTM

PR #1075, opened 2020-04-07
Emit status code = 500 on error

mvorisek:
Industry standard and it seems that it can also catch errors in demos/tests.

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -537,7 +537,7 @@ class App
                 $this->setResponseStatusCode(500);
                 $this->terminate(
                     $this->buildLateErrorStr('Callback requested, but never reached. You may be missing some arguments in request URL.'),
-                    ['content-type' => 'text/plain', self::HEADER_STATUS_CODE => 500]
+                    ['content-type' => 'text/plain']
                 );
             }
             echo $this->html->template->render();
--- a/src/App.php
+++ b/src/App.php
@@ -275,6 +275,7 @@ class App
                 'message' => $this->layout->getHtml(),
             ]);
         } else {
+            $this->setResponseStatusCode(500);
             $this->run();
         }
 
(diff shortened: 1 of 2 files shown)


mvorisek:
Based on PR #1065. Rebase once that PR is merged.

Please help with fixing the test.

src/App.php
```diff
@@ -302,32 +307,87 @@ public function outputDebug($str)
...
      *
      * @param string|array $output Array type is supported only for JSON response
-     * @param string       $contentType
+     * @param string[]     $headers Content-type header must be always set or consider using App::terminateHTML() or App::terminateJSON() methods.
      *
      * @throws \atk4\core\Exception
      * @throws ExitApplicationException
      */
-    public function terminate($output = null, string $contentType = 'text/html'): void
+    public function terminate($output = '', array $headers = []): void
```
romaninsh: @ibelar - will this break anything?

src/App.php
```diff
@@ -987,47 +1074,90 @@ function () {
...
+            ob_end_flush();
+            self::$_x[] = [$headersNew, $trace, headers_list(), http_response_code()];
         }
 
-        echo $content;
+        if (count($headersNew) > 0 && headers_sent()) {
+            echo "\n" . '!! ATK4 UI ERROR: Headers already sent, more headers can not be set at this stage. !!' . "\n";
+
+            // debug
+            echo '--> Mvorisek:' . "\n";
```
romaninsh: should this be cleaned up?

romaninsh (approved):
I approve, but @ibelar should also review.

src/App.php
mvorisek: It should not and null should never be allowed here. To not output anaything use nothing or empty string.

src/App.php
mvorisek: @romaninsh Yes, it should, see my comment:

> Based on PR #1065. Rebase once that PR is merged.

So please review the PR #1065 first :)

mvorisek:
Help wanted for the failing tests.

PR #1076, opened 2020-04-08
Use null coalescing operator instead of isset

mvorisek:


romaninsh (approved):

src/Paginator.php
```diff
@@ -94,12 +94,10 @@ public function setTotal($total)
...
      * Determine and return the current page. You can extend this method for
      * the advanced logic.
-     *
-     * @return int
      */
-    public function getCurrentPage()
+    public function getCurrentPage(): int
     {
-        return isset($_GET[$this->urlTrigger]) ? (int) $_GET[$this->urlTrigger] : 1;
+        return $_GET[$this->urlTrigger] ?? 1;
```
DarkSide666, line 100: This is not the same, no (int) casting !!!

src/Persistence/UI.php
```diff
@@ -126,8 +126,8 @@ public function _typecastLoadField(\atk4\data\Field $f, $value)
         case 'date':
         case 'datetime':
         case 'time':
-            $dt_class = isset($f->dateTimeClass) ? $f->dateTimeClass : 'DateTime';
-            $tz_class = isset($f->dateTimeZoneClass) ? $f->dateTimeZoneClass : 'DateTimeZone';
+            $dt_class = $f->dateTimeClass ?? \DateTime::class;
+            $tz_class = $f->dateTimeZoneClass ?? \DateTimeZone::class;
```
DarkSide666, line 130: this change behaviour if `dateTimeClass` was defined as `false` or empty string for example.

src/Modal.php
```diff
@@ -94,7 +94,7 @@ public function enableCallback()
             if ($this->cb->triggered() && $this->fx) {
                 $this->fx[0]($this->cb_view);
             }
-            $modalName = isset($_GET['__atk_m']) ? $_GET['__atk_m'] : null;
+            $modalName = $_GET['__atk_m'] ?? null;
```
DarkSide666, line 97: this change behavior if `$_GET['__atk_m'] == 0` (shouldn't be, but still)

src/App.php
```diff
@@ -911,7 +911,7 @@ public function getTag($tag = null, $attr = null, $value = null)
             $tag = substr($tag, 0, -1);
             $postfix = '/';
         } elseif (substr($tag, 0, 1) == '/') {
-            return isset($attr[0]) ? '</' . $attr[0] . '>' : '<' . $tag . '>';
+            return '</' . ($attr[0] ?? substr($tag, 1)) . '>';
```
DarkSide666, line 914: This is changing (wrong?) change at all. What if $attr[0] wasn't defined? Previously it output `<$tag>`, now it will be `</$tag>` !? isn't it?

romaninsh:
@DarkSide666 should this be un-merged and revisited?

src/Modal.php
mvorisek: @DarkSide666 Really? `isset` and `??` should work the same.

Test code:
```
$in = [false];
var_dump(isset($in[0]) ? $in[0] : 'x');
var_dump($in[0] ?? 'x');
```

src/Paginator.php
mvorisek: I done it using the return type ;-)

src/Persistence/UI.php
mvorisek: The same as above, `??` should be equal to `isset`. (in oposite to `?:` or `!empty`)

src/App.php
mvorisek: Can you provide a test code?

src/Modal.php
romaninsh: i think it should still work identical. ?? implies isset.

romaninsh: unless I'm wrong and it was ?: ... :O

romaninsh:
Ok, had a chat with @DarkSide666 about https://www.designcise.com/web/tutorial/whats-the-difference-between-null-coalescing-operator-and-ternary-operator-in-php and this PR looks good.

src/App.php
romaninsh: @mvorisek ^^ this seems broken. Can we revert this bit here? currently it should return '<tag>' but in your case '</..' is always returned.

src/Persistence/UI.php
romaninsh: false should still work, but it should also solve issues with implied namespaces..

src/Paginator.php
romaninsh: yes please change to `return (int)($_GET.... ?? 1)`

src/App.php
romaninsh: oh i didn't notice the substr, so this should be ok !!

src/Persistence/UI.php
mvorisek: > implied namespaces

There should not be any, can you provide more details? (but same here, this PR produces exactly the same string as the hardcoded one before ;-))

src/Paginator.php
romaninsh: oh awesome then!]

src/App.php
mvorisek: I can patch it if needed, what is the issue, can you provide a simple code/data to test the issue?

mvorisek: @romaninsh Yes, seems much better to me, I hate defining anything twice. :D

PR #1082, opened 2020-04-09
Render all exceptions using App

mvorisek:


abbadon1334:
Code is ok, but i suggest @ibelar to check this much deeper.

src/ActionExecutor/UserAction.php
```diff
@@ -611,7 +611,7 @@ protected function jsStepSubmit(string $step)
         } catch (\Exception $e) {
             $m = new Message('Error executing ' . $this->action->caption, 'red');
             $m->init();
-            $m->text->content = ($e instanceof \atk4\core\Exception ? $e->getHTML() : $e->getMessage());
+            $m->text->content = $this->app->renderExceptionHTML($e);
```
romaninsh, line 614: This implies existence of $this->app, btw - as a note - ATK Data does not expect APP class instance to exist, but it should be OK here.

romaninsh (approved):

src/ActionExecutor/UserAction.php
mvorisek: Yes, we are in UI repo here.

PR #1088, opened 2020-04-13
[fix] UserConfirmation executor + Behat test

ibelar:
- Fix UserConfirmation action executor
-  Add Behat test for UserAction executor
    - cleanup test file and scenario
- Move Country action into is own file

Screen Shot 2020-04-13 at 9 29 23 AM


src/ActionExecutor/UserConfirmation.php
```diff
@@ -184,7 +235,16 @@ protected function doFinal(View $modal)
...
+    /**
+     * Return proper js statement when action execute.
+     *
+     * @param $obj
+     * @param $id
+     *
+     * @return array
+     * @throws \atk4\core\Exception
+     */
+    protected function jsGetExecute($obj, $id): array
```
PhilippGrashoff, line 247: At least in many other atk files, protected/private functions are prefixed with an underscore, like `_jsGetExecute`. Is that still the way to go? Personally I like this naming convention.

ibelar: @PhilippGrashoff  - My understanding is that we are moving toward PSR-12 standard which forbid using a single underscore char in front of a method name.

> Method names MUST NOT be prefixed with a single underscore to indicate protected or private visibility. That is, an underscore prefix explicitly has no meaning.

PhilippGrashoff: Ah, I was not aware of this rule in PSR12. Thanks for the explanation!

src/ActionExecutor/UserConfirmation.php
```diff
@@ -50,6 +51,27 @@ public function init()
...
+     */
+    public function afterActionInit(Generic $action)
+    {
+        $getTableName = function ($arr) {
+            foreach ($arr as $k => $v) {
+                return is_numeric($k) ? $v : $k;
+            }
+        };
+
+        $table_name = is_array($action->getModel()->table) ? $getTableName($action->getModel()->table) : $action->getModel()->table;
```
DarkSide666, line 71: How model->table can be of type array ?

ibelar: @DarkSide666 - I got this from property param type
https://github.com/atk4/data/blob/d2d83c2e82ce23bdb8c61d99b78db177ee0da6f0/src/Model.php#L131

Is it still the case? If array is not supported then I can change it.

DarkSide666: I think it never is used as array and what's written in param description wasn't actually developed. But OK, let's make sure we also support array here for possible future use.

DarkSide666 (approved):

PR #1085, opened 2020-04-10
[Feature] Right layout panel

ibelar:
Implementation of Panel\Right

Display on-demand content via a right-side panel. It can display static or dynamic content via callback.
Panel is displayed (show) via a javascript event and content can be loaded via callback. The callback URL may also carry data from the triggering element, allowing to load model data into panel.

Panel look and size, as well as panel content, can be customized via HTML template or css class definition.

Sample usage:
``` php
$panel_1 = $app->layout->addRightPanel(new \atk4\ui\Panel\Right());
$btn = Button::addTo($app, ['Button 1']);
$btn->js(true)->data('btn', '1');
$btn->on('click', $panel_1->jsOpen(['btn'], 'orange'));

$btn = Button::addTo($app, ['Button 2']);
$btn->js(true)->data('btn', '2');
$btn->on('click', $panel_1->jsOpen(['btn'], 'orange'));

$panel_1->onOpen(function($p) {
    $btn_number = $_GET['btn'] ?? null;
    $text =  'You loaded panel content using button #' . $btn_number;
    Message::addTo($p, ['Panel 1', 'text' => $text]);
});
```
###  Note:
see: demos/layout-panel.php for more info.
required atkjs-ui version 1.13.0

Screen Shot 2020-04-10 at 2 01 36 PM

PR #1089, opened 2020-04-13
Fix init(), setDefaults(), normalizeClassNameApp() methods headers

mvorisek:

PR #1092, opened 2020-04-14
Fix init() method header

mvorisek:
Fixes https://github.com/atk4/ui/pull/1085 to be compatible with https://github.com/atk4/ui/pull/1089
Fixes tests

### DIFF
--- a/src/Panel/Content.php
+++ b/src/Panel/Content.php
@@ -13,7 +13,7 @@ class Content extends View implements LoadableContent
     public $defaultTemplate = 'panel/content.html';
     public $cb = null;
 
-    public function init()
+    public function init(): void
     {
         parent::init();
         $this->addClass('atk-panel-content');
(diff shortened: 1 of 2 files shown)


mvorisek:
@ibelar Do you know why the tests failed: https://travis-ci.org/github/atk4/ui/jobs/675062565#L546 and how to fix it?

Wdyt about https://github.com/atk4/ui/pull/1041 ? Currently html/DOM source and printscreen is dumped, is that enought or should somemething additonal be dumped?

ibelar:
> Wdyt about #1041 ? Currently html/DOM source and printscreen is dumped, is that enought or should somemething additonal be dumped?

@mvorisek - Yes.  I think that should be enough.  Tks!
Also issue a fix for Behat test causing issue in #1096 and I also had to include changed in this PR.

mvorisek:
Closing, fully fixed in #1096

PR #1095, opened 2020-04-14
Replace json_decode with App::decodeJson() and json_encode with App::encodeJson()

mvorisek:


DarkSide666 (approved):
This looks correct way to do decoding. Also if some change is needed in future then it will all use same method in one place.

mvorisek:
I am not exactly sure what with this as app is not guarantted to always exist.

@DarkSide666 is it ok to always require an App to be set for json_decode/encode? I prefer it, but is there any strong usecase againt it (like a lot of people use some components without App)?

PS: of course, anyone can still later create an dummy App, so it should be not a big issue

DarkSide666:
I think in UI app always exists.

mvorisek:
> It requires App and there is currently no function like default App anywhere

> I think in UI app always exists.

ok, reopening again, worth to check, but some isolated (not connected with App) renderings are done...

mvorisek:
replaced with https://github.com/atk4/ui/pull/1279

PR #1100, opened 2020-04-15
[fix] - Setting value on DropdownCascade

### LINKED ISSUE #1091: DropDownCascade with error
Line 98 has to be 'set value' instead of 'change values', otherwise the default value is not properly shown... @mvorisek @ibelar 

        if (!$this->cascadeInput->model->loaded()) {
            $this->js(true)->dropdown('change value', $this->getNewValues($this->cascadeInputValue));
        }

ibelar:
Fix: #1091


DarkSide666 (approved):
LGTM

PR #1106, opened 2020-04-15
[fix] Add css and js file to public dir

ibelar:
Add js file and public css for develop branch


mvorisek:
@ibelar Are the files tested to match the newly compiled ones in tests?

PR #1109, opened 2020-04-16
[fix] CRUD::addModalAction to use appSticky argument on Modal callback

ibelar:
Make CRUD::addModalAction() method to use app stickyCallback by default.

It will enable CRUD inside a modal callback function, when add via addModalAction, to being able to run by default:

````
$crud->addModalAction(['', 'icon'=>'eye'], 'Detail', function ($page, $id) use($crud, $app) {
    $container = $page->add(['ui' => 'basic segment']);
    $card = $container->add(new \atk4\ui\Card);
    $card->setModel($crud->model->load($id), ['name']);
    $children = $crud->model->ref( 'Children');

    $child_crud = $container->add('CRUD', ['paginator' => false]);
    $child_crud->setModel($children);
});
````

### DIFF
--- a/src/Modal.php
+++ b/src/Modal.php
@@ -39,6 +39,9 @@ class Modal extends View
     public $cb_view = null;
     public $args = [];
 
+    /** @var bool Make callback url argument stick to application or view. */
+    public $appStickyCb = false;
+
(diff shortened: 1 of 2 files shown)


DarkSide666 (approved):
looks good

PR #1111, opened 2020-04-17
App::addStyle should not encode value

### LINKED ISSUE #1110: App::addStyle content is encoded with htmlentities
When using `App::addStyle` symbols in css like `>` are converted to `&gt;`

This is caused by using `App::getTag` to create the style tag where `htmlentities` is used to encode tag content.

https://github.com/atk4/ui/blob/6cdd0112509b794302f12ab5427f2a856e9452da/src/App.php#L931

DarkSide666:
`App::addStyle` should not encode CSS style string because it could contain symbols like `>` (`tr>td` for example).

But to avoid at least some _attacks_ we still should use `strip_tags` instead.

Consider this example:
```
App::addStyle('</style><script>alert("Hacker here");</script><style>');
```

fixes #1110

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -481,7 +481,7 @@ class App
         if (!$this->html) {
             throw new Exception(['App does not know how to add style']);
         }
-        $this->html->template->appendHTML('HEAD', $this->getTag('style', $style));
+        $this->html->template->appendHTML('HEAD', $this->getTag('style', null, $style, false));
     }
 
     /**
@@ -882,10 +882,11 @@ class App
      * @param string|array $tag
      * @param string       $attr
      * @param string|array $value
+     * @param bool         $encodeValue
      *
      * @return string
      */
-    public function getTag($tag = null, $attr = null, $value = null)
+    public function getTag($tag = null, $attr = null, $value = null, bool $encodeValue = true)
     {
         if ($tag === null) {
             $tag = 'div';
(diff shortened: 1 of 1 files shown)


georgehristov (approved):
LGTM

mvorisek:
This fix should be reverted.

@DarkSide666 can't `>` be encoded to `&gt;` or with backslash https://www.w3.org/TR/css-syntax-3/#escaping ?

Otherwise we may want to remove all `<` and leave `>` and keep at least the header.

georgehristov:
@mvorisek what is the problem you see the way it is now?
The `>` is needed in some css statements and not html encoded `&gt;`.
And generally the content between `<style>` tags is not html and should not be encoded

mvorisek:
@georgehristov What is the `$encodeValue` param there if it can not be used for most of the cases and if used, then it may behave wrong - It seems like very unuseful parameter to me. Wdyt?

georgehristov:
To me the functionality is necessary as you have some tags in which the contents should not be encoded (think `<style>`, `<script>`, etc).
So for `getTag` to be completely versatile this parameter is necessary. 
And having it as optional parameter with `true` as default value which covers most use cases is a good solution IMHO.

mvorisek:
What is the use case of `$this->encodeHTML(strip_tags($value));` ?

DarkSide666:
Oh I have a bug there.
It supposed to be like this:
```
$value = $encodeValue ? $this->encodeHTML($value) : strip_tags($value);
```

Issue #1028, opened 2020-03-29
Some templates are missing common properties like id


mvorisek:
Templates like `admin.html/pug` are missing `id` property.

Is that intentional?

Element properties like `id`, `data`, ... has almost nothing to do with styling and if user/programmer wants to add them, it should be always possible no matter if defined in the template or not.

This is a feature request to introduce some special template parameter like `{$implicit_props}` which can be easily added to all templates and then the common html properties can be added with one centralized code from `View`. Another option might be to always parse the template and set the common properties by DOM (which is not easy, as templates might not be always be valid html).

romaninsh:
Some of those properties were added on “need to” basis.

DarkSide666:
Also it's possible that some templates do not need these id because they are not used there or template is part of another template.

mvorisek:
@DarkSide666 I spotted this when I needed to work with main layout - there is no id and also `jQuery` without custom selector did not worked, as the id does not exist and can not be currently added.

Quite legit use case.

DarkSide666:
Yes sure. I'm not against it. We should add `{$_id}` in all of them.

mvorisek:
templates should be redesigned to be parsed into DOM - we can then manupulate the DOM in init as needed then

PR #1112, opened 2020-04-17
[feature] - Admin Layout

ibelar:
Refactor Admin Layout
- Application name always visible;
- Left menu always same width;
- Width configurable via less;
- Reorganize demos menu and buttons

#### Note
Also refactor some demo file for faster loading.

Before: 
Screen Shot 2020-04-17 at 1 10 09 PM

After:
Screen Shot 2020-04-17 at 1 10 23 PM
 
### App name always visible when menu scrolls up:

Before:
Screen Shot 2020-04-17 at 1 14 07 PM

After:
Screen Shot 2020-04-17 at 1 14 23 PM

### Menu width always the same

Before:
Screen Shot 2020-04-17 at 1 15 55 PM

After:
Screen Shot 2020-04-17 at 1 16 27 PM

### DIFF
--- a/demos/init.php
+++ b/demos/init.php
@@ -115,6 +115,7 @@ if (file_exists('coverage.php')) {
 $app->title = 'Agile UI Demo v' . $app->version;
 
 if (file_exists('../public/atkjs-ui.min.js')) {
+    $test = __DIR__ . '/public';
     $app->cdn['atk'] = '../public';
 }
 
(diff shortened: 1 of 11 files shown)


demos/init.php
```diff
@@ -115,45 +115,50 @@ public function init(): void
 $app->title = 'Agile UI Demo v' . $app->version;
 
 if (file_exists('../public/atkjs-ui.min.js')) {
+    $test = __DIR__ . '/public';
```
mvorisek: Is this needed?

public/agileui.css
```diff
@@ -86,6 +90,16 @@ i.atk-panel-warning {
 i.atk-panel-warning.atk-visible {
   color: #F2711C;
 }
+.atk-admin-left-menu {
+  top: 47px !important;
```
mvorisek: Can the rule be rewritten without `!important` - it is otherwise hard to override this rule in custom css.

template/semantic-ui/layout/admin.html
```diff
@@ -1,13 +1,7 @@
 
 <div class="atk-layout">
```
mvorisek, line 2: Can an html ID be added here?

mvorisek (changes requested):

demos/init.php
ibelar: Oups forgot to remove it. I will do so.

public/agileui.css
ibelar: It is important because, without it, it will not work. 😄 Besides, it goes with the specific template.

template/semantic-ui/layout/admin.html
ibelar: No. This is how it is defined for this layout.

public/agileui.css
mvorisek: I meant more specific selector, or does any other rule already use `!important`?

template/semantic-ui/layout/admin.html
mvorisek: Strange, the top-most element can not have an html id?

mvorisek: Related with https://github.com/atk4/ui/issues/1028#issuecomment-606783135

ibelar: This pr is not about adding ID to the layout. Perhaps another time where we could add it to all layout as well.

public/agileui.css
DarkSide666: Maybe you can move this to `.ui.left.sidebar.atk-admin-left-menu` selector and without important?

DarkSide666 (approved):
LGTM except that `important` word there. But if there are no other way to set this, then it's fine.

ibelar:
> LGTM except that `important` word there. But if there are no other way to set this, then it's fine.

@DarkSide666  - Your suggestion worked. I did made a change. thks.

PR #1114, opened 2020-04-19
Fix closure for extended App

abbadon1334:
This way even extended ui\App get the correct call.

### DIFF
--- a/src/jsSSE.php
+++ b/src/jsSSE.php
@@ -96,10 +96,9 @@ class jsSSE extends jsCallback
         }
 
         // output headers and content
-        $app = $this->app;
-        \Closure::bind(static function () use ($app, $content): void {
-            $app->outputResponse($content, []);
-        }, null, App::class)();
+        \Closure::bind(static function () use ($content): void {
+            $this->outputResponse($content, []);
+        }, $this->app)();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/jsSSE.php
```diff
@@ -96,10 +96,9 @@ private function output(string $content): void
         }
 
         // output headers and content
-        $app = $this->app;
-        \Closure::bind(static function () use ($app, $content): void {
-            $app->outputResponse($content, []);
-        }, null, App::class)();
+        \Closure::bind(static function () use ($content): void {
```
mvorisek: This will not work, you can not rebind static closure.

But your intention with this PR is good. Rebinding `$this` seems not to be supported by any current IDE, so the best solution is probably: https://3v4l.org/FJ0YK , ie. just change the class name to the app. instance.

As far as I remember I used similar construction in App class but the static was there intentional for another reason - to not bind to this to allow GC to collect the class.

Also recheck all `Closure::bind` usages in all repos.

mvorisek (changes requested):

src/jsSSE.php
abbadon1334: > Also recheck all `Closure::bind` usages in all repos.
core, data, dsql, ui have only that occurence :
image

mvorisek (approved):
Now perfect!

PR #1118, opened 2020-04-21
enable control over action modal by option to pass defaults array

georgehristov:


src/TableColumn/ActionButtons.php
```diff
@@ -100,7 +100,7 @@ public function addButton($button, $action = null, $confirm = false, $isDisabled
      * load contents through $callback. Will pass a virtual page.
      *
      * @param \atk4\ui\View|string $button
-     * @param string               $title
+     * @param string|array         $title - model title or model seed array
```
mvorisek: There should be no `-`.

"model", what class is seeded by default?

src/TableColumn/ActionButtons.php
```diff
@@ -111,7 +111,7 @@ public function addModal($button, $title, $callback, $owner = null, $args = [])
     {
         $owner = $owner ?: $this->owner->owner;
 
-        $modal = \atk4\ui\Modal::addTo($owner, [compact('title'), 'appStickyCb' => true]);
+        $modal = \atk4\ui\Modal::addTo($owner, [is_array($title) ? $title : compact('title'), 'appStickyCb' => true]);
```
mvorisek: Replace all `compact('xx', ...)` usages to `['xx' => $xx, ...]`, compact is not refactorable.

mvorisek (changes requested):

src/TableColumn/ActionButtons.php
georgehristov: Corrected

src/TableColumn/ActionButtons.php
georgehristov: Corrected

mvorisek (approved):

src/TableColumn/ActionButtons.php
```diff
@@ -100,18 +100,20 @@ public function addButton($button, $action = null, $confirm = false, $isDisabled
...
      *
      * @return \atk4\ui\View
      */
-    public function addModal($button, $title, $callback, $owner = null, $args = [])
+    public function addModal($button, $defaults, $callback, $owner = null, $args = [])
     {
         $owner = $owner ?: $this->owner->owner;
 
-        $modal = \atk4\ui\Modal::addTo($owner, [compact('title'), 'appStickyCb' => true]);
+        $defaults = [is_array($defaults) ? $defaults : ['title' => $defaults], 'appStickyCb' => true];
```
mvorisek: Array merge is probably needed, so this should look like:
```
if (is_string($defaults)) {
    $defaults = ['title' => $defaults];
}

$defaults['appStickyCb'] = true;
```

mvorisek (changes requested):

src/TableColumn/ActionButtons.php
georgehristov: It works like it is. I did not dig deeper enough to see why it works :)
I also prefer array_merge

mvorisek (approved):
Perfect :)

PR #1119, opened 2020-04-21
Fix getRenderedModals

abbadon1334:
If App layout is not defined and use terminateJSON to output a response, in method getRenderedModals foreach over $this->html->elements not exists and raise error

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -1143,7 +1143,7 @@ class App
     public function getRenderedModals()
     {
         $modals = [];
-        foreach ($this->html->elements as $view) {
+        foreach ($this->html->elements ?? [] as $view) {
             if ($view instanceof Modal) {
                 $modals[$view->name]['html'] = $view->getHTML();
                 $modals[$view->name]['js'] = $view->getJsRenderActions();
--- a/src/App.php
+++ b/src/App.php
@@ -1143,7 +1143,7 @@ class App
     public function getRenderedModals()
     {
         $modals = [];
-        foreach ($this->html->elements ?? [] as $view) {
+        foreach ($this->html !== null ? $this->html->elements : [] as $view) {
             if ($view instanceof Modal) {
                 $modals[$view->name]['html'] = $view->getHTML();
                 $modals[$view->name]['js'] = $view->getJsRenderActions();


src/App.php
```diff
@@ -1143,7 +1143,7 @@ private function outputResponseJSON($data, array $headers = []): void
     public function getRenderedModals()
     {
         $modals = [];
-        foreach ($this->html->elements as $view) {
+        foreach ($this->html->elements ?? [] as $view) {
```
mvorisek: This is not good if `$this->html` may be null. If so use something like `$this->html !== null ? $this->html->elements : []`.

mvorisek (changes requested):

src/App.php
abbadon1334: while traversing it get [] on every check isset

mvorisek: Yes, but elements should not be empty, if so, it should be not supressed. Imagine if there will be getters, ?? will fail.

abbadon1334: Ok, is not the right way to use it but if i use App this way :
```php
$app = new App();
$app->terminateJSON(['code' => 201, 'message' => 'record created']);
```

i don't need layout and i don't init, just emit a response like a JSON API.
Just now i see that probably this approach can be better
```php
$app = new App();
$app->outputResponseJSON(['code' => 201, 'message' => 'record created']);
$app->callExit();
```
What do you think?

mvorisek: I am neutral on that, it has + and -, but terminateXX is handy to enforce it is not called multiple times (which makes sense for JSON - and for custom output streamer, you can still use outputXX()...). What are you solving exactly?

abbadon1334:
First i try to add PSR7 + PSR15 no way due to SSE.
i try to have only PSR7, and after i finish i try to add PSR7 at least for request in UI, to remove all horrible calls to superglobals.

What i try t accomplish : 
- I inject App with loaded files via ConfigTrait
- I add routing in a much simpler way than my previous fastroute addon, i use annoteted route directly on methods, like many other frameworks
- I manage Pre and Post RouteHandler method like middlewares setup directly from annotation

Now i try to do the same but for API and i think of _re-use_ some methods of ui\App for the output response.

Both above approach gives error.
Your logic of that output methods is correct, they must be not exposed.

I will find a way, let's close this for now.

mvorisek:
> i use annoteted route directly on methods, like many other frameworks

👍 

Symfony has mastered a lot of that (routing, relative path discovery, ...) and it should be probably reused. SSE is special, but Symfony has probably some solution for streamed output - in Ui/App, we should probably process each single SSE response like a full single one (i.e. to support terminateXX() or equivalent method).

In general, Atk should solve the real, value added issues and reuse industry standard tools for everything else.

mvorisek:
@abbadon1334 Is there currently any use case where this fix is needed?

If yes, we can merge it, otherwise the patch branch should be deleted.

More on why `isset` is needed: https://bugs.php.net/bug.php?id=75831

DarkSide666:
Oh I stepped on this `isset` bug (https://bugs.php.net/bug.php?id=75831) multiple times already and wasn't exactly sure where bug is - in our code or in php. Thanks - now I know it's PHP :) but we should be aware of that and work around that.

mvorisek:
@DarkSide666 green light for this?

DarkSide666 (approved):
Yes I think it's fine.

mvorisek (approved):

PR #1120, opened 2020-04-21
[feature] Maestro Layout

ibelar:
A new layout based on the admin layout with improve left menu.

 - Indicate which group and page you are in;
 - Collapsable group;
 - Clicking on a group title will bring you to the first menu item in the group;
- Proper group open on page load

## Note
 Require atkjs-ui.min.js v1.14.0
Change default admin for demos to use Maestro.

Screen Shot 2020-04-21 at 3 52 05 PM
Screen Shot 2020-04-21 at 3 52 26 PM


js/src/plugins/admin-menu.plugin.js
```diff
@@ -16,9 +16,10 @@ export default class ajaxec extends atkPlugin {
   main() {
     // menu items container.
     this.menu = this.$el.find(this.settings.menuItemsSelector);
+    console.log($.atkGetUrl(this.settings.base));
```
DarkSide666: is it ok to leave this debug info there?

js/src/plugins/sidenav.plugin.js
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Add click handler for menu group
+   * and toggler element.
+   */
+  addClickHandler() {
+    const that = this;
+    this.$el.on('click', function(e) {
+      // simulate click event on first menu item in group.
+      that.menu.find('a').first()[0].click();
```
DarkSide666: What will happen if there are no menu items in menu group and you click on group?

Also shouldn't we better use `this.settings.menuItemsSelector` css class selector here just in case in future something can change in html structure of menu? On the other hand, there should always be `<a>` element somewhere so... maybe css selector is not needed.

DarkSide666 (approved):
Looks good

js/src/plugins/sidenav.plugin.js
ibelar: Make sense. I will customize it, using a different option with default to 'a'. This way, it will be possible to customize it.

PR #1121, opened 2020-04-21
Feature/implement html decorator

### LINKED ISSUE #875: Add Table Column Decorator for HTML code
Hi,

I think its nice to have some Decorator that just outputs HTML, e.g. if you have some pre-rendered HTML in a model field that you want to output as-is.
If you think thats good to have, here is a working implementation:

```
<?php

namespace atk4\ui\TableColumn;

/**
 * Class HTML
 * @package atk4\ui\TableColumn
 * 
 * Use this decorator if you have HTML code that you just want to put into the table cell
 */
class HTML extends Generic
{

    /**
     * 
     */
    public function getDataCellHTML(\atk4\data\Field $f = null, $extra_tags = [])
    {
        return '{$_'.$f->short_name.'}';
    }


    /**
     *
     */
    public function getHTMLTags($row, $field)
    {
        return ['_'.$field->short_name => '<td>'.$row->get($field).'</td>'];
    }
}
```

### LINKED ISSUE #1117: Html tags in field
How can I render html tags in CRUD column saved in a table field? I tried `Template `class:

`$crud->addDecorator("html_field", ['Template', '<b>{$html_field}</b>']`

but it renders only `<b>` outside the field and not its content

DarkSide666:
Implements `TableColumn/HTML` decorator.

Thanks goes to @PhilippGrashoff (fix #875).
Also requested here (fix #1117)

### DIFF
--- a/demos/table.php
+++ b/demos/table.php
@@ -43,10 +43,10 @@ $table->addTotals(['name' => 'Totals:', 'salary' => ['sum']]);
 
 // ========================================================
 $my_array = [
-    ['name' => 'Vinny', 'surname' => 'Sihra', 'birthdate' => '1973-02-03'],
-    ['name' => 'Zoe', 'surname' => 'Shatwell', 'birthdate' => '1958-08-21'],
-    ['name' => 'Darcy', 'surname' => 'Wild', 'birthdate' => '1968-11-01'],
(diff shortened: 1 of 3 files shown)


ibelar (approved):
LGTM

PR #1122, opened 2020-04-22
Reorganize demos file

ibelar:
Reorganize demo files structure.

- new folder structure
- cleanup some demo
-  rename file init.php to atk-init.php
- modify db model for preventing edit, save and delete for Country


demos/collection/grid.php
```diff
@@ -1,13 +1,24 @@
 <?php
 
-require_once __DIR__ . '/init.php';
-require_once __DIR__ . '/database.php';
+chdir('..');
```
mvorisek: Is `chdir` here only for the include below? If yes, use `require_once __DIR__ . '/..' ...`

tests/DemoCallExitTest.php
```diff
@@ -305,4 +318,16 @@ public function JSONResponsePOSTDataProvider()
...
+    /**
+     * Set directory with proper separator.
+     *
+     * @param $dir
+     *
+     * @return string
+     */
+    private function getPath(string $dir): string
+    {
+        return DIRECTORY_SEPARATOR . $dir . DIRECTORY_SEPARATOR;
```
mvorisek: Is this URL path or local path? If URL, `DIRECTORY_SEPARATOR` is wrong (there should be always `/`), if local/fs, then prefix with `__DIR__`.

mvorisek (changes requested):

demos/database.php
```diff
@@ -14,7 +14,32 @@
     ]))->addMoreInfo('PDO error', $e->getMessage());
 }
 
-$app->db = $db;
+
+if (!trait_exists('ModelLockTrait')) {
```
mvorisek: Why if trait exists trait?

demos/collection/grid.php
ibelar: if chdir is not use than atk-init.php is not able to include database.php and somedatadef.php
not sure what is the best way to handle this.

I can also still use chdir and do:
````
chdir('..');
require_once __DIR__ . '/../atk-init.php';
````
what do you think is best?

File: tests/DemoCallExitTest.php (hunk 2 above)
Comment (ibelar) [reply tests/DemoCallExitTest.php]: This is for getResponseFromRequestGET($path), so I guess it is about url. I  will change it for '/'

File: demos/database.php (hunk 3 above)
Comment (ibelar) [reply demos/database.php]: Not sure what you mean here...

Comment (mvorisek) [reply demos/database.php]: Remove `if (!trait_exists('ModelLockTrait')) {`

File: demos/collection/grid.php (hunk 1 above)
Comment (mvorisek) [reply demos/collection/grid.php]: Never use `chdir`, we are not in bash ;-), and use `__DIR__` everywhere - also in `atk-init.php` which should work then.

File: demos/database.php (hunk 3 above)
Comment (ibelar) [reply demos/database.php]: done

File: tests/DemoCallExitTest.php (hunk 2 above)
Comment (ibelar) [reply tests/DemoCallExitTest.php]: fix

File: demos/collection/grid.php (hunk 1 above)
Comment (ibelar) [reply demos/collection/grid.php]: fix

File: demos/database.php [hunk 4]
```diff
@@ -2,8 +2,8 @@
 
 // A very basic file that sets up Agile Data to be used in some demonstrations
 try {
-    if (file_exists('db.php')) {
-        include_once __DIR__ . '/db.php';
+    if (file_exists(dirname(__DIR__) . '/demos/db.php')) {
```
mvorisek: @ibelar We are already in demos/, no dirname needed :)

ibelar: I thought it was safer this way. I will change it.

demos/collection/jsactions2.php
```diff
@@ -1,7 +1,12 @@
 <?php
 
-require_once __DIR__ . '/../atk-init.php';// Check this file in order to see how model action is defined.
-require_once '_includes/country_actions.php';
+require_once __DIR__ . '/../atk-init.php';
+require_once __DIR__ . '/../_includes/country_actions.php';
+
+/**
+ * Demo for Model action
+ * Action definition for Country model is located in counrty_actions.php
```
mvorisek, line 8: `counrty_actions` typo

demos/database.php
ibelar: done!

demos/collection/jsactions2.php
ibelar: Done!

demos/atk-init.php
```diff
@@ -36,7 +36,7 @@
 $app->title = 'Agile UI Demo v' . $app->version;
 
 
-if (file_exists(dirname(__DIR__) . '/public/atkjs-ui.min.js')) {
+if (file_exists(__DIR__ . '/../public/atkjs-ui.min.js')) {
```
mvorisek: Dirname here was ok, but it does not matter here and this is fine. I try to use `dirname` when the output is returned as the path looks nicer/more normalized.

ibelar: Ok - good to know.  I am always mixt up when to use what. 😄 Should we keep it this way? I do not mind to change back.

mvorisek: I think with `. /../` it is more readable, so the current form is preffered by me

ibelar: revert

mvorisek (approved):

PR #1123, opened 2020-04-22
Documentation updates

TheNetworkIsDown:
Backslashes in Class references enclosed with :php:meth: need to be escaped (double backslashes)


georgehristov (approved):
LGTM, thanks!

PR #1124, opened 2020-04-22
Simplify locale path getter

mvorisek:


src/Locale.php
```diff
@@ -11,11 +11,9 @@ public function __construct()
...
     /**
      * Get absolute Path of default translations.
-     *
-     * @return string
      */
-    public static function getPath()
+    public static function getPath(): string
     {
-        return implode(DIRECTORY_SEPARATOR, [__DIR__, '..', 'locale']) . DIRECTORY_SEPARATOR;
+        return dirname(__DIR__) . '/locale';
```
georgehristov: You need to keep `DIRECTORY_SEPARATOR` instead of `/` so I suggest
```
return dirname(__DIR__) . DIRECTORY_SEPARATOR . 'locale';
```

mvorisek: for which OS?

mvorisek:
Fixed variant of https://github.com/atk4/data/pull/568 to make the tests pass with non-existing locale directory

georgehristov (approved):
LGTM

Issue #40, opened 2017-02-22
App does not auto-run if extended.


romaninsh:
See code:

![screen-shot-2017-02-22-at-9 53 53-pm](https://cloud.githubusercontent.com/assets/453929/23229947/d45ab15c-f94a-11e6-93b1-3e0f0ab2de35.png)

romaninsh:
This works fine for me in PHP 7.1 (not the exact code above, but by cloning repo from git@github.com:gowrav-vishwakarma/atkui-demo.git)

PR #1127, opened 2020-04-23
Introduce FormField\ScopeBuilder

georgehristov:
Introduces GUI for setting up Model scopes using [VueQueryBuilder](https://dabernathy89.github.io/vue-query-builder/)

Below is based on demos `Stat` model

scope


romaninsh:
I'd like to hear what others think - should this be in this repo or as add-on?

mvorisek:
> 
> 
> I'd like to hear what others think - should this be in this repo or as add-on?

It is quite specifc, I think this repo should provide only basic and well maintained UI components and complex components only if they are used a lot and enought flexible. So I am for an extra repo.

If this component will be used a lot, it can be always moved here.

What about a 2nd repo like `ui-extra` as a central place for early / unmature components?

georgehristov:
In fact I would also agree this is not part of the base UI components and it would be better off in a separate repo.
Then we just need to collect opinions if stand-alone or as a bundle (`ui-extra` or `ui-extended` or ...)

mvorisek:
I prefer bundle/addon but with demos like in this repo. CI scripts can be copied 1:1 (incl. phpunit and Behat testing). But it should be extra, not extended. (but feel free still to propose a better name)

romaninsh:
I suppose the difference is - if this component in here, then any PRs into core repositories (atk data for example), will need to make condition builder is updated. If it's in a separate repository then @georgehristov would have to maintain it himself. 

I can see where scope builder could be used inside column filters, so that's an argument to include builder here. This would also allow other add-ons to rely on it easily.

mvorisek:
> I can see where scope builder could be used inside column filters, so that's an argument to include builder here.

makes sense to include here for another argument - it is only one file

ibelar:
If we include it in UI than I would propose adding all js related files into atkjs bundle. This way, we would not have to worry about including them into pages and it should work in Modal and VirtualPage out of the box. 

@georgehristov - I am working into creating a Fomantic-UI look and feel for this component and already have extended it using the Fomantic CSS template. I can take care of that part also if you want. Plus we need a calendar component for date field in it, so it could be included here as well. The calendar component could be used in MultiLine too.

georgehristov:
@ibelar Integrating this with FUI sounds great and really appreciated!

georgehristov:
@ibelar I have included it in atkjs in `atkComponents` as `atk-query-builder` but have not pushed this commit.
Shall I push it for reference or you are further down the road than this already?

js/src/services/vue.service.js
```diff
@@ -15,6 +16,7 @@ let atkComponents = {
   'atk-item-search' : itemSearch,
   'atk-multiline'   : multiLine,
   'atk-tree-item-selector' : treeItemSelector,
+  'atk-query-builder' : VueQueryBuilder,
```
mvorisek, line 19: @ibelar why is this needed? Do we need to change js for every Vue based component added?

src/FormField/ScopeBuilder.php
```diff
@@ -0,0 +1,576 @@
...
+     */
+    protected $query = [];
+
+    /**
+     * VueQueryBulder => Condition map of operators
+     *
+     * @var array
+     */
+    protected static $operators = [
+        'equals' => '=',
```
mvorisek: @georgehristov you use these natural strings on multiple locations, can you declare them in constants once and then use them?

Example:
```
public const OPERATOR_DOES_NOT_EQUAL = 'does not equal';

something(self::OPERATOR_DOES_NOT_EQUAL);
```

georgehristov: @mvorisek I also like using constants for definition but the challenge here is that we have operator label and operator value. Will think about how to implement it.

mvorisek: image

something that is unique... it needs to be probably the name/label :)

src/FormField/ScopeBuilder.php
```diff
@@ -524,16 +543,16 @@ public static function conditionToQuery(Condition $condition): array
 
             $map = [
                 'LIKE' => [
-                    'equals',
-                    'begins with',
-                    'ends with',
-                    'contains'
+                    self::OPERATOR_EQUALS,
+                    self::OPERATOR_BEGINS,
+                    self::OPERATOR_ENDS,
```
mvorisek: `ENDS_WITH` sound much more natural

src/FormField/ScopeBuilder.php
```diff
@@ -524,16 +543,16 @@ public static function conditionToQuery(Condition $condition): array
...
                 ],
                 'NOT LIKE' => [
-                    'does not equal',
-                    'does not begin with',
-                    'does not end with',
-                    'does not contain'
+                    self::OPERATOR_DOESNOT_EQUAL,
+                    self::OPERATOR_DOESNOT_BEGIN,
+                    self::OPERATOR_DOESNOT_END,
+                    self::OPERATOR_DOESNOT_CONTAIN,
```
mvorisek: `DOESNOT` sounds not ideal - is `NOT_CONTAINS` better?

src/FormField/ScopeBuilder.php
```diff
@@ -549,7 +568,7 @@ public static function conditionToQuery(Condition $condition): array
                 $value = implode(',', $value);
                 $operator = $map[$operator] ?? 'IN';
```
mvorisek, line 569: missed this? 3 maybe there are even more placed...

src/FormField/ScopeBuilder.php
```diff
@@ -72,30 +72,49 @@ class ScopeBuilder extends Generic
...
-        'ends with' => 'LIKE',
-        'does not end with' => 'NOT LIKE',
-        'is in' => 'IN',
-        'is not in' => 'NOT IN',
-        'matches regular expression' => 'REGEXP',
-        'does not match regular expression' => 'NOT REGEXP',
-        'is empty' => '=',
-        'is not empty' => '!=',
+        self::OPERATOR_EQUALS => '=',
+        self::OPERATOR_DOESNOT_EQUAL => '!=',
```
mvorisek: Does `negative` operators matches nulls or is there any option to set it like so - we should have real inverse operator.

src/FormField/ScopeBuilder.php
georgehristov: I tried it but I did not like it

src/FormField/ScopeBuilder.php
georgehristov: Done now, I am a proponent of good semantics in code

mvorisek: Great!

src/FormField/ScopeBuilder.php
georgehristov: Not really as this is the `atk4\data\Scope` version of operators.
Constants in the class represent the VueQueryBuilder operators

src/FormField/ScopeBuilder.php
georgehristov: Can you give an example as I do not fully understand what you mean?

src/FormField/ScopeBuilder.php
mvorisek: You marked it `protected` that I think it is ok... I do not like boths but have no better now :D

src/FormField/ScopeBuilder.php
```diff
@@ -0,0 +1,595 @@
...
+            'limit' => 250
+        ]
+    ];
+    /**
+     * Max depth of nested conditions allowed.
+     * Corresponds to VueQueryBulder maxDepth
+     *
+     * @var integer
+     */
+    public $maxDepth = 10;
```
mvorisek: Have this any performance side effect on ui side if large? (test with 10000)

If not, let's set at least to 25.

src/FormField/ScopeBuilder.php
```diff
@@ -0,0 +1,595 @@
...
+     * @var Template
+     */
+    public $scopeBuilderTemplate;
+
+    /**
+     * List of delimiters for auto-detection in order of priority
+     *
+     * @var array
+     */
+    public static $listDelimiters = [';', ',', '|'];
```
mvorisek: When `|` makes sense? I think we do not need it at all.

src/FormField/ScopeBuilder.php
georgehristov: No performance issues but this will be a hell of a Scope 😄 
I have never used more than 3 - 4 sub-levels.
Not sure if you got what this setting means.
You can always set a value when creating the ScopeBuilder object in your code

src/FormField/ScopeBuilder.php
mvorisek: table cell with value `null` is not matched with `= 1` but also NOT matched with `!= 1` (at least in MySQL/Maria)

src/FormField/ScopeBuilder.php
georgehristov: Aiming at more versatility. Some users are used to using | for a list delimiter (from DynamicsNAV as one example)

src/FormField/ScopeBuilder.php
```diff
@@ -0,0 +1,565 @@
...
+     */
+    public static function queryToScope(array $query): AbstractScope
+    {
+        $type = $query['type'] ?? 'query-builder-group';
+        $query = $query['query'] ?? $query;
+
+        switch ($type) {
+            case 'query-builder-group':
+                $components = array_map([static::class, 'queryToScope'], $query['children']);
+                $junction = $query['logicalOperator'] === 'all' ? Scope::AND : Scope::OR;
```
mvorisek: I do not like `all` name

src/FormField/ScopeBuilder.php
mvorisek: But in rare cases you may need 11, not that far from 3-4 :D

src/FormField/ScopeBuilder.php
mvorisek: I think better to not make default because it is confusing with `||`.

src/FormField/ScopeBuilder.php
mvorisek: image

In GUI we should have definitely `OR` and/or `AND` - people are used to use these even from non-SQL world

mvorisek: should be also updated in data/scope: `scope->all` is simply not ok

demos/input/query-builder.php
```diff
@@ -0,0 +1,24 @@
...
+        'choices' => [
+            ['label' => "Apple", 'value' => 'apple'],
+            ['label' => 'Banana', 'value' => 'banana']
+        ]
+    ]
+];
+
+$f = \atk4\ui\Form::addTo($app);
+$qb = $f->addField('qb', [\atk4\ui\FormField\ScopeBuilder::class]);
+$qb->rules = $rules;
```
mvorisek: @ibelar can you setup your IDE to always add end EOL for text files on save? Seen this many times and for pug/html it is currently not checked by our CI tests.

demos/input/query-builder.php
```diff
@@ -22,3 +22,7 @@
 $f = \atk4\ui\Form::addTo($app);
 $qb = $f->addField('qb', [\atk4\ui\FormField\ScopeBuilder::class]);
 $qb->rules = $rules;
+
+$f->onSubmit(function($f) {
+   echo  '<pre>' . json_encode($f->model->get('qb'), JSON_PRETTY_PRINT) . '</pre>';
```
mvorisek: Use `App->jsonEncode` (if not only for some test...)

ibelar:
@georgehristov - Does scope in data support operator 'BETWEEN'? If so, I can add an extra input to the query-builder when the user selects 'BETWEEN' as the operator. As of now, the name of the parameter for the first input is 'value'. What should be the extra one? (value_2, extra)

georgehristov:
@ibelar BETWEEN is not supported in Model\Scope.
Implementing it is too complicated and functionality is there by using 2 simple conditions joined by AND

ibelar:
@georgehristov - Ok. I will probably still add it to UI then and generate proper AND query when user select ‘BETWEEN’. It make sense on a user stand point to use it.

georgehristov:
@ibelar that is what I was going to say. If we want to introduce to UI then we should map it to 2 conditions.
Challenge here is to map it 2 way - UI to scope and scope to UI

src/FormField/ScopeBuilder.php
```diff
@@ -28,10 +29,11 @@ class ScopeBuilder extends Generic
     /**
      * Max depth of nested conditions allowed.
      * Corresponds to VueQueryBulder maxDepth.
+     * Maximum support by js component is 5.
```
georgehristov: @ibelar This limit would be too low.
Any reason to set such hardcoded limit?

demos/Dockerfile
```diff
@@ -13,10 +13,16 @@ RUN apt-get update && apt-get install -y \
...
+#Update node to v12
+RUN curl -sL https://deb.nodesource.com/setup_12.x  | bash -
+RUN apt-get -y install nodejs
+
 WORKDIR /var/www/html/
 
 RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
 
+# update npm - apt-get is not installing the latest version.
+RUN npm install -g npm
```
mvorisek: should be put probably after install nodejs

.github/workflows/unit-tests.yml
```diff
@@ -58,7 +58,7 @@ jobs:
       - name: Install JS dependencies (only for Behat)
         if: matrix.type == 'Behat'
         run: |
-          apk add npm
+          apk add --update npm
```
mvorisek: simple, nice :)

src/FormField/ScopeBuilder.php
georgehristov: Changed

.github/workflows/unit-tests.yml
ibelar: tks to Google! :)

demos/Dockerfile
ibelar: As long as it is before doing any installation. Besides, I have tested it and it was ok.

src/FormField/ScopeBuilder.php
ibelar: The marketing department suggests starting with a lower limit. We can promote it as a new feature in the next release. :D 
Seriously no specific reason. What would be a good limit?

georgehristov: :D

I have used 5-6 levels so that's why I set it to 10.
@mvorisek proposed 25 but this would be a hell of a scope

ibelar: Let's start with 10. We can always readjust if we need more.

ibelar:
@georgehristov 

This is the final look. It now supports Date, Radio, Checkbox plus regular input of course. You can see how to rule is set up for them in this temp file: /demos/input/query-builder.php

P.S. The date picker relies on v-date-picker so we can use some of its options there.

Are you ok to hook it up to data/scope? I did not add between yet. I will wait to see how it looks like once set via a model.

Screen Shot 2020-06-17 at 1 59 17 PM

js/src/components/query-builder/query-builder.component.vue
```diff
@@ -33,8 +30,9 @@
         query: this.data.query ? this.data.query : {},
         rules: this.data.rules ? this.data.rules : [],
         name: this.data.name ? this.data.name : '',
-        maxDepth: this.data.maxDepth ? ((this.data.maxDepth <= 5) ? this.data.maxDepth : 5 ): 1,
+        maxDepth: this.data.maxDepth ? ((this.data.maxDepth <= 10) ? this.data.maxDepth : 10 ): 1,
```
mvorisek, line 33: why limit on two places?

georgehristov:
@ibelar Wonderful! I will take it up for hooking to data/scope.

js/src/components/query-builder/query-builder.component.vue
ibelar: Just a safety on the component side.

mvorisek: So with 100 the frontend will catch on fire? :D

mvorisek:
image

@ibelar look very good

What about adding a little (1/3 - 1/2) space between the rules?

Also, the "X"es are very dominant and distracting. Can we display them only when a mouse is over that rule or show it with light grey color only when out of focus? (both can be can easily in CSS)

ibelar:
> What about adding a little (1/3 - 1/2) space between the rules?

I can propose one version with some space in it. Default spacing was definitely too large, but may be some space will be better.

> Also, the "X"es are very dominant and distracting. Can we display them only when a mouse is over that rule or show it with light grey color only when out of focus? (both can be can easily in CSS)

You are right, I will make it smaller. I do no like hover because it is not that intuitive. Plus, hover on a phone or tablet is not that great.

georgehristov:
@ibelar Now I have "linked" it with Model scope and updated the demo.

You need to require `"atk4/data" : "dev-feature/model-scope"` in composer.json

Some issues I noticed:
- multiple `Duplicate keys` error in console
- "Add Rule" dropdown can get quite long - maybe make scrollable?
- radio selects for "Is Commercial" field select both "Yes" and "No" on click
- I know it is default functionality in VueQueryBuilder but it is annoying that when user selects "is empty" option the input for text is still shown. Can you do something about it?
- "Currency" field which is enum type does not show a dropdown select

georgehristov:
> What about adding a little (1/3 - 1/2) space between the rules?

I find no spaces really good as it visually separates sub-scopes. In any case space should be very small for that purpose

js/src/components/query-builder/query-builder.component.vue
```diff
@@ -41,20 +41,27 @@
...
-          "removeRule": "&times;",
-          "addGroup": "Add Group",
-          "removeGroup": "&times;",
-          "textInputPlaceholder": "value",
+          addRule: "Add Rule",
+          removeRule: "small icon times",
+          addGroup: "Add Group",
+          removeGroup: "small icon times",
+          textInputPlaceholder: "value",
+          spaceRule: "fitted", // can be compact or padded
```
mvorisek: so can it be fitted?

ibelar:
Component is update 

I have reduced the size of the remove icon and change it's color so it is a lot less disturbing. I also add the possibility to add space in between rule via the component labels options. The rule is called spaceRule and you have a choice of 'fitted', 'compact', or 'padded'. Default is fitted, i.e. no space and 'padded' is the maximum space.

Here is a look with 'compact' option using new remove icon.

Screen Shot 2020-06-17 at 5 50 36 PM

mvorisek:
@ibelar ❤️ it, X should be "same light gray" as button text "Add Rule"

one more question - is that really that huge in reality? Based on your screen, there only like 8 rules fits 1200p screen/non-fullscreen browser!

it should no larger/higher than this: (click for real size)
image

js/src/components/query-builder/query-builder.component.vue
mvorisek: I think this safety check should be removed (it checks if data from php are ok?) and fallback to 10 if not specified. Simple.

ibelar:
> one more question - is that really that huge in reality? Based on your screen, there only like 8 rules fits 1200p screen/non-fullscreen browser!

It uses the Fomantic card component with fluid class. It will be as large as the container you are placing it in too. You can easily change that via `scopeBuilderTemplate` property.

ibelar:
> Some issues I noticed:
> 
> * multiple `Duplicate keys` error in console
> * "Add Rule" dropdown can get quite long - maybe make scrollable?
> * radio selects for "Is Commercial" field select both "Yes" and "No" on click
> * I know it is default functionality in VueQueryBuilder but it is annoying that when user selects "is empty" option the input for text is still shown. Can you do something about it?
> * "Currency" field which is enum type does not show a dropdown select

Yes - I will look into it.

mvorisek:
@ibelar I mean the current style is unusable - it is brutally big, like only 8-10 rules fit to the whole screen. We need like 40 (or in another words 1/3 of the height for header/footer/page title, 1/3 for some more content like manual, data, etc., and then within 1/4 - 1/3 of screen height the typical scope rules must fit.

**In other words, we have about 250px of height and 10 rules must fit in this space at least.**

Width is ok.

js/src/components/query-builder/query-builder.component.vue
ibelar: yes - good point!

ibelar:
> @ibelar I mean the current style is unusable - it is brutally big, like only 8-10 rules fit to the whole screen. We need like 40 (or in another words 1/3 of the height for header/footer/page title, 1/3 for some more content like manual, data, etc., and then within 1/4 - 1/3 of screen height the typical scope rules must fit.

Height is ok as it is. Reducing the height of each rule would only make it harder to read. That is why I remove space in-between rules so we can fit more of them. Can you show me a screen sample of what you are saying?

mvorisek:
@ibelar click your and my screenshots above - my is 1:1, your probably too and compare. Your is simply too big. This is simply not acceptable for any serious/nonsimple/backoffice rules.

ibelar:
> @ibelar click your and my screenshots above - my is 1:1, your probably too and compare. Your is simply too big. This is simply not acceptable for any serious/nonsimple/backoffice rules.

@mvorisek - My mistake, I meant a screen shot of what you would really want. Did you see something similar you would like to share. I can see if I can adjust.

mvorisek:
@ibelar see my comments - if you posted 1:1 (ie. with 100% zoom), then the height of every element needs to be reduces to fix about 30 - 40 rules to standard 1080 px height monitor.

ibelar:
@ mvorisek - The actual look is very acceptable and no more change will be done on it. Only debugging ui glitch will be done at this stage.

mvorisek:
@ibelar the example from you is 825 px height with only 5 rules (4 rules + 1 nesting). @DarkSide666 @georgehristov do you really think this is usable?

Let's take an alternative fact. Excel. The world 1st data editor. When you open it with default settings/zoom, 35 rows fits within the same height.

So your score on this is only 14.3% (5/35) 😢

georgehristov:
@mvorisek Have you tested it?

It is quite fine from my point of view. Sizes appear very UI friendly - not too big, not too small.

Below is a screenshot on normal screen Agile UI demo page

scope-builder

mvorisek:
No. **I kindly asked if provided screen is 1:1**. Now/in your screen there are 17 rows. I still think the grey box paddings can be reduced, but it seems much better. (I propose 60 px, ie. not higher than the save button below)

And the value input field should be probably streched (ie. occupy all free width).

mvorisek:
Also, I do think display "==" is much better than displaying "equals" - much better UX? You do not trust me? Check some leading softwares? :D :D

So I propose to change the operator text to "==" (and >=, !=, ...) or "== (equals)"-

And match selectbox from "And" to "And (all rules must be satisfied)" (and "Or (At least one rule much be satisfied)") (this must be tested, I am not exactly sure if it is really that better)

js/src/components/query-builder/query-builder.component.vue
ibelar: because the query can come from anywhere. If stored in DB for example, we may not have control over it. I prefer to limit it here and 10 levels should be quite enough.

georgehristov:
@ibelar I introduced some updates in PHP side
- debug is now an option that can be used any time (switched on for demo)
- inputType specific operators (`>`,`<`,`=`, etc for numeric, `before`, `after`, `on`, etc for dates)

I hope it does not interfere on your side

So @mvorisek we are now in line with "some leading softwares", breathing down their necks 😄

mvorisek:
> inputType specific operators (>,<,=, etc for numeric, before, after, on, etc for dates)

why, how is before different to "<"?

> So @mvorisek we are now in line with "some leading softwares" 😄

@georgehristov your statement or measured result? If result, please provide data! I did :)

ibelar:
@georgehristov - All should be working for the js component. Let me know if you see anything else.

georgehristov:
@ibelar Works great!

I just cannot figure out why tests failing (after I added query builder menu link for the demo)

Tests are running unbelievably slow on Windows

demos/Dockerfile
```diff
@@ -13,6 +13,10 @@ RUN apt-get update && apt-get install -y \
     && docker-php-ext-install intl \
     && docker-php-ext-install pdo pdo_mysql
 
+RUN curl -sL https://deb.nodesource.com/setup_12.x  | bash - \
+    && apt-get update && apt-get -y install nodejs \
+    && npm install -g npm
```
mvorisek: @ibelar `npm` is installed above (line `libicu-dev git jq unzip npm`) - so move `npm` here after the `apt` sources was changed and remove `npm install -g npm`?

ibelar: this is not for installing it but for updating it. It needs to be installed first. But not sure why it installed the wrong version. Probably because the image contains node v8

georgehristov:
Test fails as https://github.com/atk4/data/pull/551 is not merged yet.

@mvorisek Tests run much faster now!

demos/Dockerfile
mvorisek: The dist of the used image is conservative... After the sources are updated, I would expect we can install latest straight as I have advised above. I do not see any advantage why to install not-wanted version first.

mvorisek:
> @mvorisek Tests run much faster now!

yes, and they now also discover all demos and test them!

georgehristov:
@ibelar I have now updated this to reflect latest in atk4/data as scopes are merged now. The scope builder form control in demo however does not seem to submit data to backend. Can you pls check this?

src/FormField/ScopeBuilder.php
georgehristov: Now operators `empty`, `not empty` result in checking against `NULL` value

georgehristov:
This [Vue Query Builder](https://www.syncfusion.com/vue-ui-components/vue-query-builder) I find better but unfortunately it is paid 😄 
UI is closer to the originally used [jQuery QueryBuilder](https://querybuilder.js.org/)

src/Form/Control/ScopeBuilder.php
```diff
@@ -538,7 +544,7 @@ public static function conditionToQuery(Scope\Condition $condition, $inputsMap =
         $operator = $condition->operator;
         $value = $condition->value;
 
-        if (stripos($operator, 'like') !== false) {
+        if (str_contains($operator, 'like')) {
```
mvorisek: no, we should test for both `like`

src/Form/Control/ScopeBuilder.php
```diff
@@ -0,0 +1,611 @@
...
+        'debug' => false, // displays query output live on the page if set to true
+    ];
+    /**
+     * Max depth of nested conditions allowed.
+     * Corresponds to VueQueryBulder maxDepth.
+     * Maximum support by js component is 10.
+     *
+     * @var int
+     */
+    public $maxDepth = 3;
```
mvorisek: at least 5 please and then see if more is normally needed - user should never be limited!

src/FormField/ScopeBuilder.php
mvorisek: @georgehristov can we remove `|` from defaults?

src/FormField/ScopeBuilder.php
mvorisek: image

@georgehristov please file whole ui

then what feedback is left for this PR?

src/FormField/ScopeBuilder.php
georgehristov: Done

georgehristov:
@ibelar shall we merge this?

ibelar:
> shall we merge this?

@georgehristov - test is now failing:

1) atk4\ui\tests\DemosTest::testDemosStatusAndHtmlResponse with data set #40 ('form-control/query-builder.php')
31
Method Reference::getModel is deprecated. Use Model::getTheirModel instead

any idea?

georgehristov:
@ibelar Fixed - reflecting latest updates to atk4/data

src/FormField/ScopeBuilder.php
georgehristov: Fixed, I believe it is RTM

I will use in in atk4/login and we'll see if any polishing necessary

ibelar (approved):
Super!

PR #1133, opened 2020-04-25
[fix] Add stop function to ServerEvent js plugin

ibelar:
Allow stopping of SSE event.

### More info on demos/sse.php


mvorisek (approved):

abbadon1334:
The problem i mentioned, was related to browser locking when window unload event is trigger.
I solved in this way : 
https://github.com/atk4/ui/blob/e6d496b5a19b898a3933d49b5905b7a357cddf11/js/src/plugins/server-event.plugin.js#L32-L34
but i think that the listener must be removed.

this is an example of the code to reproduce the issue :
```php
        View::addTo($app)
            ->setElement('footer')
            ->addClass('ui basic segment mini')
            ->set('footer');

        $sse = jsSSE::addTo($app, ['showLoader' => false]);
        $sse->set(function () use ($sse) {

            $count = 0;
            ignore_user_abort(true); // i think this must be added to the jsSSE class
            while(!connection_aborted()) { // and this check too, obviously without the while
                sleep(1);
                $sse->send(new jsExpression('console.log('.$count.')'));
                $count++;
            }
        });

        $app->html->template->appendHTML('HEAD', '<script type="application/javascript">' . $sse->jsRender() . '</script>');
```

This code without that change prevent the browser reload

js/src/plugins/server-event.plugin.js
```diff
@@ -8,33 +8,44 @@ export default class serverEvent extends atkPlugin {
...
-            source.close();
+            that.source.close();
           }
       };
-      source.addEventListener("jsAction", function(e) {
+      this.source.addEventListener("jsAction", function(e) {
         apiService.atkSuccessTest(JSON.parse(e.data));
       }, false);
+      window.addEventListener('beforeunload', function(event) {
+        that.source.close();
```
mvorisek: @abbadon1334 What is the use case when this is needed?

abbadon1334: in a long run SSE, if you click the browser reload button, the browser will not reload, there is a sort of locking on browser side for the connection.
if you can, try the above code to reproduce the issue.

mvorisek: Ok, interesting, as F5 should always work... It does not work probably as the browser think that the original page is still not loaded yet due SSE.

mvorisek:
@ibelar , @abbadon1334 Travis is failing probably not due to this PR, but should be fixed to prevent false positives, can you look at it?

abbadon1334:
> @ibelar , @abbadon1334 Travis is failing probably not due to this PR, but should be fixed to prevent false positives, can you look at it?

restart the build and it worked, sometime travis drink beer on saturday, that's it.

js/src/plugins/server-event.plugin.js
abbadon1334: > Ok, interesting, as F5 should always work... It does not work probably as the browser think that the original page is still not loaded yet due SSE.

yes if you search there are many stackoverflow post about it.

I think that 2 php lines, can be added directly in class jsSSE to check if connection is still active or not after every flush of a push data.

This way at least on the next push of data the browser will be unlocked.

mvorisek:
> restart the build and it worked, sometime travis drink beer on saturday, that's it.

I know, but that is not a long term solution, there is probably some expected but not synchronized/event-based check.

js/src/plugins/server-event.plugin.js
ibelar: @abbadon1334 - I have made some changes and check for an aborted connection in jsSSE class.
Also, I add an option for the plugin to close connection on beforeUnload event. It is off by default but it can be turned on easily. 

Beside, I was trying to reproduce your issue without success, so not sure if I was doing it properly.
I did not have problem reloading the browser window at all.

Let me know what you think.

src/jsSSE.php
```diff
@@ -107,6 +113,10 @@ private function output(string $content): void
      */
     public function sendBlock(string $id, string $data, string $name = null): void
     {
+        if (connection_aborted()) {
```
mvorisek, line 132: Very important here is to check for `closeBeforeUnload` too, otherwise we should keep the SSE process running even if the connection between browser was interrupted.

ibelar: ` ignore_user_abort(true); ` in jsSSE::initSse() should take care of that. 
It will keep the script running, event if user aborted it.

mvorisek: Yes, but then this will call exit - and a lot of SSE tasks may rely on fact that they should normally always finish no matter if the connections/browser is still active or not.

Exit should be probably never called, user can call exit on their own if desired or an extra switch/property is needed for this as otherwise it can not be disabled.

js/src/plugins/server-event.plugin.js
mvorisek: @ibelar , @abbadon1334 Do you use the same browser?

I think reload should never be blocked, at least without any warning. So consider always disconnectiong (not aborting) the SSE.

src/jsSSE.php
```diff
@@ -114,7 +133,14 @@ private function output(string $content): void
     public function sendBlock(string $id, string $data, string $name = null): void
     {
         if (connection_aborted()) {
-            exit();
+            if ($this->onAborted && is_callable($this->onAborted)) {
+                call_user_func($this->onAborted, $this, $this->param);
```
mvorisek: There should be hook trigger instead. No `onAborted` method and no `onAborted ` not `params` props.

ibelar:
@mvorisek - @abbadon1334 

I have update the pr. Will now check if connection need to be keepAlive.
When keepAlive is set to false, it will behave as before (default)
When set to true, it will keep executing, even after user stop execution

I have also add a function that you can use when user stop execution.
Consider this. If use cancel, it will still run until counter reach a minimum value.
````
Class Counter
{
    public $count = 0;
}

$counter = new Counter();

$sse = \atk4\ui\jsSSE::addTo($app, ['showLoader' => false, 'keepAlive' => true]);

$sse->onAborted(function($t, $c) {
    if ($c->count > 10) {
        //minimum is reach
        exit();
    }

}, $counter);

$sse->set(function () use ($sse, $counter) {
    $test = true;
    while($test) { // and this check too, obviously without the while
        sleep(1);
        $sse->send(new \atk4\ui\jsExpression('console.log('.$counter->count.')'));
        $counter->count++;
        if ($counter->count > 100) {
            $test = false;
        }
    }
});
````

**Review by mvorisek — CHANGES_REQUESTED** [review]

File: src/jsSSE.php (hunk 3 above)
Comment (ibelar) [reply src/jsSSE.php]: make sense. done!

File: src/jsSSE.php [hunk 4]
```diff
@@ -41,12 +39,10 @@ public function init(): void
      * A function that get execute when user aborted this sse.
      *
      * @param callable  $fx
-     * @param mixed     $param
      */
-    public function onAborted(callable $fx, $param)
+    public function onAborted(callable $fx)
```
mvorisek, line 42: Should be removed, user can use `$this->onHook('aborted', $fx);` directly wih the full flexibility like priority.

ibelar: It does not prevent from using  `$this->onHook('aborted', $fx);` This is only a shortcode method. 
I believe it is easier for a new user for Atk when you get a direct function like this.

mvorisek: It only duplicates code and any change is there much more painfull to introduce. I have checked the other hooks and method like this is not common, so it should be removed for consistency anyway :)

ibelar: Do not agree with you on this one. This is a design choice.

I really think it makes the developer, especially the new ones, to be aware of the method. 
It is easy to know that method exists by looking at the code or when using code completion

The other advantage is that it makes it more backward-compatible.
 Form::onSubmit() is a good case. Imagine all the break we would have introduced when we changed the method name from addHook to onHook if people where using addHook('submit') instead of onSubmit.

src/jsSSE.php
ibelar: Done

js/src/plugins/server-event.plugin.js
ibelar: This pr should currently allow for customization. @abbadon1334 - let us know if you need more changes.

src/jsSSE.php
```diff
@@ -107,6 +128,15 @@ private function output(string $content): void
      */
     public function sendBlock(string $id, string $data, string $name = null): void
     {
+        if (connection_aborted()) {
+            $this->hook('aborted');
+
+            // stop execution when aborted if not keepAlive.
+            if (!$this->keepAlive) {
+                exit();
```
abbadon1334: i think this must be `$this->app->terminate()`

abbadon1334 (changes requested):
LGTM except the call to _exit()_ which is okish, but even if is already dead the interaction with the user, calling _exit_ there will not trigger the _beforeExit_ hook

src/jsSSE.php
mvorisek: Please do not duplicate any code, this is really not needed.

src/jsSSE.php
mvorisek: Good point @abbadon1334 , and as we are in js context, maybe even `terminateJson()` may be needed.

ibelar: I made the change using app->callExit(), so beforeExit trigger. There is no need to use app->terminate because, at this point, the user has disconnect the browser, so there is nothing to send to it.

src/jsSSE.php
```diff
@@ -107,6 +128,15 @@ private function output(string $content): void
      */
     public function sendBlock(string $id, string $data, string $name = null): void
     {
+        if (connection_aborted()) {
+            $this->hook('aborted');
```
mvorisek, line 133: Add `public const HOOK_ABORTED = 'aborted';` and use it here like `$this->hook(static::HOOK_ABORTED);`. Then remove `onAborted()` method.

mvorisek (changes requested):

src/jsSSE.php
abbadon1334: i agree with ibelar, some class needs helper functions, IMHO is not a duplication.

src/jsSSE.php
abbadon1334: agree to add constants for every classes that have hooks i think adding a detailed description, i suggest to add callback parameters and what happens if a value is returned.
HOOK_{hook_name}

i don't remember if sphinx can do it but if we write correct docblocks, we can inject in documentation directly to avoid write it 2 times.

abbadon1334 (approved):

src/jsSSE.php
abbadon1334: @ibelar LGTM

@mvorisek terminateJson i think is not needed, headers are already set and we need only to exit as usual with a `$content = ''` which is done in terminate

src/jsSSE.php
ibelar: I really do not see where is the problem with the onAborted() function. It is just a helper function for the developer. We will probably use this function a lot more for Saasty. An external developer may use this for installing addon to Saassty for example, so I think it is important. Same as onSubmit() is with form.

Your suggestion with static:: HOOK_ABORTED is good but and we should probably do it in a separate pr for all hook as well.

src/jsSSE.php
```diff
@@ -2,9 +2,7 @@
 
 namespace atk4\ui;
 
-/*
- * Implements a class that can be mapped into arbitrary JavaScript expression.
- */
+// Implements a class that can be mapped into arbitrary JavaScript expression.
```
mvorisek: @ibelar use `/**` (two stars) for proper phpdoc

ibelar: Done! Let's merge this in so it get include in next release.

src/jsSSE.php
mvorisek: @ibelar I still think this is wrong, but as the hook names needs to be put in constants everywhere, I have nothing againt this PR and actually 👍 for it!

I would also avoid comments like "Sending an sse action." when it is clear what the method does. It actually never helps, as the comments are almost never up too date and if programmer need to know more than "send()", then he goes to the method impl. code anyway.

mvorisek (approved):

PR #1135, opened 2020-04-25
Delete release.sh if not needed

mvorisek:
@romaninsh Please merge if this file can be removed, ie. if it is replaced fully by GH Actions.


romaninsh (approved):
agree

PR #1137, opened 2020-04-26
Fix script/style escaping in App::getTag()

### LINKED ISSUE #1136: Proper etago escaping in style/script
Currently there is crazy code like:
`$this->encodeHTML(strip_tags($value));`
which seems to be not usefull at all - stripping tags is something totally different than escaping.
Example from https://github.com/atk4/ui/pull/1111

The right solution for script/css element bodies is to escape "etago" tags, see:
- https://mathiasbynens.be/notes/etago
- https://www.man42.net/blog/2016/12/safely-escape-user-data-in-a-script-tag/#solution
- https://www.w3.org/TR/css-syntax-3/#escaping
- https://stackoverflow.com/a/25612313/5153116 (there can not be any child elements)

There should be a better review on these PRs and all security PRs should have tests.

mvorisek:
Fix https://github.com/atk4/ui/issues/1136

@georgehristov , does it solve https://github.com/atk4/ui/issues/1110 ?

The important fact is that `style` and `script` elements can not have child elements by definition.

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -481,7 +481,7 @@ class App
         if (!$this->html) {
             throw new Exception(['App does not know how to add style']);
         }
-        $this->html->template->appendHTML('HEAD', $this->getTag('style', null, $style, false));
+        $this->html->template->appendHTML('HEAD', $this->getTag('style', $style));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


mvorisek:
Help wanted with tests.

It should test:
a) all tags except script/style are escaped fully
b) script/body has escaped closing tags and comments

mvorisek:
@georgehristov Any feedback on this?

mvorisek:
Improves PR #1111. @romaninsh ok? @DarkSide666 ping

georgehristov:
The only hesitation I have on this is that I believe we should avoid hardcoding tags in the method (like `script` and `style`)

mvorisek:
@georgehristov isn't it by html spec? ref: https://mathiasbynens.be/notes/etago

georgehristov:
Yes, agree. If `getTag` is intended only for html tags (and not xml tags for instance) then we are fine as long as the html spec is covered (which appears to be the case).

DarkSide666 (approved):
LGTM

PR #1140, opened 2020-04-27
Test unit test path

ibelar:
Fix URI path generation for PHPUnit test caused by #1122 

Path after #1122 
 demos//folder/file.php

Now 
demos/folder/file.php

### DIFF
--- a/tests/DemoCallExitTest.php
+++ b/tests/DemoCallExitTest.php
@@ -61,7 +61,7 @@ class DemoCallExitTest extends BuiltInWebServerAbstract
         $files = [];
         $base_path = dirname(__DIR__) . DIRECTORY_SEPARATOR . 'demos';
         foreach ($directories as $dir) {
-            $dir_path = $base_path . '/' . $dir;
+            $dir_path = $base_path . DIRECTORY_SEPARATOR . $dir;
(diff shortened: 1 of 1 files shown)


tests/DemoCallExitTest.php
```diff
@@ -61,8 +61,8 @@ public function casesDemoFilesdataProvider()
         $files = [];
         $base_path = dirname(__DIR__) . DIRECTORY_SEPARATOR . 'demos';
         foreach ($directories as $dir) {
-            $dir_path = $base_path . '/' . $dir;
-            $pref = '/' . $dir . '/';
+            $dir_path = $base_path . DIRECTORY_SEPARATOR . $dir;
```
mvorisek, line 64: `'/'` will do the work too, later mixed with slash anyway...

mvorisek (approved):

tests/DemoCallExitTest.php
ibelar: $dir_path is used for scandir function, so probably better using the dir seperator.

mvorisek: @ibelar no, all php functions should accept `/` for any platform as a path separator

PR #1142, opened 2020-04-28
Fix - Coverage

### LINKED ISSUE #1141: Coverage is almost destroyed by PR "Reorganize demos file"
see https://github.com/atk4/ui/pull/1139

ibelar:
Coverage is back!

fix #1141


mvorisek (approved):
@ibelar The directory is resolved probably against the repo root but it was changed originally with `chroot`. Of course without `chroot` it is better and with absolute path even more better! Great, good job!

PR #1128, opened 2020-04-23
Normalize Unit Testing workflow across repos

mvorisek:
- Fixes #1071
- Fixes right image version
- Fixes #1108
- Fixes #1069 - 10x 🚀
- Moves Behat before phpunit on Travis - display results for the frontend gurus first ;-)

Migrating of Behat itself from Travis is not part of this PR.

This PR is 1:1 replacement of the original one without any BC.

Xdebug is disabled if no coverage is required (reuired = "latest" PHP which is send to CodeCov) which is the magic behind 10x faster phpunit.

PR #1143, opened 2020-04-28
Fix Wizard current step

mvorisek:


src/Wizard.php
```diff
@@ -62,7 +62,7 @@ public function init(): void
         parent::init();
         $this->stepCallback = Callback::addTo($this, ['urlTrigger'=>$this->name]);
 
-        $this->currentStep = $this->stepCallback->triggered() ?: 0;
+        $this->currentStep = (int) $this->stepCallback->triggered() ?: 0;
```
mvorisek, line 65: @ibelar, @DarkSide666 so this https://github.com/atk4/ui/blob/06fc859685cdf5f9f3d19772f4305e9eac2d99c0/src/Callback.php#L140 should return string then?

Or introduce triggeredStr() and triggered()? Or propose a better alternative.

mvorisek:
@ibelar thanks, fix applied to https://github.com/atk4/ui/pull/1128

PR #1148, opened 2020-04-29
Remove phpunit from Travis and report merged coverage from Github Actions (/wo Behat)

mvorisek:
Migrating Behat from Travis is not part of this PR.


mvorisek:
~3% drop in coverage is because previously the Behat tests were counted too. We should report coverage only from phpunit (and in the future we may report 2nd coverage from Behat if needed) - this is important to be able to run the tests separately and prevent coverage changing between jobs.

romaninsh (approved):
LGTM

PR #1149, opened 2020-04-29
Test JS CS in CI

mvorisek:


mvorisek:
@ibelar What do you propose, what command to run to check the JS CS?

What about prettier? https://prettier.io/docs/en/integrating-with-linters.html

ibelar:
> @ibelar What do you propose, what command to run to check the JS CS?
> 
> What about prettier? https://prettier.io/docs/en/integrating-with-linters.html

Sure!

mvorisek:
replaced with https://github.com/atk4/ui/pull/1280

PR #1151, opened 2020-04-29
[fix] - Header Menu Size

### LINKED ISSUE #1156: Fix admin left menu styling
Currently the left menu has some unlogical look & feel:

1. HEADER POSITION / UNIFY DIVIDERS
notice the 1px line between the top menu and the bottom/left menu. This is because the left menu has position 47 px from top instead of 48 px enforced by PR https://github.com/atk4/ui/pull/1151

The top menu has nice horizontal separators:
image
can the same sepator be used between the whole menu and the left menu?

2. LEFT MENU TOP MARGIN
once the issue above is solved, then remove "margin-top: 12px;" from the left menu so that the 1st left menu item start immediatelly with any extra space.

ibelar:
Make sure there is no gap between the top and navigation menu.
fix #1156

### DIFF
--- a/public/agileui.css
+++ b/public/agileui.css
@@ -101,7 +101,7 @@ footer.atk-footer .ui.segment {
   margin: 0;
 }
 .ui.left.sidebar {
-  z-index: 103;
+  z-index: 99;
 }
 .atk-right-panel {
   -webkit-transition: box-shadow 0.2s, right 0.2s;
--- a/public/agileui.css
+++ b/public/agileui.css
@@ -46,9 +46,11 @@
   width: 260px;
 }
 .atk-sidenav-content {
-  margin-top: 12px;
   overflow-x: hidden;
   overflow-y: auto;
+}
+.atk-sidenav-content > .item:first-child::before  {
+  display: initial !important;
 }
 /* A selected item in admin menu */
 .ui.sidebar.atk-sidenav .menu > .item.active {
--- a/public/agileui.css
+++ b/public/agileui.css
@@ -49,7 +49,7 @@
   overflow-x: hidden;
   overflow-y: auto;
 }
-.atk-sidenav-content > .item:first-child::before  {
+.ui.vertical.menu > .atk-sidenav-content > .item:first-child::before {
   display: initial !important;
 }
 /* A selected item in admin menu */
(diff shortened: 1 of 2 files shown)


ibelar:
Fix #1156

mvorisek:
> 
> 
> Fix #1156

Added my CSS rules to that issue, see. And add "fix #nn" text always to the issue description, otherwise it does not work, ie. the issue will be not closed otherwise.

public/agileui.css
```diff
@@ -101,7 +101,7 @@ footer.atk-footer .ui.segment {
   margin: 0;
 }
 .ui.left.sidebar {
-  z-index: 103;
+  z-index: 99;
```
mvorisek, line 106: Not a fix to #1156

mvorisek (changes requested):

public/agileui.css
mvorisek: Not sure what this should fix on #1156, as it does not solved it it should be probably reverted.

But otherwise PR is read to be merged

mvorisek (approved):

mvorisek:
@ibelar #1156 should be fixed, please review carefully and better rerun less build, I edited it manually.

DarkSide666 (approved):
Fixes issue for me, so LGTM

mvorisek:
> Fixes issue for me, so LGTM

@DarkSide666 because I fixed it :)

DarkSide666:
Thank you! :)

mkrecek234:
Still have the same issue, no change: Looks like that (with Maestro theme) - also manually changing the z-index does not have an effect :
image

ibelar:
Are you pulling css locally or via jsDeliver? If from jsDeliver, then give it a day at least to be updates correctly.

mvorisek:
> 
> 
> Are you pulling css locally or via jsDeliver? If from jsDeliver, then give it a day at least to be updates correctly.

PR https://github.com/atk4/ui/pull/1168 for faster collaboration ? @mkrecek234 does this help?

PS using old Admin, Maestro might need more work, please double check/fix my commits if needed.

mkrecek234:
Works now, probably then was that caching/CDN delay problem. Is it intentional to have a vertical line right to the app header? I think before there was none, probably again subjective, but looked better to me, more as a title, than a menu: 
Ohne Titel

mvorisek:
@mkrecek234 Great, thank you for your fbk. This PR should have no affect on the horizontal/top panel excep the updated height.

PR #1153, opened 2020-04-29
Fix paths in demos

mvorisek:
@ibelar please test

### DIFF
--- a/demos/atk-init.php
+++ b/demos/atk-init.php
@@ -35,7 +35,6 @@ if (file_exists(__DIR__ . '/coverage.php')) {
 $app->db = $db;
 $app->title = 'Agile UI Demo v' . $app->version;
 
-
 [$rootUrl, $relUrl] = preg_split('~(?<=/)(?=demos(/|\?|$))|\?~s', $_SERVER['REQUEST_URI'], 3);
 $demosUrl = $rootUrl . 'demos/';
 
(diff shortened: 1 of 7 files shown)


demos/atk-init.php
```diff
@@ -121,28 +124,24 @@
...
+    // Send to demos source page in github.
     $ex = [
-        new \atk4\ui\jsExpression('const baseUrl = []', [$url]),
+        new \atk4\ui\jsExpression('const baseUrl = []', ['https://github.com/atk4/ui/blob/develop']),
         new \atk4\ui\jsExpression('window.open(baseUrl + document.location.pathname, "_blank")'),
     ];
 
-    // Send to demos source page in github.
     \atk4\ui\Button::addTo($layout->menu->addItem()->addClass('aligned right'), ['View Source', 'teal', 'icon' => 'github'])
         ->on('click', $ex);
```
mvorisek: @ibelar remove the `$img` below if it is not needed

demos/atk-init.php
```diff
@@ -42,19 +42,21 @@
...
-// Need for phpUnit test only for producing right url.
+// Need for phpunit only for producing right url.
 $layout->name = 'atk_admin';
 $layout->id = $layout->name;
 
 if ($layout instanceof \atk4\ui\Layout\Navigable) {
-    $layout->addMenuItem(['Welcome to Agile Toolkit', 'icon' => 'gift'], ['/demos/index']);
+    $demosUrl = './demos/';
+
+    $layout->addMenuItem(['Welcome to Agile Toolkit', 'icon' => 'gift'], [$demosUrl]);
```
mvorisek: the original `index` was a typo?

mvorisek: I tested it now... It is horrible idea to append `.php` suffixes somewhere. User should either pass always a full path. Should be fixed.

Also reported at https://github.com/atk4/infrastructure/issues/13#issuecomment-621232048

demos/atk-init.php
ibelar: fine. not sure why it was there in first place.

demos/atk-init.php
```diff
@@ -42,19 +42,21 @@
...
 $layout = $app->layout;
-// Need for phpUnit test only for producing right url.
+// Need for phpunit only for producing right url.
 $layout->name = 'atk_admin';
 $layout->id = $layout->name;
 
 if ($layout instanceof \atk4\ui\Layout\Navigable) {
-    $layout->addMenuItem(['Welcome to Agile Toolkit', 'icon' => 'gift'], ['/demos/index']);
+    $demosUrl = preg_replace('~(?<=/demos/).+~s', '', $_SERVER['REQUEST_URI']);
+
```
ibelar: Why do you need 'REQUEST_URI' here? The value '/demos/' is only for setting proper link to menu item. There is no need fro preg_replace.

mvorisek: Consider having /demos/ in non web root and/or beiing in a demo in subdir.

mvorisek: @ibelar so ok to merge?

ibelar (approved):

mvorisek:
@ibelar fixed also GH source link, check and merge if all ok

PR #1155, opened 2020-05-01
[fix] Reload on Callback

ibelar:
Allow View to properly reload during a Callback.

Fix problem where an executing callback was to reload an external view but the view could not reload because the callback terminates first.

### Note
This might cause issue where user extends jsCallback and uses jsCallback::terminate() directly. It would need to be change to jsCallbakc::terminateAjax()

### DIFF
--- a/src/Callback.php
+++ b/src/Callback.php
@@ -133,6 +133,27 @@ class Callback
     }
 
     /**
+     * Terminate this callback
+     * by rendering the owner view.
+     */
+    public function terminate()
+    {
+        if ($this->canTerminate()) {
+            $this->app->terminateJSON($this->owner);
+        }
+    }
+
+    /**
+     * Prevent callback from terminating during a reload.
+     */
+    protected function canTerminate(): bool
+    {
+        $reload = $_GET['__atk_reload'] ?? null;
+
+        return !$reload || $this->owner->name === $reload;
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


src/Callback.php
```diff
@@ -132,6 +132,27 @@ public function set($callback, $args = [])
...
+            $this->app->terminateJSON($this->owner);
+        }
+    }
+
+    /**
+     * Prevent callback from terminating during a reload.
+     */
+    protected function canTerminate(): bool
+    {
+        $reload = $_GET['__atk_reload'] ?? null;
```
mvorisek, line 151: Seems like a hack to me which will probably not work if nested multiple times. Few more hacks and the code will explode.

mvorisek (changes requested):

src/Callback.php
ibelar: @mvorisek - chill out man - all lights are green - nothing will explode 😄 
Besides, this fixes a real problem - if you have something better to propose please do.

DarkSide666 (approved):
LGTM.
Reloading is always one view reload so it's ok to use GET parameter here. Maybe we can introduce some additional request parameter processing class or something, but ... this is enough - it gets job done and is simple, so why not.

Anyway we can change something later if we want/need, but for now it works and that's most important for related projects to succeed.

PR #1157, opened 2020-05-01
Fix admin left menu dynamic visibility

mvorisek:
Fix issue intoduced in #1112. Visibility is controlled by `isMenuLeftVisible` prop and needs to be not always forced to visible.

### DIFF
--- a/template/semantic-ui/layout/admin.html
+++ b/template/semantic-ui/layout/admin.html
@@ -1,6 +1,6 @@
 
 <div class="atk-layout">
-  <div class="ui left vertical menu visible inverted labeled sidebar atk-sidenav">{$LeftMenu}</div>
+  <div class="ui left vertical menu inverted labeled sidebar atk-sidenav">{$LeftMenu}</div>
   <main class="atk-mainContainer atk-admin-layout">{$TopMenu}
     <div class="atk-mainContainerWrapper">{$Content}</div>
   </main>
--- a/src/Layout/Admin.php
+++ b/src/Layout/Admin.php
@@ -108,7 +108,7 @@ class Admin extends Generic implements Navigable
                 $this->menuLeft->addItem(['Dashboard', 'icon' => 'dashboard'], ['index']);
             }
             if ($this->isMenuLeftVisible) {
-                $this->menuLeft->addClass('visible');
+                $this->menuLeft->js(true)->parent()->addClass('visible');
             }
         }
 
--- a/template/semantic-ui/layout/admin.pug
+++ b/template/semantic-ui/layout/admin.pug
@@ -1,5 +1,5 @@
 .atk-layout
-  .ui.left.vertical.menu.visible.inverted.labeled.sidebar.atk-sidenav
+  .ui.left.vertical.menu.inverted.labeled.sidebar.atk-sidenav
     | {$LeftMenu}
   main.atk-mainContainer.atk-admin-layout
     | {$TopMenu}


template/semantic-ui/layout/admin.html
```diff
@@ -1,6 +1,6 @@
 
 <div class="atk-layout">
-  <div class="ui left vertical menu visible inverted labeled sidebar atk-sidenav">{$LeftMenu}</div>
+  <div class="ui left vertical menu inverted labeled sidebar atk-sidenav">{$LeftMenu}</div>
```
ibelar, line 3: This will close the left menu on each page load and this is not the intended behavior.

mvorisek: See https://github.com/atk4/ui/blob/3fd37f0bee0d43f01e9eccb41e96b9dc3466d5ca/src/Layout/Admin.php#L111

If you need to use it anywhere else, you have to add the visible class yourself. But if it is hardcoded, it it wrong and it breaks what Admin behaviour. Ok to merge?

ibelar:
let's merge #1151 first and see what can be done after. 
Bringing the z-index lower for the left menu should fix the border state in #1156

mvorisek:
This visibility fix is unrelated with any current PRs!

template/semantic-ui/layout/admin.html
ibelar: Then  make it open by default ;) not close by default

mvorisek: But then it can flash... I am just fixing what was broken by PR #1112. Is there any issue with this fix?

ibelar: > This will close the left menu on each page load and this is not the intended behavior.

The left menu should be open on page load on large screen!

mvorisek: That is what the option `isMenuLeftVisible` is for and it worked perfectly, you just have to solve the hardcoded visibility in the template - this is solved by this PR and check, make the `isMenuLeftVisible` working again. Btw. `isMenuLeftVisible` is set to true by default, so your wanted behaviour will be matched.

ibelar: then set it to true by default ;-)

mvorisek: https://github.com/atk4/ui/blob/3fd37f0bee0d43f01e9eccb41e96b9dc3466d5ca/src/Layout/Admin.php#L42 it is on by default when the non-working autoopener code is fixed.

What need to be done next:
- fix https://github.com/atk4/ui/blob/3fd37f0bee0d43f01e9eccb41e96b9dc3466d5ca/src/Layout/Admin.php#L111 probably by moving/fixing it to init() here https://github.com/atk4/ui/blob/3fd37f0bee0d43f01e9eccb41e96b9dc3466d5ca/src/Layout/Admin.php#L57 and do the same what the 1st click of the visibility button does
- commit it on top of this PR
- merge it

ibelar: and why not doing it in this PR then?

mvorisek: I have no resource to fixing these... I am not a front end guy :)

ibelar: This PR will break the intended behavior, either submit a fully working PR or just submit an issue and I will check when I have some time.

mvorisek: Please help with this line https://github.com/atk4/ui/pull/1157/files#diff-5bb52268570a74fcdec90d6f8b8f3ab9R111 , I do not know the right Atk way how to add the visibility class.

ibelar: Leave it to me. The template needs to be changed. I will do it when I have a chance.

mvorisek:
@ibelar PR is complete and fully working, please check and merge

ibelar (approved):
Look good now!

mvorisek:
@ibelar it does fixed the original switching behaviour, but if default to on, the menu "flashes" between loads which degrades the UX. Can you fix this in server-side render?

georgehristov:
I have below code in my custom layout to control the state of navigation menu. It basically persists the state in the session. So basically adding the `SessionTrait` to `Layout` and using following code.
Good solution? Shall I make a PR?

```
$this->isMenuLeftVisible = $this->learn('menu', $this->isMenuLeftVisible);
        
$this->burger->add(['Icon',	'content'])->on('click', [
        		(new jQuery('.ui.left.sidebar'))->toggleClass('visible'),
        		(new jQuery('.epesi-logo'))->toggleClass('expanded'),
        		(new jQuery('body'))->toggleClass('atk-leftMenu-visible'),
        		$this->burger->add('jsCallback')->set(function($j, $visible) {
        			$this->memorize('menu', filter_var($visible, FILTER_VALIDATE_BOOLEAN));
        		}, [$this->menuLeft->js(true)->hasClass('visible')])
)])
]);
```

mvorisek:
@georgehristov it is not about persisting the state - it is about adding/removing `visible` class on the server side thus the browser can render the menu immediatelly which will prevent the flashing introduced by solving this in the JS - which takes some time before the JS is triggered/applied. Your code does not solve this (not-using the JS).

georgehristov:
I see now. I do not have this problem as I use routing where only the content area of the page is updated on page change.

mvorisek:
@georgehristov do you solve there a history? https://github.com/atk4/ui/issues/1191

georgehristov:
Yes, there is a browser history. I have not had problem with that.

mvorisek:
> I see now. I do not have this problem as I use routing where only the content area of the page is updated on page change.

I am lost, you either reload the full page - and this issue is affecting you too - or you use JS to reload the inner component, but then need the history API.

PR #1158, opened 2020-05-01
Remove support of relative class names

mvorisek:
Relative names are absolutely no longer needed when classes are referred by `::class` expr.

This PR removes the deprecated behaviour. It breaks BC, but it prevent bad use of unrefactorable classes which is stronger.

To fix your projects check the results of regex (for bests result use case-sensitive matching mode and do in the order like here):

- `atk4\ui\FormLayout\Section` (used only with `->addSubLayout()`)
   ```
   ['"]([^'" ]+\\)?(Accordion|Columns|Generic|Tabs)['"]
   ```
- `atk4\ui\FormLayout`
   ```
   ['"]([^'" ]+\\)?(_Abstract|Columns|Custom|Generic)['"]
   ```
- `atk4\ui\Layout`
   ```
   ['"]([^'" ]+\\)?(Admin|Centered|Column|Generic|Maestro|Navigable|Product)['"]
   ```
- `atk4\ui\FormField`
   ```
   ['"]([^'" ]+\\)?(AutoComplete|DropDownCascade|Hidden|Lookup|Password|TreeItemSelector|Calendar|DropDown|Input|Money|Radio|UploadImg|CheckBox|Generic|Line|MultiLine|TextArea|Upload)['"]
   ```
- `atk4\ui`
   ```
   ['"]([^'" ]+\\)?(Accordion|HelloWorld|jsNotify|Locale|Table|AccordionSection|Console|Icon|jsPaginator|LoremIpsum|Tab|CRUD|Image|jsReload|Menu|Tabs|App|DropDownButton|Item|jsSearch|Message|TabsSubView|BreadCrumb|DropDown|ItemsPerPageSelector|jsSortable|Template|Button|jQuery|jsSSE|Modal|Text|CallbackLater|Exception|jsCallback|jsToast|Paginator|View|Callback|jsChain|jsVueService|VirtualPage|CardDeck|jsConditionalForm|jUniv|Wizard|Card|Form|jsExpressionable|Label|Popup|CardSection|GridLayout|jsExpression|ProgressBar|CardTable|Grid|jsFunction|Lister|Step|Columns|Header|jsModal|Loader|TableColumn)['"]
   ```

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -102,7 +102,9 @@ jobs:
 
       - name: Check Coding Style (only for CodingStyle)
         if: matrix.type == 'CodingStyle'
-        run: vendor/bin/php-cs-fixer fix --dry-run --using-cache=no --diff --diff-format=udiff --verbose --show-progress=dots
+        run: |
+          if [ "$(find demos/ -name '*.php' -print0 | xargs -0 grep -L "namespace atk4\\\\ui\\\\demo;" | tee /dev/fd/2)" ]; then echo 'All demos/ files must have namespace declared' && (exit 1); fi
+          vendor/bin/php-cs-fixer fix --dry-run --using-cache=no --diff --diff-format=udiff --verbose --show-progress=dots
 
       - name: Upload coverage logs 1/2 (only for "latest" Phpunit)
         if: env.LOG_COVERAGE
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ In 1.7 we now rely on VueJS for new components and the first component to make u
 $f = \atk4\ui\Form::addTo($app);
 
 // Add multiline field and set model.
-$ml = $f->addField('ml', ['MultiLine', 'options' => ['color' => 'blue']]);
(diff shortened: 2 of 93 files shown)


mvorisek:
Coverage is not changed at all - https://codecov.io/gh/atk4/ui/pull/1158/changes

mvorisek:
@romaninsh legit?

romaninsh:
Agree with @mvorisek - it's an old and bad coding style, should probably be dropped in favour of namespaces.

romaninsh:
@mvorisek - could this be implemented with a "warning" at first ? through a logger mechanics.

mvorisek:
> could this be implemented with a "warning" at first ? through a logger mechanics.

@romaninsh It can, but you will not gain anything except error in a different format, I think the best ways to update your code is to list all your classes and then search for `'([^']+\\)?(short_name_1|short_name_2|...)'` and fix all results represention a class (as it may be regular string)

DarkSide666 (approved):
Probably this will break quite a lot of projects, but ... yeah, this is right way to do things so _let's go_ with it.

Also it would be good if you could add this regex you mentioned somewhere in release notes so developers can use it when refactoring their projects on newer toolkit version.

mvorisek:
> this regex

I will do shortly, 👌

PR #1167, opened 2020-05-02
Fix wrong URL fallbacks

mvorisek:
The fallbacks were wrong as the expected the App to be called in the same dir - with subdirectory, the files were failing to load.

There is no need for them as they are set by defaut and can be set to `../public` by user if needed.

BLOCKED BY: https://github.com/atk4/ui/issues/1189

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -449,22 +449,18 @@ class App
     public function initIncludes()
     {
         // jQuery
-        $url = $this->cdn['jquery'] ?? '../public';
-        $this->requireJS($url . '/jquery.min.js');
+        $this->requireJS($this->cdn['jquery'] . '/jquery.min.js');
 
         // Semantic UI
-        $url = $this->cdn['semantic-ui'] ?? '../public';
-        $this->requireJS($url . '/semantic.min.js');
-        $this->requireCSS($url . '/semantic.min.css');
+        $this->requireJS($this->cdn['semantic-ui'] . '/semantic.min.js');
+        $this->requireCSS($this->cdn['semantic-ui'] . '/semantic.min.css');
 
         // Serialize Object
-        $url = $this->cdn['serialize-object'] ?? '../public';
-        $this->requireJS($url . '/jquery.serialize-object.min.js');
+        $this->requireJS($this->cdn['serialize-object'] . '/jquery.serialize-object.min.js');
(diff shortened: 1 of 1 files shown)


romaninsh (changes requested):
Currently user can create "public" folder and the files would be included from it locally. If this PR goes through, we'll be always using CDN.

Commercial projects would typically use locally generated files, so to me it seems like a BC break.

mvorisek:
Explained in the description why this is wrong. Bad usage can be very easily solved by setting the cdns locally.

DarkSide666:
Idea as such is good - developer can always set cdn = '../public' himself, but still this is small compatibility break.

mvorisek:
@romaninsh so ok?

romaninsh:
I'm not against this change but it will be a BC.

romaninsh:
@mvorisek - we're still trying to figure out a good way to deal with breaking changes. Last meeting we agreed to label those issues and if any of the issues with BC are to be merged, we would need to issue a new minor release, e.g. 2.1.x.

It would also be important for all the breaking changes to provide a guide no how to update check if your app is affected and how to upgrade. It would be nice to see this mechanism working!

I'll raise some awareness of this issue and we can pilot it forward.

Ideally we would also need a guide in the documentation explaining how to implement local file builds.

romaninsh:
@mvorisek I'd like to deal with https://github.com/atk4/ui/issues/1189 this weekend and send out 2.1 release for ATK UI. After that I'll merge this PR.

Until then - please provide a guide how existing projects can "identify" BC break and prepare for it ahead of time.

mvorisek:
> Until then - please provide a guide how existing projects can "identify" BC break and prepare for it ahead of time.

I have already proposed simple changelog/migration guide on Wiki. Anyone can then "identify BC breaks" on that. All BC breaks should target to always reject wrong/original input.

mvorisek:
Removed "BLOCKED" as "BC-break" is enought. This change should not affect almost any user (if any) as the CDNs are normally set by default, thus no fallback.

DarkSide666 (approved):
You can always set `$this->cdn['jquery'] = '../public'` and then it will be compatible, so this is actually quite small BC break. Good to merge.

PR #1168, opened 2020-05-02
Fix cdn caching for develop

mvorisek:
Current CDN caches the resource for about a day, this fixes it for the develop.

Also Unify CDNs in favor of Cloudflare.

This PR helps, confirmed by https://github.com/atk4/ui/pull/1151#issuecomment-623169453

> Works now, probably then was that caching/CDN delay problem.


romaninsh (approved):
This wasn't used because there were claims that githubusercontent is discontinued, maybe that's not the case, I can't find any info about this, so approving.

PR #1170, opened 2020-05-03
Toast Demo titles and messages are messed up

TheNetworkIsDown:
Correct the display of Toast demo


georgehristov (approved):
LGTM, thanks!

PR #1177, opened 2020-05-05
Delete agileui.js

mvorisek:
Old stale file, use can create if needed.

WARNING: This PR is to test the feedback, need more work.


romaninsh:
Isn't it needed? Could we create it for release / demo?

mvorisek:
Your friends with help you to decide if this is $0 worth or not!

DarkSide666:
I think this file wasn't needed and probably can/should be removed.
But let @ibelar review this.

ibelar:
Sure this can be deleted. Not use anymore. Behat test setup would need to remove compression of this file as well.

acicovic:
So since @ibelar said this can be deleted, perhaps we should go forward and merge this?

ibelar:
@acicovic - Better need to wait. There are still work need to be done on this PR. Behat test setup need to be check. I will set it up as draft for now.

romaninsh:
Wouldn't that break menu?

ibelar:
@romaninsh - No. Side menu does not rely on this file anymore.

mvorisek:
@romaninsh As far as I know this file is specified at least in release script - that is why I marked this PR as draft, drafts should be never merged by someone else.

PR #1162, opened 2020-05-02
Test generated JS/CSS files in CI

mvorisek:
Very important generated files match the source code.

`--compatibility=ie8` not needed if support for IE10 is enough (it should be in y2020).

PR #1178, opened 2020-05-06
Generate JS files in CI/CD

romaninsh:
Superseding https://github.com/atk4/ui/pull/1162 ( check that diff - https://github.com/atk4/ui/pull/1162/files )


.travis.yml
```diff
@@ -59,7 +59,7 @@ before_script:
     - cp demos/db.example.php demos/db.php
 
     - (cd js; npm install --loglevel=error; npm run build)
-    - lessc public/agileui.less public/agileui.css  --clean-css="--s1 --advanced --compatibility=ie8" --source-map
+    - lessc public/agileui.less public/agileui.css --clean-css="--s1 --advanced" --source-map
     - uglifyjs --compress -- public/agileui.js > public/agileui.min.js
```
ibelar: Uglifyjs can be removed here since we do not need agileui.js anymore. There is probably no ned to install uglifyjs as well.

.github/workflows/unit-tests.yml
```diff
@@ -68,14 +68,22 @@ jobs:
...
+          # atk js
+          (cd js && npm run build)
+
+          # release command:
+          # lessc public/agileui.less public/agileui.css --clean-css="--s1 --advanced" --source-map
+          lessc public/agileui.less public/agileui.css
+
+          # app js
+          # ignored from git - check syntax only
+          uglifyjs --compress -- public/agileui.js > public/agileui.min.js
```
ibelar: No need for uglifyjs since we are not using agileui.js anymore.

mvorisek:
See https://github.com/atk4/ui/pull/1179#issuecomment-624508082 . I no longer think this is good, these files were added 20 days ago by @ibelar only as "temporary" and I think the negative points predominate by having these files in git.

This PR does not provide any value except it remove security checks (diffs) from my PR. 😒

ibelar (approved):
LGTM

PR #1179, opened 2020-05-06
Restore files as this is breaking existing ATK installations

romaninsh:
Partial revert of #1178 - I had to return minified files as this is breaking existing ATK installations.

We should leave those files until the following URLs no longer get traffic:
 - https://cdn.jsdelivr.net/gh/atk4/ui@develop/public/atkjs-ui.min.js
 -  https://cdn.jsdelivr.net/gh/atk4/ui@develop/public/agileui.css


romaninsh:
bypassing approval because of potential production outage.

mvorisek:
@romaninsh these files were added 20 days by @ibelar in https://github.com/atk4/ui/pull/1106 as later explained by him as "temporary". So as fast you revert these again as less the traffic will be.

Initially I was for adding files like these to the repo too. I thinked over it twice and considering the discussion behind I do not think this is good anymore.

I think there is no strong advantage over the new solution with builds in the cloud. But bundling the compiled files in develop branches has strong disadvantage of having to always recompile the files, consider:
- impossible to merge with git/GH only
- impossible to do small fixes without IDE/compiling toolset
- the need to always compile these files in production mode
- the need to stage/manage multiple files in git
There may be more points not to do so.

You almost beat me by including some unknown/potentially malicious/compiled file and you are self doiing it? You try to be an example and then self approvings PR?

I have nothing against it, but if I do exactly the same, you are creaming like police officer and start counting the violations. Here is an evidence:
- https://github.com/atk4/infrastructure/pull/14
- https://github.com/atk4/infrastructure/pull/15
Is that right?

romaninsh:
Yeah, I haven't considered that ATK is relying on jsdeliver that's pointing to develop branch.

After some time we can nuke the files out, but we need to do a release first. 

So technically files are there, they are same as @ibelar have bundled himself, but i'll reject all attempts to change those files and they will be removed soon.

Regarding policing - you are correct and I would really appreciate if you could join our meeting this Friday so that we could find a good ways for the future. Our existing practices are not ideal, but if we change them we need agreements from everybody.

mvorisek:
> Yeah, I haven't considered that ATK is relying on jsdeliver that's pointing to develop branch.

I still think supporting never released temporary solution introduced less than a month ago is wrong and it should be reverted without any delay. Do I miss some strong usecase to support with this PR except a few people checkouted the develop/unstable version in past days?

> Meeting

I am very clear about me, my intentions and my very limited time. For these reasons I am not attending this meeting. In short I find this project interesting as a low-code framework but the core APIs need to be:
- strict (= less possibilities to write something, no silenced errors etc.)
- well defined (=intuitive from code itself)
- and well maintable (=refactorable).

As maintaining a private fork is very costly I am offering my work to be integrated directly. I know some PRs like https://github.com/atk4/data/pull/565 can break full BC if the usage of these APIs is other than strict, but this is expected and all even the largest projects went thru this incl. Symfony, phpunit...

For these reasons I already did:
- fixed CS - for stricter syntax, for less merge conflicts, less individual freedom...
- fixed phpunit - for latest devtools, for better CI, for faster iterations/dev
- fixed some APIs already - for intuitive and refactorable code (like `component::addTo()`)

I think we should do these API changes in 2.x and offer existing users to stay on 1.x, their locked versions and/or SIMPLE CHANGELOG what to do to upgrade from xx to yy. I stress the word simple, as we should write there only major changes that affects almost everyone and keep minor changes for error exceptions. (we are not NASA and making this changelog long will destroy the usability, thus noone will be using it). This changelog can be a single wiki page.

For the largest changes in ui/data we may even want to offer a scripted fixer. I did this for `::addTo` refactoring, script here https://github.com/mvorisek/atk4_ui_add_migrate/blob/master/ui_add_migrate.php - if was not a simple function rename, it was quite complex as it rebuilt arguments, refactored string names to classes, fixed a lot of silent mistakes... If I remember correctly there were like ~1000 locations to change and I fixed even commented code with it. But there are like less than 5 so big changes in the project lifetime. But I solved all the issues from A-Z with minimal BC break, you probably would not even noticed expect the new recommended syntax. Responsibility. If you would require me to wait between small changes, it would be probably never done or least least not in that quality. If you tell me to wait till tomorrow, I have no time to wait till another tomorrow.

I have no time to provide support nor review PRs of others in general. I belive in people and their own responsibility. In examples - @DarkSide666 extended data with WITH support. I have no time to check and review each single of PR like these. But I trust him that he put a max. efford in it and went thru all possibilities before merging the code. If something is broken, I expect him to react fast. In median time in an hours to provide a fix or at least reaction. I belive that concept of having to approve PRs is great/needed for unexperienced/foreign contributors but otherwise it is very costly and does solve less than 1/2 of the issues. Ask yourself, how many time you caught an mistake before review/merge and after it? Now consider the cost in resources. Is better to catch a few mistakes in review or reduce distraction of the devs to minimum againt having them happy and ready to fix also uncaught mistakes when needed?

So BC and trust are my points to you guys to discuss. Not agreeing in these points is basically unacceptable to me to cooperate any further.

I do not care if PHP will be deprecated now or a month later. I can live with both decisions but If I can have a preference, I will choose to drop support for any other PHP version older than 2 months (=stable and avaialble in Debian/Ubuntu thru sury repo) for develop. (as for ex. in PHP 8.0 you can use `static` return type and let the devs rock even more :) ) This question is not about BC in user's code, BC can stay, it is about if we want to have a free hands inside or be slowly dying in nonsense.

For the longer future I think this project should be actually reduced to be as thin as possible (ie. to use mature libs as much as possible) and focused strongly on its USP (low code ui, data abstraction etc.). You can measure my words on this at https://github.com/atk4/ui/graphs/contributors and also at other repos. I usually simplify code and shoot everyone who have the courage to copy paste anything. Code, comments, ...

I also think this project can be almost magically improved by rendering the frontend thru Vue/React (and we already has almost the ideal structures like templates, data, callbacks) which can reduce the JS to minimum and offer maximum control of frontend thru backend (understand like no totally limitted conditional fields etc., simply use robust React components and do whatever you want purely in PHP) and other never-beatable technologies.

So this is my complete answer to every meeting this year.

PR #1180, opened 2020-05-06
Doc: update "Javascript mappings"

TheNetworkIsDown:


docs/js.rst
```diff
@@ -71,57 +68,57 @@ All of the mentioned 4 examples will produce a valid "action" object that can be
...
+Agile UI also provides support for an `on` event binding. This allows to apply events on
 multiple elements::
 
-    $buttons = new Buttons();
+    $buttons = $app->add('View');
 
-    \atk4\ui\Button::addTo($buttons, ['One']);
-    \atk4\ui\Button::addTo($buttons, ['Two']);
-    \atk4\ui\Button::addTo($buttons, ['Three']);
+    $buttons->add(['Button', 'One']);
```
georgehristov: Any reason to revert the code to "old style" adding?

georgehristov (commented):
Thanks a lot but please advise on the sample code snippets that were reverted to `View::add` instead of `View::addTo`

docs/js.rst
```diff
@@ -71,57 +68,57 @@ All of the mentioned 4 examples will produce a valid "action" object that can be
...
+Finally, you can specify the name of the JavaScript event::
 
     $view->js('click')->hide();
 
-Agile UI also provides support for an `on` event binding. This allows you to apply events on
+Agile UI also provides support for an `on` event binding. This allows to apply events on
 multiple elements::
 
-    $buttons = new Buttons();
+    $buttons = $app->add('View');
```
ibelar: You are right Buttons::class does not exist. 😄
May be we can show how to create a button container using Fomantic UI like this:
````
$buttons = View::addTo($app, ['ui' => 'basic buttons']);
````

File: docs/js.rst [hunk 3]
```diff
@@ -71,57 +68,57 @@ All of the mentioned 4 examples will produce a valid "action" object that can be
...
 
-    \atk4\ui\Button::addTo($buttons, ['One']);
-    \atk4\ui\Button::addTo($buttons, ['Two']);
-    \atk4\ui\Button::addTo($buttons, ['Three']);
+    $buttons->add(['Button', 'One']);
+    $buttons->add(['Button', 'Two']);
+    $buttons->add(['Button', 'Three']);
```
ibelar, line 98: Please keep format for adding view like this:
````
Button::addTo($buttons, [$seed])
````
It has been update like this in all docs and we should keep it standard among documentation.

File: docs/js.rst [hunk 4]
```diff
@@ -238,56 +232,38 @@ Several usage cases for plain `js()` method. The most basic scenario is to perfo
...
 .. php:method:: on(String $event, [String selector], $callback = null)
 
     Returns chain that will be automatically executed if $event occurs. If $callback is specified, it
     will also be executed on event.
 
-The following code will show 3 buttons and clicking any button will hide itself. Only a single action is created::
+The following code will show three buttons and clicking any one will hide it. Only a single action is created::
 
-    $buttons = Buttons();
+    $buttons = new \atk4\ui\View;
```
ibelar, line 243: same here for Buttons

docs/js.rst
TheNetworkIsDown: Uh oh, sorry for that. I thought that was the "old" syntax. In fact when I tried with my install of 2.x (maybe a month old?) that didn't work ("addTo not found")...

ibelar: It was probably because the release was not synced yet with the latest core package when you tried, because addTo is part of the atk4/core package. 
Good changes in the doc. tks.

romaninsh: yeah, that's only in the "develop" branch an in the ALPHA release of 2.1.

Both syntaxes work, but we want to be consistent for new users.

ibelar (approved):
Well done! Thks

PR #1181, opened 2020-05-06
[fix] Radio Input Field

ibelar:
Fix radio input when using strict comparison.


romaninsh (approved):
better comparing..

romaninsh:
I approve that, but consider what if $value argument is non-string.

ibelar:
I have discussed this with @DarkSide666 - Looks like $value is always a string, coming from atk4/data, even when using `'values' => [1 => 'one'] ` in field definition.
On the other side, `$lister->model->id` may contain int or string.

PR #1182, opened 2020-05-06
[fix] Css z-index value

ibelar:
Will display the left navigation menu on top of the overlay when used in smaller screen sizes.


romaninsh (approved):
LGTM

PR #1183, opened 2020-05-08
[fix] Card Deck Reload on first page

ibelar:
Now, rely on Paginator to get page numbers.

Prevent problems when reloading on the first page.


DarkSide666 (approved):
LGTM

PR #1187, opened 2020-05-13
[feature] Allow Modal to render in VirtualPage

ibelar:
This will allow model action executor to be able to run in VirtualPage

### DIFF
--- a/src/VirtualPage.php
+++ b/src/VirtualPage.php
@@ -33,7 +33,7 @@ class VirtualPage extends View
         parent::init();
 
         $this->cb = $this->_add([Callback::class, 'urlTrigger' => $this->urlTrigger ?: $this->name]);
-        $this->app->stickyGet($this->name);
+        $this->app->stickyGet($this->cb->urlTrigger);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/VirtualPage.php
```diff
@@ -146,10 +146,21 @@ public function getHTML()
...
+                    $modalHtml .= $view->getHTML();
+                    $this->app->layout->_js_actions = array_merge($this->app->layout->_js_actions, $view->_js_actions);
+                }
+            }
+
             $this->app->layout->template->setHTML('Content', parent::getHTML());
             $this->app->layout->_js_actions = array_merge($this->app->layout->_js_actions, $this->_js_actions);
 
             $this->app->html->template->setHTML('Content', $this->app->layout->getHTML());
+            $this->app->html->template->setHtml('Modals', $modalHtml);
```
PhilippGrashoff: setHTML or setHtml? when using strict_types this will fail if case does not match afaik.

ibelar: Thanks for catching. Updated!

DarkSide666 (approved):
LGTM

romaninsh (approved):

PR #1190, opened 2020-05-14
[fix] Filter Column when using decorator

### LINKED ISSUE #1188: Decorated grid column prevents use of addFilterColumn()
If you are decorating a grid column with 
`$grid->addDecorator('name', ['Link', ['ticket'], ['tid' => 'id']]);`
the Filter column does not work anymore:
`$grid->addFilterColumn();` 

throws the error "Error : Call to a member function addPopup() on array" on Grid.php:473

ibelar:
Fix: #1188

### DIFF
--- a/src/Table.php
+++ b/src/Table.php
@@ -329,7 +329,7 @@ class Table extends Lister
      *
      * @return mixed
      */
-    public function getColumn(string $name)
+    protected function getColumn(string $name)
     {
         // NOTE: It is not guaranteed that we will have only one element here. When adding decorators, the key will not
         // contain the column instance anymore but an array with column instance set at 0 indexes and the rest as decorators.
--- a/src/Table.php
+++ b/src/Table.php
@@ -276,7 +276,7 @@ class Table extends Lister
 
         // create column popup.
         foreach ($cols as $colName) {
-            $col = $this->columns[$colName];
+            $col = $this->getColumn($colName);
             if ($col) {
                 $pop = $col->addPopup(new FilterPopup(['field' => $this->model->getField($colName), 'reload' => $this->reload, 'colTrigger' => '#' . $col->name . '_ac']));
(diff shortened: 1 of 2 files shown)


ibelar:
@mvorisek - look like #1137 broke cs fixer test. I can push the fix to this pr, but could you please make sure that the suggested changes are ok?

mvorisek:
> 
> 
> @mvorisek - look like #1137 broke cs fixer test. I can push the fix to this pr, but could you please make sure that the suggested changes are ok?

👍 , suggested changes are perfect

src/Table.php
```diff
@@ -326,6 +326,18 @@ public function getColumnDecorators($name)
...
+    /**
+     * Return column instance or first instance if using decorator.
+     *
+     * @param $name
+     *
+     * @return mixed
+     */
+    public function getColumn($name)
+    {
+        return is_array($this->columns[$name]) ? $this->columns[$name][0] : $this->columns[$name];
```
mvorisek, line 337: is `$this->columns[$name]` guaranteed to have always only one element? If no, the decorator API should be unified. And then this method should be not needed.

mvorisek (changes requested):

src/Table.php
ibelar: No, it is not guaranteed. When adding decorators, the key will not contain the column instance anymore but an array with column instance set at 0 indexes and the rest as decorators. This is enough for fixing this issue right now. We can work on unifying decorator API in a separate PR.

mvorisek: Introducing new methods in a hope to fix them later is bad as it takes much more efford to remove them later.

If this is only hotfix, make the method private and comment this there - and at best create a PR for it and add that link there too.

DarkSide666 (approved):
Added comment and changed to protected. OK for now.
Probably this new method should also be reused in other methods where we access columns property directly.

PR #1194, opened 2020-05-15
Use MB functions

DarkSide666:


src/FormField/DropDown.php
```diff
@@ -110,8 +110,8 @@ class DropDown extends Input
      * function($value, $key) {
      *     return [
      *        'value' => $key,
-     *        'title' => strtoupper($value),
-     *        'icon'  => strpos('Month', $value) !== false ? 'calendar' : '',
+     *        'title' => mb_strtoupper($value),
+     *        'icon'  => mb_strpos('Month', $value) !== false ? 'calendar' : '',
```
mvorisek: strpos

src/FormField/MultiLine.php
```diff
@@ -912,7 +912,7 @@ private function getDummyExpression($exprField, $model)
         preg_match_all('/\[[a-z0-9_]*\]|{[a-z0-9_]*}/i', $expr, $matches);
 
         foreach ($matches[0] as $match) {
-            $fieldName = substr($match, 1, -1);
+            $fieldName = mb_substr($match, 1, -1);
```
mvorisek: no sense, this is not an operation on text, instead it removes ascii brackets (yes, op on text :D)

src/App.php
```diff
@@ -1092,7 +1092,7 @@ protected function outputResponse(string $data, array $headers): void
                     http_response_code($v);
                 } else {
                     $kCamelCase = preg_replace_callback('~(?<![a-zA-Z])[a-z]~', function ($m) {
-                        return strtoupper($m[0]);
+                        return mb_strtoupper($m[0]);
```
mvorisek: Not a good review, regex is ascii only!

src/App.php
```diff
@@ -927,10 +927,10 @@ public function getTag($tag = null, $attr = null, $value = null)
 
         if ($value !== null) {
             $result = [];
-            foreach ((array)$value as $v) {
+            foreach ((array) $value as $v) {
                 if (is_array($v)) {
                     $result[] = $this->getTag(...$v);
-                } elseif (in_array($tag, ['script', 'style'])) {
+                } elseif (in_array($tag, ['script', 'style'], true)) {
```
mvorisek, line 933: Hey, let fix this in develop immediatelly.

src/App.php
```diff
@@ -683,7 +683,7 @@ public function url($page = [], $needRequestUri = false, $extraRequestUriArgs =
 
         if ($this->page === null) {
             $requestUrl = $this->getRequestURI();
-            if (substr($requestUrl, -1, 1) === '/') {
+            if (mb_substr($requestUrl, -1, 1) === '/') {
```
mvorisek: only missleading

src/App.php
```diff
@@ -583,7 +583,7 @@ public function loadTemplate($name)
         $template = new Template();
         $template->app = $this;
 
-        if (in_array($name[0], ['.', '/', '\\'], true) || strpos($name, ':\\') !== false) {
+        if (in_array($name[0], ['.', '/', '\\'], true) || mb_strpos($name, ':\\') !== false) {
```
mvorisek: only missleading

src/App.php
```diff
@@ -352,7 +352,7 @@ public function terminate($output = '', array $headers = []): void
             $headers['content-type'] = $this->response_headers['content-type'];
         }
 
-        $type = preg_replace('~;.*~', '', strtolower($headers['content-type'])); // in LC without charset
+        $type = preg_replace('~;.*~', '', mb_strtolower($headers['content-type'])); // in LC without charset
```
mvorisek: content type (and headers in general) are strictly taken ascii only, I would never use mb_ here

src/App.php
```diff
@@ -290,12 +290,12 @@ protected function normalizeHeaders(array $headers): array
     {
         $res = [];
         foreach ($headers as $k => $v) {
-            if (is_numeric($k) && ($p = strpos($v, ':')) !== false) {
-                $k = substr($v, 0, $p);
-                $v = substr($v, $p + 1);
+            if (is_numeric($k) && ($p = mb_strpos($v, ':')) !== false) {
```
mvorisek: see below, op on headers, I do not think this is good, this method should be also as fast as possible

features/bootstrap/FeatureContext.php
```diff
@@ -124,7 +124,7 @@ public function dump($arg1)
     public function iDontSee($arg1)
     {
         $element = $this->getSession()->getPage()->find('xpath', '//div[text()="' . $arg1 . '"]');
-        if (strpos('display: none', $element->getAttribute('style')) !== false) {
+        if (mb_strpos('display: none', $element->getAttribute('style')) !== false) {
```
mvorisek, line 127: let's live with this... Even if `mb_strpos` should be never needed if both strings are utf-8 valid.

demos/form/form2.php
```diff
@@ -42,10 +42,10 @@
 $form->onSubmit(function ($f) {
     // In-form validation
     $errors = [];
-    if (strlen($f->model['first_name']) < 3) {
+    if (mb_strlen($f->model['first_name']) < 3) {
```
mvorisek, line 45: this is an example where it makes real sense...

demos/collection/jssortable.php
```diff
@@ -13,7 +13,7 @@
 
 $lister = \atk4\ui\Lister::addTo($view, [], ['List']);
 $lister->onHook('beforeRow', function ($l) {
-    $l->current_row['iso'] = strtolower($l->current_row['iso']);
+    $l->current_row['iso'] = mb_strtolower($l->current_row['iso']);
```
mvorisek, line 16: why not, but I would never probably write this for ISO, which is normally ASCII only, but fine, less fixes to check year later :D

mvorisek (changes requested):

src/App.php
DarkSide666: correct

demos/collection/jssortable.php
DarkSide666: user theoretically can input anything in iso field :D

src/App.php
DarkSide666: review and approve and you can merge this in :)

mvorisek (approved):
ok

PR #1200, opened 2020-05-16
Add NS to demos

### LINKED ISSUE #1199: Add some NS to demos
Otherwise some absolute type names are without the leading slash.

And then probably mass import ui and data root namespaces. And rerun fixer.

mvorisek:
fix #1199

Needed even if there is no autoloading on these, CS fixer otherwise behave differently without NS (and PHP too, as leading `\` is not required otherwise for absolute names)


mvorisek:
@ibelar, hi, you were refactoring these week or two ago... Can you check the changes and resolve the class names with the current CS settings?

I think you there are only two options
- extend the classes anonymously like `$cl = new class() extends XXX {};` (so no need for actual/new class name)
- move them to a separate and equally names files

Then this PR is complete.

demos/collection/multitable.php
```diff
@@ -5,7 +5,7 @@
 require_once __DIR__ . '/../atk-init.php';
 
 // Re-usable component implementing counter
-class Finder extends \atk4\ui\Columns
+$finder = new class() extends \atk4\ui\Columns
```
mvorisek: Good idea, you can even wrap this inside get_class() :D (like get_class(new class ()...{})), otherwise the additional instance might be very confusing, addTo will always create a new instance

And name it more like `$finderClass`.

DarkSide666: addressed

DarkSide666 (approved):
@mvorisek please fix conflicts and then it's ready to merge.

DarkSide666:
Not everything was finished here, but let's merge this in and if needed create new PR with additional changes in demo scripts.

romaninsh:
@mvorisek - this did make demos harder to read. Their purpose was to teach others how to use features - but with anonymous classes it is a non-standard approach.

another consideration could be to assign dynamic classes into objects (drop get_class).

mvorisek:
@romaninsh this is needed for CS to work - the only option is to separate the classes with proper autoloading.

PR #1201, opened 2020-05-16
Convert callables to closures and move hook names to constants

mvorisek:
merge after https://github.com/atk4/data/pull/597

REVERT FIRST COMMIT BEFORE MERGE

only cleanup, only changes in `atk4/data` were required, compatible with current `atk4/core`

using more specific names, names should be never used directly (the HOOK_xxx constants should theoretically have any unique name without functional change)

Regex to locate old names:
```
['"](afterExecute|onStep|beforeExit|beforeRender|submit|displayError|displaySuccess|loadPOST|beforeRow|afterRow|getHTMLTags|aborted)['"]
```

New hook names:
```
'afterExecute' => \atk4\ui\ActionExecutor/Basic::HOOK_AFTER_EXECUTE

'onStep' => \atk4\ui\ActionExecutor/UserAction::HOOK_STEP

'beforeExit' => \atk4\ui\App::HOOK_BEFORE_EXIT
'beforeRender' => \atk4\ui\App::HOOK_BEFORE_RENDER

'submit' => \atk4\ui\Form::HOOK_SUBMIT
'displayError' => \atk4\ui\Form::HOOK_DISPLAY_ERROR
'displaySuccess' => \atk4\ui\Form::HOOK_DISPLAY_SUCCESS
'loadPOST' => \atk4\ui\Form::HOOK_LOAD_POST

'beforeRow' => \atk4\ui\Lister::HOOK_BEFORE_ROW
'afterRow' => \atk4\ui\Lister::HOOK_AFTER_ROW

'getHTMLTags' => \atk4\ui\TableColumn\Generic::HOOK_GET_HTML_TAGS

'aborted' => \atk4\ui\jsSSE::HOOK_ABORTED
```

### DIFF
--- a/demos/collection/table.php
+++ b/demos/collection/table.php
@@ -27,7 +27,7 @@ $table->addColumn('date');
 $table->addColumn('salary', new \atk4\ui\TableColumn\Money());
 $table->addColumn('logo_url', [new \atk4\ui\TableColumn\Image()], ['caption' => 'Our Logo']);
 
-$table->onHook('getHTMLTags', function ($table, atk4\data\Model $row) {
+$table->onHook(\atk4\ui\TableColumn\Generic::HOOK_GET_HTML_TAGS, function ($table, atk4\data\Model $row) {
     switch ($row->id) {
         case 1: $color = 'yellow';
 
--- a/demos/collection/table2.php
+++ b/demos/collection/table2.php
@@ -22,7 +22,7 @@ $table->template->appendHTML('SubHead', '<tr class="center aligned"><th colspan=
 $table->template->appendHTML('Body', '<tr class="center aligned"><td colspan=2>This is part of body, goes before other rows</td></tr>');
 
 // Hook can be used to display data before row. You can also inject and format extra rows.
-$table->onHook('beforeRow', function (atk4\ui\Table $table) {
+$table->onHook(\atk4\ui\Lister::HOOK_BEFORE_ROW, function (atk4\ui\Table $table) {
     if ($table->current_id === 2) {
         $table->template->appendHTML('Body', '<tr class="center aligned"><td colspan=2>This goes above row with ID=2 (' . $table->current_row->get('action') . ')</th></tr>');
     } elseif ($table->current_row->get('action') === 'Tax') {
--- a/docs/label.rst
+++ b/docs/label.rst
@@ -122,7 +122,7 @@ Added labels into Table
 You can even use label inside a table, but because table renders itself by repeating periodically, then
 the following code is needed::
 
-    $table->onHook('getHTMLTags', function ($table, Model $row) {
+    $table->onHook(\atk4\ui\TableColumn\Generic::HOOK_GET_HTML_TAGS, function ($table, Model $row) {
         if ($row->id == 1) {
             return [
                 'name'=> $table->app->getTag('div', ['class'=>'ui ribbon label'], $row->get('name')),
--- a/src/ActionExecutor/UserConfirmation.php
+++ b/src/ActionExecutor/UserConfirmation.php
@@ -238,7 +238,7 @@ class UserConfirmation extends Modal implements jsInterface_, Interface_
             $this->hide(),
             $this->ok->js(true)->off(),
             $this->cancel->js(true)->off(),
-            $this->hook('afterExecute', [$obj, $id]) ?: $success ?: new jsToast('Success' . (is_string($obj) ? (': ' . $obj) : '')),
+            $this->hook(Basic::HOOK_AFTER_EXECUTE, [$obj, $id]) ?: $success ?: new jsToast('Success' . (is_string($obj) ? (': ' . $obj) : '')),
         ];
     }
 
--- a/src/CRUD.php
+++ b/src/CRUD.php
@@ -158,7 +158,7 @@ class CRUD extends Grid
 
         if ($executor instanceof UserAction) {
             foreach ($this->onActions as $k => $onAction) {
-                $executor->onHook('onStep', function ($ex, $step, $form) use ($onAction, $action) {
+                $executor->onHook(UserAction::HOOK_STEP, function ($ex, $step, $form) use ($onAction, $action) {
                     $key = key($onAction);
                     if ($key === $action->short_name && $step === 'fields') {
                         return call_user_func($onAction[$key], $form, $ex);
--- a/src/CRUD.php
+++ b/src/CRUD.php
@@ -152,7 +152,7 @@ class CRUD extends Grid
     protected function initActionExecutor(Generic $action)
     {
         $executor = $this->getExecutor($action);
-        $executor->onHook('afterExecute', function ($ex, $return, $id) use ($action) {
+        $executor->onHook(ActionExecutor\Basic::HOOK_AFTER_EXECUTE, function ($ex, $return, $id) use ($action) {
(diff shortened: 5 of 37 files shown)


src/Form.php
```diff
@@ -244,10 +247,12 @@ public function setModel(Model $model, $fields = null)
 
     /**
      * Adds callback in submit hook.
+     *
+     * @return $this
      */
-    public function onSubmit(callable $callback)
+    public function onSubmit(\Closure $callback)
```
mvorisek, line 259: This is probably the only method where alias for `$this->onHook(self::HOOK_SUBMIT, $callback);` makes sense.

demos/collection/table.php
```diff
@@ -27,7 +27,7 @@
 $table->addColumn('salary', new \atk4\ui\TableColumn\Money());
 $table->addColumn('logo_url', [new \atk4\ui\TableColumn\Image()], ['caption' => 'Our Logo']);
 
-$table->onHook('getHTMLTags', function ($table, atk4\data\Model $row) {
+$table->onHook(\atk4\ui\TableColumn\Generic::HOOK_GET_HTML_TAGS, function ($table, atk4\data\Model $row) {
```
DarkSide666, line 30: Shouldn't this hook reside in Table class not TableColumn\Generic ?

demos/collection/table2.php
```diff
@@ -22,7 +22,7 @@
 $table->template->appendHTML('Body', '<tr class="center aligned"><td colspan=2>This is part of body, goes before other rows</td></tr>');
 
 // Hook can be used to display data before row. You can also inject and format extra rows.
-$table->onHook('beforeRow', function (atk4\ui\Table $table) {
+$table->onHook(\atk4\ui\Lister::HOOK_BEFORE_ROW, function (atk4\ui\Table $table) {
```
DarkSide666, line 25: This is confusing even if correct. Previously it was a lot easier to just use hook name, but now developer should also know which class to use for hook constant :(
Maybe could use `\atk4\ui\Table::HOOK_BEFORE_ROW` here instead? Can these constants be used in extended classes?

docs/label.rst
```diff
@@ -122,7 +122,7 @@ Added labels into Table
 You can even use label inside a table, but because table renders itself by repeating periodically, then
 the following code is needed::
 
-    $table->onHook('getHTMLTags', function ($table, Model $row) {
+    $table->onHook(\atk4\ui\TableColumn\Generic::HOOK_GET_HTML_TAGS, function ($table, Model $row) {
```
DarkSide666, line 125: Same here - isn't this hook for Table class?

src/ActionExecutor/UserConfirmation.php
```diff
@@ -238,7 +238,7 @@ protected function jsGetExecute($obj, $id): array
             $this->hide(),
             $this->ok->js(true)->off(),
             $this->cancel->js(true)->off(),
-            $this->hook('afterExecute', [$obj, $id]) ?: $success ?: new jsToast('Success' . (is_string($obj) ? (': ' . $obj) : '')),
+            $this->hook(Basic::HOOK_AFTER_EXECUTE, [$obj, $id]) ?: $success ?: new jsToast('Success' . (is_string($obj) ? (': ' . $obj) : '')),
```
DarkSide666, line 241: Can't we use something like `self::HOOK_AFTER_EXECUTE` or `static::HOOK_AFTER_EXECUTE` here?

src/CRUD.php
```diff
@@ -152,13 +152,13 @@ public function setModel(Model $m, $fields = null): Model
     protected function initActionExecutor(Generic $action)
     {
         $executor = $this->getExecutor($action);
-        $executor->onHook('afterExecute', function ($ex, $return, $id) use ($action) {
+        $executor->onHook(ActionExecutor\Basic::HOOK_AFTER_EXECUTE, function ($ex, $return, $id) use ($action) {
```
DarkSide666, line 155: Probably better to use `$executor->onHook($executor::HOOK_AFTER_EXECUTE, ...)` here.

src/CRUD.php
```diff
@@ -152,13 +152,13 @@ public function setModel(Model $m, $fields = null): Model
...
             return $this->jsExecute($return, $action);
         });
 
         if ($executor instanceof UserAction) {
             foreach ($this->onActions as $k => $onAction) {
-                $executor->onHook('onStep', function ($ex, $step, $form) use ($onAction, $action) {
+                $executor->onHook(UserAction::HOOK_STEP, function ($ex, $step, $form) use ($onAction, $action) {
```
DarkSide666, line 161: And here is need to know that hook step hook belongs to UserAction class and previous hook after execute hook belongs to ActionExecutor\Basic class... No, I do not like that.
Probably here it should better be `$executor->onHook($executor::HOOK_STEP, ...)` and I do not care in which of executor parent classes this constant is defined.

Same syntax probably could be used everywhere `$obj->onHook($obj::HOOK_NAME, ...)` instead of using exact class names.

src/Form.php
```diff
@@ -502,7 +513,7 @@ public function loadPOST()
     {
         $post = $_POST;
 
-        $this->hook('loadPOST', [&$post]);
+        $this->hook(self::HOOK_LOAD_POST, [&$post]);
```
DarkSide666, line 516: What if I extend this class? Shouldn't use `static` instead in this case and similar cases elsewhere?

src/Table.php
```diff
@@ -489,7 +489,7 @@ public function renderView()
         $this->_rendered_rows_count = 0;
         foreach ($this->model as $this->current_id => $tmp) {
             $this->current_row = $this->model->get();
-            if ($this->hook('beforeRow') === false) {
+            if ($this->hook(self::HOOK_BEFORE_ROW) === false) {
```
DarkSide666, line 492: see - here you use `self::HOOK_BEFORE_ROW`, but in other places (classes) used like
`$table->onHook(Lister::HOOK_BEFORE_ROW, ...)`

src/TableColumn/Generic.php
```diff
@@ -302,7 +307,7 @@ public function getHeaderCellHTML(Field $field = null, $value = null)
             throw new \atk4\ui\Exception(['How $table could not be set??', 'field' => $field, 'value' => $value]);
         }
 
-        if ($tags = $this->table->hook('getColumnHeaderCell', [$this, $field, $value])) {
+        if ($tags = $this->table->hook(self::HOOK_GET_HADER_CELL_HTML, [$this, $field, $value])) {
```
DarkSide666: typo error - HADER

src/TableColumn/Generic.php
```diff
@@ -21,6 +21,11 @@ class Generic
     use \atk4\core\TrackableTrait;
     use \atk4\core\DIContainerTrait;
 
+    /** @const string */
+    public const HOOK_GET_HTML_TAGS = self::class . '@getHTMLTags';
+    /** @const string not used, make it public if needed or drop it */
+    private const HOOK_GET_HADER_CELL_HTML = self::class . '@getHeaderCellHTML';
```
DarkSide666: typo error - HADER

DarkSide666 (changes requested):

DarkSide666 (approved):
OK, this will be a bit confusing at start to search for hook name in table column class, but actually call hook from Table class, but I guess it's correct. Lets go with that.

PR #1202, opened 2020-05-17
Use session persistence

DarkSide666:
Will use session persistence instead of session trait in model.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
     "php": ">=7.2.0",
     "ext-json": "*",
     "atk4/core": "dev-develop",
-    "atk4/data": "dev-develop",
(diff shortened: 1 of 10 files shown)


composer.json
```diff
@@ -36,7 +36,7 @@
     "php": ">=7.2.0",
     "ext-json": "*",
     "atk4/core": "dev-develop",
-    "atk4/data": "dev-develop",
+    "atk4/data": "dev-develop#2837b1029af4043b1cb8b30b7297ebfa0913d271",
```
romaninsh, line 39: ? what is this?

DarkSide666: Just for testing. Please review https://github.com/atk4/data/pull/598 first and then I can proceed here.

DarkSide666:
No need for this - @mvorisek provided simpler PR which does the trick.

PR #1204, opened 2020-05-18
Fix misc NS

### LINKED ISSUE #1062: Remove src/misc directory
See https://github.com/atk4/ui/tree/develop/src/misc

Can we refactor the `misc/ModelProxy` to `ModelProxy`?

Is the `proxy` word explaining good enought what this model does? See usages, used only on few placed when there is no `Model` given and checked only on one place if it can be saved - I am sure we can check this differently so maybe we can remove the `misc/ModelProxy` class completely.

mvorisek:
close #1062, it can not be removed easily

full BC as names are CI


DarkSide666 (approved):
good

PR #1207, opened 2020-05-19
Pass value with correct escaping, not embed

romaninsh:
URGENT - error appearing on https://ui.agiletoolkit.org/demos/collection/tablefilter.php right now:

Screenshot 2020-05-19 at 13 24 07

Looks like value is embedded into the expression. 

```
$m->addExpression('is_uk', 'if([iso] = "GB", 1, 0)')->type = 'boolean';
```

It should use a proper escaping.


romaninsh:
Testing is blocked by "Object should have NameTrait applied to use session" error...

DarkSide666 (approved):

PR #1208, opened 2020-05-19
Fix FilterModel/Generic with NameTrait removed from Model

mvorisek:
fix BC-break introduced by https://github.com/atk4/data/pull/596


romaninsh (approved):
LGTM

PR #1212, opened 2020-05-19
Fix current_row type mismatches

### LINKED ISSUE #1197: Lister does use undefined property "current_row"
https://github.com/atk4/ui/blob/3fd37f0bee0d43f01e9eccb41e96b9dc3466d5ca/src/Lister.php#L139

Either define it properly or do not use it at all.

### LINKED ISSUE #1209: Demos Table2.php issue
Table current_row is now an array : https://github.com/atk4/ui/blob/0652ab852a40a89f8a49a1806b51fba063854842/src/Table.php#L491

in demos/table2.php is used as a model
https://github.com/atk4/ui/blob/0652ab852a40a89f8a49a1806b51fba063854842/demos/collection/table2.php#L28

mvorisek:
fixes #1209
fixes #1197

### DIFF
--- a/src/Lister.php
+++ b/src/Lister.php
@@ -43,6 +43,9 @@ class Lister extends View
      */
     public $ipp;
 
+    /** @var Model */
+    public $current_row;
+
     /**
      * Initialization.
      */
@@ -141,7 +144,7 @@ class Lister extends View
 
         // Iterate data rows
         $this->_rendered_rows_count = 0;
-        foreach ($this->model as $this->current_id => $ignore) {
+        foreach ($this->model as $ignore) {
             $this->current_row = $this->model;
             if ($this->hook(self::HOOK_BEFORE_ROW) === false) {
                 continue;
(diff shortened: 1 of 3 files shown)


DarkSide666:
I think that it should be like this
```
foreach ($this->model as $junk) {
    $this->current_id = $this->model->id;
    ...
```
but need to check how bad this BC breaking change could be.

romaninsh:
I think there is a certain sense if current_row contains return of model iterator (e.g. see https://github.com/atk4/data/pull/591#issuecomment-641104171),

if we intend to drop array-access then why use array here? Isn't that inconsistent?

src/Lister.php
```diff
@@ -141,7 +144,8 @@ public function renderView()
 
         // Iterate data rows
         $this->_rendered_rows_count = 0;
-        foreach ($this->model as $this->current_id => $this->current_row) {
+        foreach ($this->model as $ignore) {
+            $this->current_row = $this->model;
```
DarkSide666, line 148: I guess you could just do `foreach ($this->model as $this->current_row)` no?

DarkSide666 (approved):

PR #1215, opened 2020-05-22
[fix] Crud using action modifier

ibelar:
- CRUD now uses the action modifier value to react on the grid.
- minor cs fix for MultiLine field


DarkSide666 (approved):
LGTM

PR #1216, opened 2020-05-25
[feature] Display form field using field name in template

ibelar:
Take full control of your form layout.
 - separate form template from inputs template for better control; 
 - allow using field name in template 
    
```
       <div class="ui attached segment">
          <div class="ui grid">
            <div class="row">
              <div class="eight wide column">
                {$from}
                {$to}
              </div>
              <div class="eight wide column">
                {$country}
                {$contains}
              </div>
            </div>
          </div>
        </div>
  ```

### Demo
- html-layout.php in Form section.

### Preview

Screen Shot 2020-05-25 at 11 44 48 AM


PhilippGrashoff:
@ibelar looks nice. Two questions I have:
1) This is a bit like Custom FormLayout, but better? Because in Custom Form layout, you would have to add labels manually?
2) I found it terribly hard to move the text from `ui => ['hint' => 'SomeHintForThatField']` from the input to the label of the field. Will this be easier with this change?

Cheers

ibelar:
> 1. This is a bit like Custom FormLayout, but better? Because in Custom Form layout, you would have to add labels manually?

Exactly; now you can simply drop the name of the field in the template and it will render it correctly. You do not need to add all fields as well, cause the Content tag will render the remaining field.

> I found it terribly hard to move the text from ui => ['hint' => 'SomeHintForThatField'] from the input to the label of the field. Will this be easier with this change?

Good point. It is easy to move by changing the input template, you can also pass another input template starting with this PR. But you made me realize that setting it might still be difficult. I will update PR for it.

DarkSide666 (approved):
Looks nice

PR #1217, opened 2020-05-30
[fix] Left menu visibility

ibelar:
Set visibility server-side instead of via javascript
- prevent menu flickering.


DarkSide666 (approved):
OK

PR #1218, opened 2020-06-01
Always convert large (>= 2^53) integers to string for JS

mvorisek:
Important as IDs are often integers and they must never be rounded.

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER


DarkSide666 (approved):
LGTM

PR #1219, opened 2020-06-02
WIP Close html tag immediatelly - add nested Form support

mvorisek:
replaced with https://github.com/atk4/ui/pull/1275

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -555,9 +555,9 @@ class App
      *
      * @return Template
      */
-    public function loadTemplate($name)
+    public function loadTemplate($name, /* DO NOT MERGE THIS CHANGE! */$cl = null)
     {
-        $template = new Template();
+        $template = $cl ?? new Template();
         $template->app = $this;
 
         if (in_array($name[0], ['.', '/', '\\'], true) || strpos($name, ':\\') !== false) {
--- a/src/Form.php
+++ b/src/Form.php
@@ -157,6 +157,28 @@ class Form extends View
      */
     public function init(): void
     {
+        // DEBUG FOR NESTED FORMS
+        // see https://github.com/atk4/ui/pull/1219
+        $this->template = $this->app->loadTemplate($this->defaultTemplate, new class() extends Template {
+            public $form;
+
+            protected function recursiveRender($template)
+            {
+                $out = parent::recursiveRender($template);
+
+                return $this->fixInnerHtml($out);
+            }
+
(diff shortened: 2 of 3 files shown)


ibelar:
@mvorisek - Not sure exactly what you are trying to achieve but here are my two cents on it.

- Nested form tag is not allowed in HTML and I some browser, Chrome at least, will not render the forms correctly when nested.
- Do you refer to having two separate properly enclosed form tag with their input outside of their forms? Something like this: https://www.impressivewebs.com/html5-form-attribute/
This would probably require big changes in how atk4/ui is handling form submission. Fomantic API request is handling form data serialization for us and also sending the request. Data serialization is handled automatically because of jQuery serialize plugin that gather all input type control in an enclosing form. Thus separating the input from inside its form tag might be challenging.

mvorisek:
@ibelar The usecase is to allow full AST/flexibility, ie. to allow to add subcomponents to main forms which uses their independed form.

> Nested form tag is not allowed in HTML and I some browser, Chrome at least, will not render the forms correctly when nested.

This is correct and thank you for finding this out.

> Fomantic API request is handling form data serialization for us and also sending the request. Data serialization is handled automatically because of jQuery serialize plugin that gather all input type control in an enclosing form. Thus separating the input from inside its form tag might be challenging.

Yes, I am aware of this + having to use possible different CSS selectors (but as long as the rules does not use `form` / `id >` / `id +` it should be full BC). I will try to implement this and let's see if it is easily possible or not.

Selecting owner form should be actually not that complicated as you can traverse thru the parent tags and instead of looking/stopping at `form`, you stop at `div + form`. `SUI` probably uses `jQuery`. I think it is possible to modify it on runtime to allow this change. If it will be possible and the demos will work without any issues, are you ok to give this PR a green light?

ibelar:
Of course. The challenge might come from Fomantic API request itself because they are gathering the form data with the request, i.e. it serializes the form using form tag. It may easily change but I do not know. 

One more thing is that conditional form js plugin would required refactoring but that should not be a big deal. 

As if this is a green light or not, we should ask @romaninsh. But if all does not break bc, I do not see any problem with it.

romaninsh:
Adding 🏷️ BC-BREAK because this may affect CSS styles around forms in custom projects.

I do not understand the benefit of this PR. We can consider for V3.

mvorisek:
@romaninsh simple: nested component design, very important!

PR #1220, opened 2020-06-02
Detect unwanted output before App::outputResponse()

mvorisek:
Until now it was detected only if the output was >= than `output_buffering` php.ini threshold and thus already detected by `headers_sent()` few lines below

no BC break for normal code


DarkSide666 (approved):
LGTM

PR #1221, opened 2020-06-02
Remove obsolete App::normalizeClassNameApp()

mvorisek:
Already obsolete, should be compatible with current core which tests if the method exists

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -486,14 +486,6 @@ class App
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mvorisek:
@DarkSide666 CS fixed already in #1220 , so it should be merged first...

DarkSide666 (approved):
ok

PR #1223, opened 2020-06-02
bypass cache for demo ui site build.

romaninsh:
This adds a new step into "atk-demo" build pipeline, which would specify a commit to be used by dependencies.

This helps us to avoid packagist cache and add "triggers" for dependency libraries too

Screenshot 2020-06-02 at 18 41 54

Screenshot_2020-06-02_at_18 17 09


tools/inject-dependency-sha.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+cat composer.json | \
+  jq '.require["atk4/core"]=.require["atk4/core"]+"#'`git ls-remote git@github.com:atk4/core.git develop | cut -f1`'"' |\
+  jq '.require["atk4/data"]=.require["atk4/data"]+"#'`git ls-remote git@github.com:atk4/data.git develop | cut -f1`'"' \
```
mvorisek: Add dsql as well

mvorisek (changes requested):

.codefresh/ui-build-deploy.yaml
```diff
@@ -10,6 +10,13 @@ steps:
     stage: clone
     repo: atk4/ui
     revision: "${{CF_BRANCH}}"
+
+  fix_sha:
+    stage: clone
+    title: Inject dependency SHA to bypass cache
+    commands:
+      - ./tools/inject-dependency-sha.sh
```
mvorisek: I think it is still better to do in DockerFile, as it can be run also locally.

mvorisek (changes requested):

tools/inject-dependency-sha.sh
romaninsh: dsql is not a direct dependency of atk4/ui, i'll try adding it though.

.codefresh/ui-build-deploy.yaml
romaninsh: if files are produced on a separate step, it allows us to cache docker layers more efficiently when dependencies haven't been updated.

romaninsh: just thought of another reason for this being in the pipeline...

We are likely to have 2 pipelines - one for stable builds (e.g. master) and another for unstable (develop).

tools/inject-dependency-sha.sh
romaninsh: actually - can we divert the 'dsql' dependency addition into 2.2 ?

DarkSide666 (approved):
LGTM

.codefresh/ui-build-deploy.yaml
mvorisek: Really? composer install can not be cached, so if this is after it...

PR #1229, opened 2020-06-02
fix/Popup demo

ibelar:
Fix Popup console error, callback and session retrieve value.
 - cause by anonymous class


DarkSide666:
This PR is another reason to hate these anonymous classes in demos :)
For now let's just fix them, but in future we should get rid of them and refactor demos to separate classes out in separate php files as requested.

DarkSide666 (approved):

PR #1230, opened 2020-06-02
fix/Recursive demo

ibelar:
fix recursive demo

UPDATE:

> the problem is coming from having an anonymous class name with jsReload or callback. The element id is base on class name but it seem to be causing a problem when anonymous class are used, plus when reload, name does not seem to be always the same. I was able to fix the other demo by adding name and id manually but it can not be done for this one. Since adding an object to self on reload will cause to re-add object in a loop.

### DIFF
--- a/demos/others/recursive.php
+++ b/demos/others/recursive.php
@@ -35,4 +35,4 @@ class recursive extends \atk4\ui\View
 }
 
 $view = \atk4\ui\View::addTo($app, ['ui' => 'segment']);
-$view->add(new Recursive());
+$view->add(new recursive());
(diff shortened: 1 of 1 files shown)


demos/others/recursive.php
```diff
@@ -19,22 +20,21 @@ public function init(): void
...
+                break;
         }
     }
-});
+}
 
 $view = \atk4\ui\View::addTo($app, ['ui' => 'segment']);
-
-$mySwitcherClass::addTo($view);
+$view->add(new Switcher());
```
DarkSide666: new recursive() here

DarkSide666 (changes requested):

demos/others/recursive.php
ibelar: yeah...

DarkSide666 (approved):

PR #1231, opened 2020-06-02
Disable exec (and simillar) functions in demo Docker config

mvorisek:
UPDATE: we should disable exec instead like (for whole cloud env):

image

### DIFF
--- a/demos/interactive/console.php
+++ b/demos/interactive/console.php
@@ -76,7 +76,7 @@ $t = $tt->addTab('exec() chain', function ($t) {
     });
 });
 
-if (false) { // THIS DEMO IS DISABLED FOR CLOUD USE
+if (empty($_ENV['CLOUD_DEMO'])) { // THIS DEMO IS DISABLED FOR CLOUD USE
     $t = $tt->addTab('composer update', function ($t) {
         \atk4\ui\Header::addTo($t, [
             'icon' => 'terminal',
--- a/demos/interactive/console.php
+++ b/demos/interactive/console.php
@@ -76,26 +76,25 @@ $t = $tt->addTab('exec() chain', function ($t) {
     });
 });
 
-/* THIS DEMO IS DISABLED FOR CLOUD USE
- *
-$t = $tt->addTab('composer update', function ($t) {
-    \atk4\ui\Header::addTo($t, [
-        'icon' => 'terminal',
-        'Command execution',
-        'subHeader' => 'it is easy to run server-side commands and stream output through console',
-    ]);
+if (false) { // THIS DEMO IS DISABLED FOR CLOUD USE
+    $t = $tt->addTab('composer update', function ($t) {
+        \atk4\ui\Header::addTo($t, [
+            'icon' => 'terminal',
+            'Command execution',
+            'subHeader' => 'it is easy to run server-side commands and stream output through console',
+        ]);
 
-    $w = \atk4\ui\Message::addTo($t, ['This demo may not work', 'warning']);
-    $w->text->addParagraph('This demo requires you to have "bash" and "composer" installed and may display error if the process running PHP does not have write access to the "vendor" folder and "composer.*".');
+        $w = \atk4\ui\Message::addTo($t, ['This demo may not work', 'warning']);
+        $w->text->addParagraph('This demo requires you to have "bash" and "composer" installed and may display error if the process running PHP does not have write access to the "vendor" folder and "composer.*".');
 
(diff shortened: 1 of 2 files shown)


mvorisek:
@ibelar how to test the original issue, or is the fix actually not needed?

demos/interactive/console.php
```diff
@@ -77,26 +77,25 @@ public function generateReport()
...
 
-/* THIS DEMO IS DISABLED FOR CLOUD USE
- *
-$t = $tt->addTab('composer update', function ($t) {
-    \atk4\ui\Header::addTo($t, [
-        'icon' => 'terminal',
-        'Command execution',
-        'subHeader' => 'it is easy to run server-side commands and stream output through console',
-    ]);
+if (false) { // THIS DEMO IS DISABLED FOR CLOUD USE
```
mvorisek: @romaninsh better to comment code like this way...

DarkSide666: How this is better?

mvorisek: see commits history, legit?

DarkSide666 (commented):
@ibelar changed this because something didn't work this way. Not sure what exactly - let's wait when he comes online (at our afternoon)

demos/interactive/console.php
DarkSide666: Still not sure how it's better because commenting out also goes in commits history, but ... ok, who cares :)

mvorisek:
> 
> 
> @ibelar changed this because something didn't work this way. Not sure what exactly - let's wait when he comes online (at our afternoon)

@DarkSide666 let him know to always be concrete in the PR desc, so there are no questions like this one needed

demos/interactive/console.php
mvorisek: It is important you understand it. If you refactor the code (CS, rename, ..., want to find usages, ...) this way it is still standard AST, comment is string, very, very bad!

romaninsh: Commits history works either way - but i think this would cause warnings with code reviewers about unreachable code. 

Perhaps we could use environment variable to detect and disable certain things in the cloud, i'll add "$_ENV['CLOUD_DEMO']="1", could you add it instead of "true" then?

romaninsh: https://github.com/atk4/infrastructure/blob/master/helm/atk-demo/templates/deployment.yaml#L45

mvorisek: > Commits history works either way

Same for you. If you refactor the code (CS, rename, ..., want to find usages, ...) this way it is still standard AST, comment is string, very, very bad! Where refactoring, you do not work with git history at all.

> Perhaps we could use environment variable to detect and disable certain things in the cloud, i'll add "$_ENV['CLOUD_DEMO']="1", could you add it instead of "true" then?

Why not, ok from my side :)

mvorisek: @romaninsh done, please check

Also note that "increase coverage by 0.45%". Another reason to not comment code :)

ibelar:
@mvorisek - the problem is coming from having an anonymous class name with jsReload or callback. The element id is base on class name but it seem to be causing a problem when anonymous class are used, plus when reload, name does not seem to be always the same. I was able to fix the other demo by adding name and id manually but it can not be done for this one. Since adding an object to self on reload will cause to re-add object in a loop.

We can refactor the demo differently of course but it will beat the purpose of this demo.

any idea how we can use standard class instead of anonymous?

mvorisek:
> the problem is coming from having an anonymous class name with jsReload or callback.

I can fix asap in core, we can use the parent class name if anonymous - can you describe how to reproduce on one demo?

> any idea how we can use standard class instead of anonymous?

only by moving the files to different location

demos/interactive/console.php
mvorisek: changed to much better and secure way using `disable_functions`

romaninsh:
use $ENV['CLOUD_DEMO'] to detect cloud:

https://github.com/atk4/infrastructure/commit/81692fd770944da423e530aa1c1caf58447d570c#diff-bc373cad86a055dc902dd6712bc8a9c4R45

romaninsh:
👍 for the Dockerfile change (disabling dangerous  functions).

romaninsh (approved):

georgehristov:
@mvorisek this broke the UI demo https://ui.agiletoolkit.org/
demo

mvorisek:
@georgehristov already proposed a fix yesterday https://github.com/atk4/ui/pull/1272

PR #1232, opened 2020-06-03
ViewTester fixes

DarkSide666:
* fix interactive / loader
* fix interactive / modal


mvorisek:
~I do not care about Docker, I mean ok with me, but~ do we really need to include tests for 1, 2 demos? Seems like bad practice too me so I advise to not do so.

DarkSide666:
Yeah I was thinking about moving that class to demos/ViewTester, because it wasn't used anywhere else. But then again - maybe some other repos use it somewhere...
OK, who cares - you're right.

mvorisek:
I mean to not include any tests from demos. Tests should be possible to delete without loosing any functionality. These are unwritten rules for good design. (I see my comment was not very clear, I updated it now)

DarkSide666:
Check out now.

demos/ViewTester.php
```diff
@@ -1,6 +1,6 @@
 <?php
```
mvorisek: Move probably to `demos/_includes/`

mvorisek:
Ok to merge

PR #1234, opened 2020-06-03
Delete CHANGELOG.md and build/add it to releases

mvorisek:
If ok, remove also in other repos. Currently, having 2 years old CHANGELOG.md with changes for old versions only does not make any sense.

UPDATE from Discord communication:
@romaninsh will be good if you can take care of the release process:
- forget about legacy/*, master, ... branches and add 2.minor.x braches instead - like in develop, releaes should be simple done from their HEAD + protect branches matching pattern "\.\d+\.x(-\w+)?" as well as all tags (yes, all, no one should push any tag)
- releases itself should be done like this: checkout head commit, update composer.json + save release log to CHANGELOG.md ( see https://github.com/atk4/ui/pull/1234 ) + run tests + commit "Release xxx" + add/push tag - no PR for releases etc.
- fix failing "atk4/ui.agiletoolkit.org" in atk4/core (or remove it from core status at all, it is very unrelated)


DarkSide666:
I agree with @mvorisek about this. Such old changelog is useless anyway.

romaninsh:
😭 it had a sentimental value. It is duplicating https://github.com/atk4/ui/releases?after=1.0.0 however, so i'll merge.

romaninsh (approved):
😭 ok.

mvorisek:
@romaninsh please more tears for other repos

PR #1237, opened 2020-06-03
Refactor ::addToWithClassName() to ::addToWithCl()

mvorisek:
merge after: https://github.com/atk4/core/pull/217

I do not like shortened names, but this one will be very frequently used later - I think `fromSeedWithCl` is better than `fromSeed2`

### DIFF
--- a/src/ActionExecutor/Basic.php
+++ b/src/ActionExecutor/Basic.php
@@ -125,7 +125,7 @@ class Basic extends \atk4\ui\View implements Interface_
 
         $this->addHeader();
 
-        \atk4\ui\Button::addToWithClassName($this, $this->executorButton)->on('click', function () {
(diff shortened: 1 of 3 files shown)


DarkSide666 (approved):
All tests fail, but this PR as such looks fine.

romaninsh:
This fix have made it into 2.1 branch of UI but relies on the  2.2 functionality in ATK Core. I have reverted it in 2.1 branch to release 2.1.1. 

In the future we should agree that no PRs are merged that can cause problems like this while releases are created across all repositories.

Ideally we would like to have all releases on the same day, but if that is difficult, I'd like to ask all maintainers to WAIT on merging any suspicious PRs into "develop" of atk4/ui.

PR #1238, opened 2020-06-04
Normalize View->add() to always require a seed class like ->factory() already does

mvorisek:
Historically `View->add()` did not required a seed class but factory and other new methods like `addTo()` or `fromSeed` does - drop this old behaviour for consistency and prevent unwanted errors/confusions.

BC breaking of course, but ui is already fully compatible due past refactoring :)


DarkSide666 (approved):
ok

PR #1239, opened 2020-06-04
Revert fix demo popup - not needed after atk4/core#220

ibelar:
It can be revert. Fix in core with https://github.com/atk4/core/pull/220


mvorisek:
@ibelar also RTM

DarkSide666 (approved):
looks fine

PR #1242, opened 2020-06-05
Refactor exception constructor calls to not use array

mvorisek:
refactor for: https://github.com/atk4/core/pull/223

this PR itself is pure refacotirng without any functionality change thus 100% BC


src/FormField/Generic.php
```diff
@@ -78,7 +79,7 @@ public function init(): void
 
         if ($this->form && $this->field) {
             if (isset($this->form->fields[$this->field->short_name])) {
-                throw (new \atk4\ui\Exception('Form already has a field with the same name'))
+                throw (new Exception('Form already has a field with the same name'))
```
mvorisek, line 82: This PR does not target these changes... But ok.

src/TableColumn/Generic.php
```diff
@@ -303,7 +303,7 @@ public function getTag($position, $value, $attr = [])
     public function getHeaderCellHTML(Field $field = null, $value = null)
     {
         if (!$this->table) {
-            throw (new \atk4\ui\Exception('How $table could not be set??'))
+            throw (new Exception('How $table could not be set??'))
```
mvorisek, line 306: why was this not fixed by CS fixer?

src/FormField/Generic.php
DarkSide666: I just fix where i see these if file is already open :)

DarkSide666 (approved):

PR #1243, opened 2020-06-05
Fix fixed names in tests

mvorisek:
Proper fix for:
image

related with https://github.com/atk4/core/pull/225


mvorisek:
replaced by https://github.com/atk4/ui/pull/1245

PR #1251, opened 2020-06-07
Fix missing DIContainerTrait when StaticAddToTrait is used

mvorisek:
Refactor code for https://github.com/atk4/core/pull/221

fully documented behaviour that DIContainerTrait trait much be used when StaticAddToTrait is used


romaninsh (approved):
LGTM

PR #1252, opened 2020-06-08
Fix usage of hasXXX(), it now return bool only

mvorisek:
in future, `hasField` might be introduced back but to return strictly bool value as expected, but it is no BC break now and will not be later - `tryGetField` will remain

### DIFF
--- a/src/FormLayout/_Abstract.php
+++ b/src/FormLayout/_Abstract.php
@@ -20,18 +20,13 @@ abstract class _Abstract extends \atk4\ui\View
      * Places field inside a layout somewhere. Should be called
      * through $form->addField().
      *
-     * @param string                   $name
      * @param array|string|object|null $decorator
      * @param array|string|object|null $field
      *
      * @return \atk4\ui\FormField\Generic
      */
-    public function addField($name, $decorator = null, $field = null)
+    public function addField(string $name, $decorator = null, $field = null)
     {
-        if (!is_string($name)) {
-            throw new Exception('Format for addField now require first argument to be name');
-        }
-
         if (!$this->form->model) {
             $this->form->model = new \atk4\ui\Misc\ProxyModel();
         }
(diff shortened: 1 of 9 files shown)


georgehristov (commented):
`$row->get($field)` should also work even if `$field` is a `\atk4\data\Field`. Any reason to change this?

mvorisek:
@georgehristov `$row->get($field)` is was unified to accept only string field name at least for now - only a few uses, I make these decisions not only on "what is expected", but also on data, ie. "how this is usually used"

georgehristov:
I really don't like the direction of stripping down the versatility...

mvorisek:
> 
> 
> I really don't like the direction of stripping down the versatility...

I may seem so - but compare how many uses are "correct/canonical" and how many are like to be fixed - it is more than 100:1. Without the data, I would say probably the say the same as you...

Let's get this merged.

If you prefer array access like model access, getting values by field objects, you made extend the data Model and use it within your projects as you like.

But in lib, we need more standardized types and workflows, so everyone can understand it and the hidden bugs are discovered (for ex. mixed types of "current_row" and more than 10 others issues).

src/Table.php
```diff
@@ -194,29 +194,27 @@ public function addColumn($name, $columnDecorator = null, $field = null)
...
-        } elseif (is_object($field)) {
-            throw (new Exception('Duplicate field'))
-                ->addMoreInfo('name', $name);
         } else {
-            $field = $existingField;
+            $existingField = $this->model->getField($name);
+
+            if (is_array($field)) {
+                // Add properties to existing field
+                $existingField->setDefaults($field);
```
georgehristov: How about
```
$field = $existingField->setDefaults($field);
```

src/FormLayout/_Abstract.php
```diff
@@ -40,27 +35,27 @@ public function addField($name, $decorator = null, $field = null)
...
-                throw (new Exception('Duplicate field'))
-                    ->addMoreInfo('name', $name);
             } else {
-                $field = $existingField;
+                $existingField = $this->form->model->getField($name);
+
+                if (is_array($field)) {
+                    // Add properties to existing field
+                    $existingField->setDefaults($field);
+                    $field = $existingField;
```
georgehristov: How about

```
$field = $existingField->setDefaults($field);
```

src/FormLayout/_Abstract.php
```diff
@@ -40,27 +35,27 @@ public function addField($name, $decorator = null, $field = null)
...
-        if ($name) {
-            $existingField = $this->form->model->hasField($name);
-        }
-
         try {
-            if (!$existingField) {
+            if (!$this->form->model->hasField($name)) {
                 // Add missing field
                 if ($field) {
                     $field = $this->form->model->addField($name, $field);
```
georgehristov: If you are anyway modifying maybe use one-liner here?

```
$field = $this->form->model->addField($name, $field ?: []);
```

georgehristov (changes requested):

src/FormLayout/_Abstract.php
mvorisek: This was not intended to be fixed by this PR... But I will fix this now.

src/Table.php
mvorisek: This was not intended to be fixed by this PR... But I simplified it...

src/FormLayout/_Abstract.php
mvorisek: Done - tests are passing even without `?: []` - this needs to be decised later if seed should support `null` or not - I already created an issue for it - https://github.com/atk4/core/issues/226 - but passing null/string seems to be used a lot - todo for later.

If array for seed will be strictly required, than `?: []` will still be not the best one here - we should probably change the method default parameter value...

Ok otherwise, can you approve?

georgehristov (approved):

PR #1260, opened 2020-06-09
Always convert large (>= 2^53) integers to string in JSON output

### LINKED ISSUE #1258: CRITICAL: Lookup integer values are rounded again
fix https://github.com/atk4/ui/pull/1218 is proper

but it does not fix the manual cast needed in `Lookup` ( removed in that PR - https://github.com/atk4/ui/pull/1218/files#diff-d602f517388e9f70d88089beccf908e4L225  )

TODO: investigate why the global JS expr. casting rules did not applied and fix - do not add manual cast - it should be not needed!

mvorisek:
fixes https://github.com/atk4/ui/issues/1258 , issue reintroduced in https://github.com/atk4/ui/issues/1258

Very hackery, but intended as `App::encodeJson()` can be called with array with JsonSerializable objects which can produce JS functions (in theory anything) which is no longer valid JSON to reparse.

This approach should safe for JSON/JS strings (ie. nothing is converted in them).

BC breaking changes:
- (>= 2^53) integers in JSON output are always converted to strings to be safe in JS
- method `App::isJsRequest()` was renamed to `App::isJsUrlRequest()`


DarkSide666 (approved):
I'm not good at such long regexes, but I think it's good :)

PR #1264, opened 2020-06-09
Improve Fix/layout switching #1263

mvorisek:
PR good, but class paths must be always refered with ::class and never manupulated like strings/concated/prefixed

no functional change, leaving the URL input manually prefixed

### DIFF
--- a/demos/atk-init.php
+++ b/demos/atk-init.php
@@ -45,8 +45,7 @@ if (file_exists(__DIR__ . '/../public/atkjs-ui.min.js')) {
 }
 
 // allow custom layout override
-$layoutClass = $app->stickyGET('layout') ? 'atk4\ui\Layout\\' . $app->stickyGET('layout') : \atk4\ui\Layout\Maestro::class;
-$app->initLayout(\atk4\ui\Layout\Generic::factory([$layoutClass]));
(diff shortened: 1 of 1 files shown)


demos/atk-init.php
```diff
@@ -44,9 +44,9 @@
     $app->cdn['atk'] = $rootUrl . 'public';
 }
 
-// enable layout change.
-$layout = '\\atk4\\ui\Layout\\' . ($app->stickyGET('layout') ?: 'Maestro');
-$app->initLayout(new $layout());
+// allow custom layout override
+$layoutClass = $app->stickyGET('layout') ? 'atk4\ui\Layout\\' . $app->stickyGET('layout') : \atk4\ui\Layout\Maestro::class;
+$app->initLayout(\atk4\ui\Layout\Generic::factory([$layoutClass]));
```
mvorisek: @ibelar is different class than a subtype of `\atk4\ui\Layout\Generic` passed in Behat tests?

Behat tests are not passing with
https://github.com/atk4/ui/blob/9b7801dc908491a10b02ce638a9c602f0f5b14ec/demos/atk-init.php#L49

ibelar (approved):
Good!

PR #1265, opened 2020-06-10
Delete .github_changelog_generator

mvorisek:
related with https://github.com/atk4/ui/pull/1234@romaninsh or is it still needed? (it is not in other repos)


DarkSide666 (approved):
Probably not.

PR #1270, opened 2020-06-11
Disable action support

DarkSide666:
In legacy branches we didn't have user actions support in model and without that it's not possible to easily enable/disable table action buttons by using callback (for each row).
This solution here is slightly similar with the one we now have in up-to-date UI (by using data user actions enabled property as callback).


ibelar (approved):
LGTM

PR #1273, opened 2020-06-12
Do not use deprecated properties to check trait presence

mvorisek:
related with https://github.com/atk4/core/pull/227

- DIContainer is highly expected on added objects
- Sesstion trait check was redundand (checked only on `$this`, but a few lines above the trait was added to that class)


georgehristov (approved):
LGTM

PR #1275, opened 2020-06-13
Close html tag immediatelly - add nested Form support

mvorisek:
Nesting html `<form>` tags is not possible, hovewer they can be closed immediatelly and linked with the input controls/tags, see https://stackoverflow.com/questions/3430214/form-inside-a-form-is-that-alright#21900324

Atk should support AST as much as possible, this is also how some React components cope with this html limitation.

forms spec: https://www.w3.org/TR/html52/sec-forms.html


src/Form.php
```diff
@@ -626,14 +658,14 @@ public function ajaxSubmit()
         });
 
         //var_dump($cb->getURL());
-        $this->js(true)
+        $this->js(true, null, $this->formElement)
```
mvorisek, line 661: @ibelar can you please help with this?

What I have done:
- replaced the owner <form> tag with div
- added hidden/immediatelly closed <form> tag and linked the controls to it (with form attribute)

- now load /demos/form/form.php and compare the rendered source from this PR and develop -> this seems to be good

- I tried to replace the JS selectors by setting the 3rd param to be `$this->formElement`

but it seems to not work, the the content is not reloaded properly (compare vs. expected result on the demo above on develop)

it seems like some atk issue. JQuery serialization seems to be working well with the linked (ie. not in subtree of <form>) controls...

src/App.php
```diff
@@ -554,9 +554,9 @@ public function init(): void
      *
      * @return Template
      */
-    public function loadTemplate($name)
+    public function loadTemplate($name, /* DO NOT MERGE THIS CHANGE! */$cl = null)
```
mvorisek: FYI: for quick `form="form id"` render, ignore for now, it works...

src/Form.php
```diff
@@ -159,8 +166,33 @@ public function __construct($defaults = [])
      */
     public function init(): void
     {
+        // DEBUG FOR NESTED FORMS
+        // see https://github.com/atk4/ui/pull/1219
+        $this->template = $this->app->loadTemplate($this->defaultTemplate, new class() extends Template {
```
mvorisek: for quick `form="form id"` render, ignore for now, it works...

template/semantic-ui/form.html
```diff
@@ -1,3 +1,4 @@
-<form id="{$_id}" class="{_ui}ui{/} {$class} {_class}form{/}" style="{$style}" {$attributes}>
+<div id="{$_id}" class="{_ui}ui{/} {$class} {_class}form{/}" style="{$style}" {$attributes}>
```
mvorisek, line 1: FYI: this should be final version

src/Form.php
```diff
@@ -48,6 +48,13 @@ class Form extends View
      */
     public $canLeave = true;
 
+    /**
+     * HTML <form> element, all inner form controls/elements are linked to it.
+     *
+     * @var View
+     */
+    public $formElement;
```
mvorisek, line 57: good name? Or `htmlForm`?

mvorisek:
@ibelar, @georgehristov, can you please help with this?

What I have done:
- replaced the owner \<form> tag with div
- added hidden/immediatelly closed <form> tag and linked the controls to it (with form attribute)

- now load /demos/form/form.php and compare the rendered source from this PR and develop -> this seems to be good

- I tried to replace the JS selectors by setting the 3rd param to be `$this->formElement`

but it seems to not work, the the content is not reloaded properly (compare vs. expected result on the demo above on develop)

it seems like some atk issue. JQuery serialization seems to be working well with the linked (ie. not in subtree of \<form>) controls...

ibelar:
@mvorisek - I have taken a quick look at this. 

First: The HTML element set to fire the Fomantic api request is usually set to the form id. So form submit correctly but it is replacing content of the form, which in this PR has a display: none css property.

You would need to adjust this line Form::class:
```
 $this->js(true)
            ->api(array_merge(['url' => $cb->getJSURL(), 'method' => 'POST', 'serializeForm' => true], $this->apiConfig))
            ->form(array_merge(['inline' => true, 'on' => 'blur'], $this->formConfig));
```

Probably replacing with the id selector where you want to replace content.
```
new jQuery('#my_id')->js(true)->api()...
```

The other problem and I think this would be a major one, is the way Fomantic-UI is handling its form module. Especially when you need to display an error on a specific field. They have a namespace/selector mapping that you can set, but I never try and do not know if it will work. See https://fomantic-ui.com/behaviors/form.html#dom-settings

I am afraid that even if you manage to adjust field selector, their search for input is limited to the enclosed form tag, i.e. search is done inside form, not outside.

I have test manually in browser console and Fomantic throws error:
````
$('#atk_layout_maestro_tabs_tabssubview_form_2').form('add prompt', 'name', 'error');
````

**Comment by mvorisek** [comment]
> ```
> $this->js(true)
>         ->api(array_merge(['url' => $cb->getJSURL(), 'method' => 'POST', 'serializeForm' => true], $this->apiConfig))
>         ->form(array_merge(['inline' => true, 'on' => 'blur'], $this->formConfig));
> ```

This was the issue - this means two things that were written together (not horibly bad, but I did not realized it, thanks)

> The other problem and I think this would be a major one, is the way Fomantic-UI is handling its form module. Especially when you need to display an error on a specific field. They have a namespace/selector mapping that you can set, but I never try and do not know if it will work. See https://fomantic-ui.com/behaviors/form.html#dom-settings

> I am afraid that even if you manage to adjust field selector, their search for input is limited to the enclosed form tag, i.e. search is done inside form, not outside.

It seems `.form()` on owning div is working great. It seems they already count with input controls not wrapped in \<form>.

So current situation is:

- form.php demo works for me now without any issue
- CRUD works as well
- all submit/reset calls like `$form->js()->submit()` needs to be updated to `$form->js()->submit(null, null, $form->formElement)`
  - submit/reset are only the methods we need to care https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement#Methods
  - do we want to add there method to the owning div like https://stackoverflow.com/questions/4670361/how-to-add-my-own-methods-to-htmlelement-object ?
- https://github.com/atk4/ui/pull/1275/commits/2b1bbc7b545d32eccc01f5d04b29f3d012961302#diff-2c77116739ece89c91385095c838f093R670 is wrong, right?

@ibelar wdyt?

File: src/Form.php [hunk 6]
```diff
@@ -159,8 +166,35 @@ public function __construct($defaults = [])
...
+                {
+                    $out = parent::recursiveRender($template);
+
+                    return $this->fixInnerHtml($out);
+                }
+
+                private function fixInnerHtml(string $str): string
+                {
+                    $innerFormTags = ['button', 'datalist', 'fieldset', 'input', 'keygen', 'label', 'legend', 'meter', 'optgroup', 'option', 'output', 'progress', 'select', 'textarea'];
+                    $str = preg_replace('~<(' . implode('|', $innerFormTags) . ')(?!\w| form=")~i', '$0 form="' . $this->form->formElement->name . '"', $str);
```
mvorisek: Is replace like this ok for production? It does not replace anything if `form` attribute is presented, but it handles every case, even if no `Form\Field` element was used.

Nested forms should be supported as well.

src/Form.php
```diff
@@ -191,8 +225,8 @@ protected function initLayout()
         if ($this->buttonSave) {
             $this->buttonSave = $this->layout->addButton($this->buttonSave);
             $this->buttonSave->setAttr('tabindex', 0);
-            $this->buttonSave->on('click', $this->js()->form('submit'));
-            $this->buttonSave->on('keypress', new jsExpression('if (event.keyCode === 13){$([name]).form("submit");}', ['name' => '#' . $this->name]));
+            $this->buttonSave->on('click', $this->js(null, null, $this->formElement)->form('submit'));
+            $this->buttonSave->on('keypress', new jsExpression('if (event.keyCode === 13){$([name]).form("submit");}', ['name' => '#' . $this->formElement->name]));
```
ibelar: It looks like pressing 'Enter' is not submitting the form anymore.

template/semantic-ui/form.html
ibelar: Loader is not showing anymore when form is submitting. May by you should set class to 'basic segment' for loader to show if Fomantic api is called using this element. Or added StateContext in apiConfig.

src/Form.php
ibelar: I am not the best with name but formElement sounds right. :)

src/Form.php
mvorisek: I tested "enter", but only with focus on submit button.

mvorisek:
@ibelar I remind myself with this PR, can you look for the JS issues, so we can integrate quickly?

ibelar:
@mvorisek - I will but I won’t be able look at this week. Next week most probably.

template/semantic-ui/form.html
mvorisek: @ibelar issue is related with "loading" class set to \<form>, see Discord for more

src/Form.php
```diff
@@ -728,7 +728,7 @@ public function ajaxSubmit()
         $this->on('change', 'input, textarea, select', $this->js()->form('remove prompt', new JsExpression('$(this).attr("name")')));
 
         if (!$this->canLeave) {
-            $this->js(true, (new JsChain('atk.formService'))->preventFormLeave($this->name));
+            $this->js(true, (new JsChain('atk.formService'))->preventFormLeave($this->name), $this->formElement);
```
mvorisek: @ibelar please also check this selector/commit https://github.com/atk4/ui/pull/1275/commits/cd33102170cd0f452aa77a9ad81183e58cfa5dee

ibelar (approved):
LGTM

PR #1277, opened 2020-06-13
Fast testing without webserver

### LINKED ISSUE #1069: Each phpunit test takes about 600 ms
There are about 370 tests and phpunit takes 220 seconds, which equals to about 600 ms per test.

Analyse why the tests are so slow. Is that due webserver based on built-in php webserver?

Each basic atk4 ui response should not take longer than 60 ms.

### LINKED ISSUE #1278: Output all data strictly thru App in demos
image

mvorisek:
fix https://github.com/atk4/ui/issues/1069
fix https://github.com/atk4/ui/issues/1278

- intended for quick local testing - internal webserver is slow, even more slower on Windows
- also important to be able to collect performance data from one phpunit/process which tests everything


tests/BuiltInWebServerAbstract.php
```diff
@@ -109,9 +118,87 @@ private static function getPackagePath($directory = null, $_ = null): string
...
     private function getClient(): Client
     {
+        if (static::isWithWebserverTest()) {
+            $handler = null;
+        } else {
+            $handler = function (\Psr\Http\Message\RequestInterface $request, array $options) {
+                try {
+                    // backup globals
+                    $bak = [];
+                    foreach ([
```
abbadon1334: it will be not bad to add `laminas/laminas-diactoros` and set it as an injectable property of the App, and remove all direct access to superglobal, it can be a first step in direction of PSR-7.
```php
ServerRequestFactory::fromGlobals(
            $_SERVER,
            $_GET,
            $_POST,
            $_COOKIE,
            $_FILES
        );
```

mvorisek:
98% done, test time on Windows reduced from 180 seconds to 11 seconds.

Windows is slow because of php has quite slow file access on Windows. I already have a patch prepared for Atk to cache templates in memory which reduces the test (actually app tun time) time from 11 to 7 seconds.

With single process tests more performance data can be collected and badly coded hot-spots can be fixed.

abbadon1334 (changes requested):
i want to see if coverage is respected with that change on exit/noexit of demos.
The rest is ok after solving 7.2 tests

abbadon1334 (approved):
LGTM
i still think that including `laminas\laminas-diactoros` in ui can simplify many things on how we manage superglobals.

Why i don't see the ghaphical coverage comparison/report ?

mvorisek:
> i still think that including `laminas\laminas-diactoros` in ui can simplify many things on how we manage superglobals.

phpunit can handled restore, but we might not want to do so, as we should emulate browser more than focus on native cli globals

we should not also set too many global variables as with bad setups a lot of them is not also set

> Why i don't see the ghaphical coverage comparison/report ?

because you look at wrong report - for fast results it it only in `latest / pull request` (other jobs are run without xdebug/coverage which is much faster)

image

PR #1279, opened 2020-06-14
WIP: Replace json_decode with App::decodeJson() and json_encode with App::encodeJson()

mvorisek:
based on discussion `App` should be expected to be always presented

need more work...


mvorisek:
implemented in https://github.com/atk4/ui/pull/1466

PR #1282, opened 2020-06-14
Add type validation to factory calls

mvorisek:


src/Table.php
```diff
@@ -216,7 +216,7 @@ public function addColumn(?string $name, $columnDecorator = null, $field = null)
 
         if ($field === null) {
             // column is not associated with any model field
-            $columnDecorator = $this->_addUnchecked($this->factory($columnDecorator, ['table' => $this]));
+            $columnDecorator = $this->_addUnchecked(Table\Column::fromSeed($columnDecorator, ['table' => $this]));
```
mvorisek, line 219: FYI: `_addUnchecked` already checks type, but this is more like side effect because add does not work and it can change in the future...

ibelar (approved):
LGTM

PR #1286, opened 2020-06-15
Fix Docker/web without dev deps

mvorisek:
issue introduced in https://github.com/atk4/ui/pull/1284/files#diff-6f464e6eb275052ab5ae7e046c6a8514R10 as it required/tested `dev` file to be present


georgehristov (approved):
LGTM

PR #1291, opened 2020-06-15
Fix/demo intro

### LINKED ISSUE #1290: Demo fails on Array_ persistence creation
Is this issue due to changes in Array persistence?

demo

georgehristov:


demos/tutorial/intro.php
```diff
@@ -137,10 +137,13 @@ function init(): void {
 */
 
 session_start();
+
+$_SESSION['x'] = $_SESSION['x'] ?? [];
+
 $session = new \atk4\data\Persistence\Array_($_SESSION['x']);
```
mvorisek: better `?? []` here.

demos/tutorial/intro.php
```diff
@@ -137,10 +137,13 @@ function init(): void {
...
 
 $form = \atk4\ui\Form::addTo($app);
-$form->setModel(new DemoInvoice($session))
+$form->setModel(new \atk4\ui\demo\DemoInvoice($session))
```
mvorisek: should be not needed

georgehristov: It doesn't work without, most probably because code is eval'ed

mvorisek: I see, eval code without namespace, absolute path is probably better... Please check all other eval code.

mvorisek (approved):
Seems good, I added save hook also for the 2nd Model... Revert if not needed.

PR #1294, opened 2020-06-16
Transform SQL demo data to Sqlite and use Sqlite as default for demos

### LINKED ISSUE #1175: DEMO to use SQLite
ATK's support of sqlite has improved. It is also bundled in PHP.

It would be great to run ATK demo with SQLite by default. SQLite file should also be created by demos/Dockerfile.

This would eliminate need for DSN environment variable and we could run demo just with:

```
docker run -p 8080:80 -t atk4/ui-demo:develop
```

mvorisek:
fix https://github.com/atk4/ui/issues/1175


mvorisek:
@ibelar can you fix these tests:
```
    features/basicexecutor.feature:4
    features/basicexecutor.feature:10
    features/basicexecutor.feature:20
```
to make them working with pure Sqlite?

demos/db.default.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace atk4\ui\demo;
+
+// to use MySQL database:
+//   1. copy this file to "db.php"
+//   2. uncomment the line below (and update the configuration if needed)
+//   3. remove the Sqlite code from the new file
+// $db = new \atk4\data\Persistence\SQL('mysql:dbname=atk4_test__ui;host=mysql', 'atk4_test', 'atk4_pass');
+
+$sqliteFile = __DIR__ . '/_demo-data/db.sqlite';
```
romaninsh, line 13: Can you confirm that the SQL files well remain read-only in "atk-demo" ? We might be running multiple service pods without session affinity.

romaninsh (approved):

demos/db.default.php
mvorisek: I can confirm that web demo will continue using the DSN env. var, see https://github.com/atk4/ui/pull/1294/files#diff-cdc5998b685321f6451f8c451fcfee93R42

ibelar:
@mvorisek - Behat test only runs the demo file. There is probably something wrong with file database so import is not working. Did you try running it manually?

mvorisek:
> 
> 
> @mvorisek - Behat test only runs the demo file. There is probably something wrong with file database so import is not working. Did you try running it manually?

I run in only thru Travis. DB/Sqlite is otherwise working, I already fixed there two issues:
- text columns were initially imported to Sqlite with case sensitive collation
- `IF` expr is not supported in Sqlite

so I can confirm DB is working there for 100%, but I do not know how to fix the tree failing test above.

Can you check that locally (remove/rename your db.php if you have it on your side to enforce the Sqlite use) and the fix the Travis?

demos/db.default.php
romaninsh: Oh ok, so this only affect local demo. Thanks!

ibelar:
I am not using sql/lite so I cannot test it my self. But for sure, there is something going on when Import action is fired in File Model.

mvorisek:
@ibelar you can test is actually very easily and after this PR it will be default for everyone :)

To use Sqlite locally do the following:
- checkout this PR
- delete (or temporary rename to something else) `db.php` (only if you use custom `db.php` config)
- run `php demos/_demo-data/create-sqlite-db.php`
- now run demos/tests - the DB should be Sqlite (you can test it by deleting `demos/_demo-data/db.sqlite`)

romaninsh:
I tried few things but have reverted my changes.

PR #1299, opened 2020-06-16
[fix] File Import Demo

ibelar:
Remove record saving ability in File imported action.
see: https://ui.agiletoolkit.org/demos/collection/actions.php


mvorisek (changes requested):
image

DarkSide666 (approved):
it would be better to allow to save this in DB, but as hot-fix this is fine for now.

PR #1301, opened 2020-06-18
Cache template files in memory

mvorisek:
This change cuts the testing time from 8.6 seconds to 1.2 seconds.

But it is not related with tests, it actually improves the total load time!

In reality, the effect is lower:
- as render tree is normally not that large as all demos rendered at once
- tested on Windows, php on linux has much lower FS overhead

But this PR is still great if it can cut the average load time to half with a few lines of code :D

Now we can also think about implementing rendering of tables etc. with full native templating power.

Effect can be measured by a few page refreshes, example on `/demos/collection/crud.php`:

image


ibelar (approved):
Nice!

PR #1303, opened 2020-06-18
Add phpdoc to anonymous classes in demos

### LINKED ISSUE #1236: Remove get_class(new class(... from demos
Handy solution compatible with CS fixer, but not refactorable, so the classes should be put in separate files.

I think it is also possible to add one class before which will be renamed to the filename and another class names will remain. Extremely ugly, but it is a solution for CS fixer.

Different settings of CS fixer may be possible, but I do not know (I did not done any research) if it is possible and still allow to run it from single command (I means directly by CS fixer).

mvorisek:
partially fixes https://github.com/atk4/ui/issues/1236

I would still prefer moving these classes within `_includes`, but only for shorter per page code and possibly reuse across demos, but this PR solves the hinting/refactoring issue.

Also, the hint type is wrong/fake, it is `string` in reality...

### DIFF
--- a/demos/_unit-test/console_run.php
+++ b/demos/_unit-test/console_run.php
@@ -9,6 +9,7 @@ use atk4\ui\jsSSE;
 /** @var \atk4\ui\App $app */
 require_once __DIR__ . '/../init-app.php';
 
+/** @var \atk4\ui\View $testRunClass */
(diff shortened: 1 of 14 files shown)


ibelar (approved):
LGTM

PR #1304, opened 2020-06-19
Fix Demo - Crud caption for Anonymous class

abbadon1334:
https://ui.agiletoolkit.org/demos/collection/crud3.php

image

### DIFF
--- a/demos/collection/crud3.php
+++ b/demos/collection/crud3.php
@@ -12,6 +12,8 @@ $modelClass = get_class(new class() extends \atk4\data\Model {
     use ModelLockTrait;
 
     public $table = 'test';
+    
+    public $caption = 'Country';
 
     public function init(): void
     {
--- a/demos/collection/crud3.php
+++ b/demos/collection/crud3.php
@@ -12,7 +12,7 @@ $modelClass = get_class(new class() extends \atk4\data\Model {
     use ModelLockTrait;
 
     public $table = 'test';
-    
+
     public $caption = 'Country';
 
     public function init(): void


ibelar (approved):
Thanks!

mvorisek (changes requested):
Seems this can not be solved differently...

mvorisek (approved):

ibelar:
@abbadon1334 - Please fix CodeStyling prior to merge.

mvorisek (changes requested):
I think we should at least fix the issue - if class is anonymous, we should use parent class name.

ibelar:
> I think we should at least fix the issue - if class is anonymous, we should use parent class name.

@mvorisek - not sure what you mean here. This anonymous class extend data\Model. You want the CRUD button to display 'Add Model'? I think 'Add Country' is more appropriate.

mvorisek:
> You want the CRUD button to display 'Add Model'?

at least - see https://github.com/atk4/data/pull/633 - please review, then check if it works - and then this can be merged... But test my PR first.

ibelar:
> at least - see [atk4/data#633](https://github.com/atk4/data/pull/633) - please review, then check if it works - and then this can be merged... But test my PR first.

@mvorisek - To be honest, I do not see the necessity of running an extra test in data/Model for this. This anonymous class behaviour would probably be used rarely. All you have to do is change the caption, simple.

mvorisek:
@ibelar I meant to test the PR manually if it works,,,

image

done it by myself now

mvorisek (approved):

PR #1305, opened 2020-06-20
[refactor] use Model\UserAction class

georgehristov:
follows the refactoring from https://github.com/atk4/data/pull/634
Refactoring from `UserAction/Generic` to more semantically clear `Model/UserAction`

_Note: test failing due to dependency on https://github.com/atk4/data/pull/634_


mvorisek:
Merge `develop` and check feedback for constants in `data`. I absolutely hate SCOPE_xxx for something that is not related with Scope directly.

I can then quick review both and we can merge...

georgehristov:
Replaced by #1308

PR #1308, opened 2020-06-22
[refactor] model user action executors

georgehristov:
Refactoring of user action executor classes for clarity and consistency

Depends on https://github.com/atk4/data/pull/634, extends https://github.com/atk4/ui/pull/1305

BC break: 
- `instanceof` deprecated classes and interfaces will not work as expected

_Note: test failing due to dependency on https://github.com/atk4/data/pull/634_


mvorisek (approved):

mvorisek:
Run unit tests locally before merging, there are still uses of deprecated/originally named methods.

```
There were 9 failures:

1) atk4\ui\tests\DemosTest::testDemosStatusAndHtmlResponse with data set #10 ('collection/actions.php')
Deprecated: Method Model::addAction is deprecated. Use Model::addUserAction instead...
```

mvorisek:
@georgehristov can you fix so we can merge?

georgehristov:
Yes, I will work on it tonight and fix it

Issue #1024, opened 2020-03-29
Simplify/fix names like DropDown to Dropdown


mvorisek:
Class names likes `DropDown` should be changed to `Dropdown`. This is a standard across almost every UI framework - just google "DropDown".

`Drop` class will never exists or it will mean something totally different. In contrast and for this reason classes like `CardSection` are legit.

Normally, this will be a huge BC break, but as PHP classes are case-insensitive, this change should be only cosmetic cleanup.

Please provide also feedback for renaming `jsXxx` classes to `JsXxx` as required by PSR-1 standard:

> Class names MUST be declared in `StudlyCaps`.

Is PR for this welcomed?

rule:
### use capital letter when you will use it in snake case - ex. `mysql`, not `my_s_q_l`

romaninsh:
We had some internal policy for this but not for DropDown specifically. I agree with your proposal as long as it is BC

mvorisek:
implemented

PR #1309, opened 2020-06-23
Rename "FormField" NS to "Form\Field" + rename "Generic" classes to the NS (like Form\Field)

georgehristov:
Replaced by: https://github.com/atk4/ui/pull/1311

Merge after: https://github.com/atk4/ui/pull/1308

Organize NS for Form field and layouts under Form namespace.

- Elegant and clear syntax when importing the `atk4\ui\Form` namespace only

```
use atk4\ui\Form;

new Form\Field\Checkbox();
new Form\Layout\Columns();
```

- docs updated with explicit variable naming and new namespace importing for clarity, e.g.
`$group->addField('gender', [Form\Field\Dropdown::class, 'values' => ['Female', 'Male']]);`
`$f` -> `$form`, `$gr` -> `$group`, etc

BC break: 


mvorisek:
@georgehristov 

At the first sight, FormField looks ok, but the points are good!

My feedback:
- can you split the initial commit into two - one for rename and another one for BC (it can be then checked much easier)
- can be add an another label? :D
- use better PR names
- if waiting for some PR - alway specify in desc!
- will be good to finish the previous refactoring PRs - we may change the largely before...
- will be good if you can add NS "\atk4\ui\Form" uses (and remove all ...Form/Field)
- will be good if this will be even fixed by CS fixer (to not allow use .../Field or simillar)
- we need to merge this asap
- https://github.com/atk4/ui/issues/1024 - can you also make PR for this - change all `jsXxx` to `JsXxx`, `CRUD` to `Crud` and other badly named classes/NSes?
- all done by hand, or with any tool? (like renaming the variables - but for me, it seems not complete - maybe to commit the renames as 3rd commit (can be done easily by starting from zero, renaming NS, then the BC classes, and then restore the current PR))

georgehristov:
@mvorisek 

- I do not understand the split you mention - I just created the new namespace with class names correct there. 2 good things in one go.
- previous cascading PRs are RTM, subject to review and approval + timing for merge between releases. This one has one or two lines depending on them but anyway I branched it from there.
- good point about using `atk4\ui\Form` NS only, code has been updated. Would appreciate if you work on the CS Fixer
- https://github.com/atk4/ui/issues/1024 is on my agenda, including refactoring Table NS to `atk4\ui\Table`, etc
- variables renaming in demo is by hand file by file to avoid issues. If some remaining process should continue. But target in code and demos is to set standard of explicit variable names and avoid `$m`, `$f`, `$c`, etc

Merging timing (if at all) is of course at the discretion of community and members.
I do not expect that it would go down easy (due to BC-break with `instanceof`) but I am sure it is a step in the right direction so I made it.

mvorisek:
> I do not understand the split you mention - I just created the new namespace with class names correct there. 2 good things in one go.

Then, when I view commit by commit, I can see what was updated - with one commit, I see that the old classes changed completely and a lot of new was added - with two commits, I should see refactoring + added BC classes in new commit - it should be easy, when you not commit the updates of the BC/original classes in the 1st commit

> good point about using atk4\ui\Form NS only, code has been updated. Would appreciate if you work on the CS Fixer

at least manually for now

> variables renaming in demo is by hand file by file to avoid issues. If some remaining process should continue. But target in code and demos is to set standard of explicit variable names and avoid $m, $f, $c, etc

should be done once there is no other open refactoring PR

georgehristov:
Replaced by https://github.com/atk4/ui/pull/1311

PR #1311, opened 2020-06-23
Rename FormField NS to Form\Control (and FormLayout to Form\Layout)

### LINKED ISSUE #794: Docs: Refer to Form Fields as 'FormField' always?
Currently writing some documentation for Multiline Component, I remember how much I struggled differing atk4\data\Fields and atk4\ui\FormFields as the latter usually represent the former.

I thought it might be a good Idea to refer to form fields as 'FormField' always in the docs. Data fields are already referred to as 'Field' usually. This would make it more clear if we are talking about the data field storing the data or the ui form field representing that data.

georgehristov:
fixes #794

**Introduce Form namespace**

- Elegant and clear syntax when importing the `atk4\ui\Form` namespace only

```
use atk4\ui\Form;

new Form\Control\Checkbox();
new Form\Layout\Columns();
```

- docs updated with explicit variable naming and new namespace importing for clarity, e.g.
`$group->addField('gender', [Form\Control\Dropdown::class, 'values' => ['Female', 'Male']]);`
`$f` -> `$form`, `$gr` -> `$group`, etc

**BC break:** 

Autoload of old classes will trigger `DEPRECATED` warning

**Migration:**

Simple regex: `(?<!\w)FormField(?!\w)`

Search/replace in files

1. class `FormField\Generic` to `Form\Control`
2. class `FormLayout\Generic` to `Form\Layout`
2. namespace `atk4\ui\FormField` to `atk4\ui\Form\Control`
2. namespace `atk4\ui\FormLayout` to `Form\Layout`
3. class `FormField\CheckBox` to `Form\Control\Checkbox` _(change of class name case only)_
4. class `FormField\DropDown` to `Form\Control\Dropdown` _(change of class name case only)_
5. class `FormField\DropDownCascade` to `Form\Control\DropdownCascade` _(change of class name case only)_
6. class `FormField\MultiLine` to `Form\Control\Multiline` _(change of class name case only)_
7. class `FormField\UploadImg` to `Form\Control\UploadImage`
8. class `FormField\TextArea` to `Form\Control\Textarea` _(change of class name case only)_


PhilippGrashoff:
I like the idea about the Form namespace. However, I know at least of 3 people (me included) that confused Model Fields and Form Fields when starting with ATK. My idea to avoid this was to refer to Form Fields as FormField always (Docs, Methods, whatever). So Form would have getFormFields(), getFormField() methods etc.

I doubt having Form/Field namespace alone will do the job to avoid this confusion in future. How about Form/FormField?

Also see #794

georgehristov:
@PhilippGrashoff the idea is to force always use the Form namespace which will also solve the problem for newcomers
So you declare `use atk4\ui\Form` and then use `Form\Field` or `From\Field\Checkbox` for instance.
I have changed all ui demos accordingly so demos code is self explanatory in this regard and avoids confusion.
Same will be refactored for atk4/data and Model namespace introduced so you will use `Model\Field` or `Model\Field\Boolean` for instance to refer to model fields.
I think this solves the problem in a very elegant way using power of php namespaces.

mvorisek:
@georgehristov FormField rename - this is valid only if you can guarantee than FormField can not be used on its own without a Form - is this 100% true?

Based on my search it is currently used at least in Grid for searchbox /wo Form. Please comment on this case.

Based on my research the <input> /wo <form> is valid by html spec - https://stackoverflow.com/questions/3294572/is-input-well-formed-without-a-form . With the same reasoning, the FormLayout can (does it work now in Atk now?) be used without Form probably too.

This is also related with my PR https://github.com/atk4/ui/pull/1275 to support nested forms. We may simply always require a form to support all features. Only with this approach we can guarantee, that the input in for. ex. Grid was not mistakely wrapped inside another form (thus it is not serialized etc).

How does other popular frameworks, even not PHP ones, solves this?

georgehristov:
We are mixing several issues here:

- first impression / learning curve
This is purely DX / perception issue. By introducing a clear structure and importing only Form namespace in demos / docs / code we keep it clear to what the code is referring to `From\Field\Checkbox` instead of `Checkbox` only. 
Model fields will be `Model\Field\Boolean` for instance.
Devs can then implement imports as they like in their code but many times this is influenced by the base code of the toolkit/framework.
One needs to understand what model and form are from the docs and then which field is used in the demos.
My other idea of calling `Form\Field` something different (like `Form\Input`) was rejected. This would put even more distinction and clarity IMHO.
So apart from that I believe this is as clear as it gets for avoiding confusion.

- functionality
ATK now supports using form field without attaching it to a form. This is a matter of a different discussion.

mvorisek:
> ATK now supports using form field without attaching it to a form. This is a matter of a different discussion.

it is not, because this can very easily imply that decision to move it under Form NS si wrong - it can be used without Form, it is just one of a View building block addable to anywhere, element, which can be used in Form but you do not have too.

But I think this is a step in the right direction and there are no additional downsides (like with your other refactoring PRs). I will check now.

After mergin this, we should:
- advise people to import `atk4\ui\Form` NS only (answer to @PhilippGrashoff)
- rethink https://github.com/atk4/ui/issues/794 - we need to provide consistent API, ie. we should be able to tell instantly what methods like `addField` etc. accept in ui

src/Form/Field/ScopeBuilder.php
```diff
@@ -0,0 +1,611 @@
+<?php
```
mvorisek: @georgehristov with there PRs, I have to be able to rely on you that these mistakes are not present

src/Form/Field/Input.php
```diff
@@ -0,0 +1,262 @@
...
+     * @param string        $spot   Template spot
+     *
+     * @return Button
+     */
+    protected function prepareRenderButton($button, $spot)
+    {
+        if (!is_object($button)) {
+            $button = new Button($button);
+        }
+        if ($button instanceof \atk4\data\Model\Action) {
```
mvorisek: again, rename not introduced yet!

src/Form/Field/ScopeBuilder.php
georgehristov: @mvorisek that is why we have the review process ...

src/Form/Field/Upload.php
```diff
@@ -154,7 +154,7 @@ public function setFileId($id)
     /**
      * Add a js action to be return to server on callback.
      */
-    public function addJSAction($action)
+    public function addJsAction($action)
```
mvorisek: perfect!

mvorisek (approved):
@georgehristov LGTM, merge immediatelly if you are done

src/Form/Field/ScopeBuilder.php
mvorisek: yes, but you should optimize your workflow to prevent such mistakes - if you do it cleanly, it can not happen

georgehristov:
@mvorisek thanks for review and fixes.
As this is major I will await @romaninsh, @ibelar @DarkSide666 or @abbadon1334 to also review and approve

mvorisek:
Fine, they have 24 hours :)

two more things:
- is it ok with to remove the old classes after after 1/2 year instead? (there will be enough releases in between)
- can you do same (feel free on top of this PR with "merge after: #1311" text) for `TableColumn`?

georgehristov:
Timeframe I will not set :)
Table namespace is in progress ...

PhilippGrashoff:
To me, the learning curve/mixing form fields wiht model fields issue is not entirely solved, even with the namespaces. The namespace structure will be reflected in method names, so Model will have getField() and Form will have getField().
This is especially tricky as a form has a model. So we have
```
$form->getField('somename'); //returns a Form/Field
$form->model->getField('somename'); //returns the model field
```
As long as this situation exists, there will always be confusion.
Also have a look at the help channel in Discord, IIRC there lately was another one confused where to define fields: In the form or in the model. (Though this also due to the existance of Form->addField(), which adds a FormField and a Model field :) )

mvorisek:
I agree terminology around core things is very important. But I am not sure if it is the right moment to rename the methods.

Based on https://www.w3.org/TR/html52/sec-forms.html#attributes-common-to-form-controls the only alternative is Control. I am not sure if it is better - control does not imply field in something.

@PhilippGrashoff wdyt? Is your concern solely about the methods? What do you propose? Are they any other ilogical/confusing terms somewhere else?

georgehristov:
@PhilippGrashoff I still believe a different name is the best way to distinguish form fields and I came up with 'element' as it really is a View element.
So 
`Form\Element\Checkbox`
`$form->getElement('name');`
`$form->model->getField('name');`

Now is the best time for this change together with NS refactor.

So @romaninsh, @DarkSide666, @ibelar, @mvorisek, @abbadon1334  wdyt

mvorisek:
@georgehristov I done some more research.

Firstly, you can very quickly notice, that form controls are can be divided into two groups:
- \<input>
- "other form control"

evidence:
- Bootstrap - no support for complex types like Dropdown, skipping
- SUI - https://semantic-ui.com/modules/dropdown.html - nothing special about them (just like any html element)
- Angular - https://angular.io/api/forms/FormControl - they use term `control` solely (but there is no complex type is in their package like dropdown)
- Django - https://docs.djangoproject.com/en/3.0/ref/forms/fields/ - they use term `field` solely (but there is no complex type is in their package like dropdown)
- https://devdojo.com/guide/html/form-elements (link from Google for "HTML control") - `element` seems to be used solely for HTML element, not on higher abstractions
- Grommet - https://v2.grommet.io/components - nice overview of that above

this confirms the following:
- \<input> based element are called almost solely `input` / `field` / `control`
- `input` seems to be never used for something different than \<input>
- `field` seems to be never used for components that can provide multiple values (ie. no multiline)
- `control` seems to be used more widely (in Angular is seems like we use `field` now), in Grommer is it now used for `inputs`, but the message is clear from their overview page

so my opinion is:
- deprecate `Line` (but keep `MultiLine`, because we do not have enought abstraction for JS now to make it universal for every input control)
- keep current structure and rename `field` to `control` (as `form control` is term used widely for more abstraction than element/input) - @PhilippGrashoff wdyt, does unified and uniq term `control` (`Form\Control` NS, methods like `addControl`) solves your feedback? @georgehristov wdyt?

For the future, we may consider moving other classes under more groups/NSes like `Media`, `Utilities` (inspired by Groomet, but grouping for classes like `Grid`under `Layout` seems interesting, but for current complex and not enough abstraction support yet seems wrong to me)

georgehristov:
@mvorisek I finished with `Element` refactor before I saw your research. It can be easily modified based on feedback though.

- `Element` is a term already used in ATK for exactly descendants of `View` nested in another `View`.
- a bit of inspiration from legacy PHP PEAR HTML_Quickform where it is also `$form->getElement` and refers to the complete unit (label + input + validation feedback, etc)

mvorisek:
@georgehristov exactly, inspired by HTML and we should abstract to standard building block/component level

another reason - what is the difference between View and Element? None. So this seems like only confusing to me.

would be also good if you can also review my PRs where I requested a review from you

georgehristov:
Ok, `Control` is also fine with me but with slight preference to `Element`.
Let's see feedback from others.

mvorisek:
@georgehristov
- element is already used internally - so form element (or any View class) should now have elements (but no a form one) too?
- are you aware, that we need `field` like name which represents an input control, that "can hold data and allow to edit them". `element` makes me think form button should be also an `element`, but in reality, I think this is not/should be not true

mvorisek:
> deprecate Line (but keep MultiLine, because we do not have enought abstraction for JS now to make it universal for every input control)

opened an issue for it https://github.com/atk4/ui/issues/1316

as ui is a top repo (no usage in `data`), I think we can do it in this PR too

georgehristov:
Think of `Form\Element` NS defining field specific elements (also having `fieldElements` property in `Form`).
Button is an element but not `fieldElement`. It can also be referred to as Control as it controls the form.

mvorisek:
@georgehristov did you read my comments on today for this? `Button` already has elements! Any View does :)

georgehristov:
@mvorisek I know.
The most correct word as referred to in ATK should be `Form\Decorator`. More food for thought...

mvorisek:
That is not correct, it is not a decorator of form.

It is field (so an option to stay) or control - see Angular and Grommet. Or feel free to provide results of your own research.

The important point is, that we need a word representing a component, which offers a bidirectional mapping of form data. Therefore, any general name is incorrect.

georgehristov:
More ideas - switch to `Form\FieldDecorator` and `Table\ColumnDecorator` for consistency?
We have `decoratorFactory` methods in both `Form` and `Table` and many variables named according to this logic.

Clear, distinctive and consistent ...

mvorisek:
I opened https://github.com/atk4/ui/issues/1317 to actually drop any decorator naming in sense of seed

> More ideas - switch to Form\FieldDecorator and Table\ColumnDecorator for consistency?

I think you did not solved anything. You will still use (excl. class names/NSes) word `field` :D (and you will have to use FormDecorator, ... to be consistent)

ibelar:
> Ok, `Control` is also fine with me but with slight preference to `Element`.
> Let's see feedback from others.

I also prefer Control over Field or Element. They can be made from a single HTML element or a group of HTML elements, like dropdown, multiline, tree selector, etc. 

The W3 org recommendation also refer to form controls: https://www.w3.org/TR/html52/sec-forms.html#sec-forms

> A form is a component of a Web page that has form controls, such as text fields, buttons, checkboxes, range controls, or color pickers. A user can interact with such a form, providing data that can then be sent to the server for further processing (e.g., returning the results of a search or calculation). No client-side scripting is needed in many cases, though an API is available so that scripts can augment the user experience or use forms for purposes other than submitting data to a server.

Some framework refers to them as Control also:
 - Vuetify: https://vuetifyjs.com/en/
 - Buefy: https://buefy.org/documentation
 - Bootstrap as the 'form-control' css class

georgehristov:
Now re-written for `Form\Control` including docs

Docs will benefit from another pair of eyes scrolling though ...

docs/autocomplete.rst
```diff
@@ -1,15 +1,15 @@
 
 .. _autocomplete:
 
-==================
-AutoComplete Field
-==================
+====================
+AutoComplete Control
```
mvorisek: ... Form Control?

docs/multiline.rst
```diff
@@ -120,14 +120,14 @@ If you want to edit them along with the user, Multiline is set up in a few lines
...
-Now, there is another Multiline form field to add, edit or delete the users email addresses:
+Now, there is another Multiline form contol to add, edit or delete the users email addresses:
 
 .. image:: images/multiline_email_address.png
 
 
 Multiline and Expressions
 =========================
-If a Model has Expressions, they automatically get updated when a field value is changed. A loading icon on the ``+`` sign indicates that the expression values are updated.
+If a Model has Expressions, they automatically get updated when a control value is changed. A loading icon on the ``+`` sign indicates that the expression values are updated.
```
mvorisek: prepend "form " before any control?

georgehristov: done

tests/FormTest.php
```diff
@@ -101,7 +101,7 @@ public function testFormSubmit()
 
     public function testTextArea()
     {
-        $this->f->addField('TextArea');
+        $this->f->addControl('TextArea');
```
mvorisek, line 104: we should fix case everywhere

tests/FormTest.php
```diff
@@ -101,7 +101,7 @@ public function testFormSubmit()
 
     public function testTextArea()
```
mvorisek, line 102: we should fix case everywhere (ie. Textarea)

demos/input/multiline.php
```diff
@@ -40,14 +41,14 @@ public function init(): void
     $inventory->saveAndUnload();
 }
 
-$f = \atk4\ui\Form::addTo($app);
-$f->addField('test');
+$form = Form::addTo($app);
+$form->addField('test');
 // Add multiline field and set model.
-$ml = $f->addField('ml', [\atk4\ui\FormField\MultiLine::class, 'options' => ['color' => 'blue'], 'rowLimit' => 10, 'addOnTab' => true]);
+$ml = $form->addField('ml', [Form\Control\Multiline::class, 'options' => ['color' => 'blue'], 'rowLimit' => 10, 'addOnTab' => true]);
```
mvorisek: is `addField` here correct?

src/Form/AbstractLayout.php
```diff
@@ -0,0 +1,210 @@
...
+            throw (new Exception('Unable to add form field', 0, $e))
+                ->addMoreInfo('name', $name)
+                ->addMoreInfo('control', $control)
+                ->addMoreInfo('field', $field);
+        }
+
+        return $this->addControlElement($control, $field);
+    }
+
+    protected function addControlElement($decorator, $field)
```
mvorisek: Keep `_addControl`, we actually add Control here

demos/form/form-section.php
```diff
@@ -26,93 +28,93 @@
...
+$cols_layout = $form->layout->addSubLayout(Form\Layout\Section\Columns::class);
 
 $c1 = $cols_layout->addColumn();
 $c1->setModel($m, ['iso', 'iso3']);
 
 $c2 = $cols_layout->addColumn();
 $c2->setModel($m, ['numcode'/*, 'phonecode'*/]);
 
-$f->addField('phonecode');
+$form->addField('phonecode');
```
mvorisek: check `(add|get)(field|element)` matches

after I unsupressed the deprecation output, there are a lot of warnings

but warnings are probably not enought to check, as not all callbacks are triggered in tests

src/Form/Layout.php
```diff
@@ -117,7 +117,7 @@ public function addGroup($label = null)
      *
      * @return static
      */
-    public function addSubLayout($seed = self::class, $addDivider = true)
+    public function addSubLayout($seed = [self::class], $addDivider = true)
```
mvorisek, line 120: FYI: this was missed during merge...

georgehristov: Strange because i manully corrected it and force pushed.
Still tying to sort out the Table\Column refactor after the strict array seed merge...
We should not overlap with this as it causes a lot of time wasted.

mvorisek: image

I now checked the commit history and this PR was failing before merge...

`Table\Column`, yes, originally, you started on develop:
image

so it will be probably better to simply merge `develop` there (and change target branch in that PR to `develop`), as strict seeds PR is easy and resolving conflicts is basically adding `[` and `]` around based on that diff
- if you do not have the orignal branch based on `develop` - I pushed it under `refactor_tablecolumn_develop_based`

georgehristov: One failing test on Column\FilterPopup remaining.
Problem is these files were completely moved so merging will not help. It is anyway based on latest develop / form-namespace.

mvorisek: I fixed this PR. Originally, there was a bad merge and fix that was waiting here was not/still is not applied to develop, to merged it manually here.

How much RTM is this PR?

georgehristov: After one or two more reviews we can merge. Also release schedule to be considered - if it will be included in 2.2 or 2.3

mvorisek: We merge very quickly to prevent conflicts. Releases are up to Romains, in theory, you can pick any datr, take latest commit from each repo and do a release on your own if needed...

One more thing - in this PR, we normalize DropDown to Dropdown for Form\FieldDropdown. Immediately after this PR, we should fix all other names in whole ui repo.

src/Form.php
```diff
@@ -197,15 +207,25 @@ protected function initLayout()
...
+
+    /**
+     * Setter for control display rules.
      *
      * @param array $rules
      *
      * @return $this
      */
-    public function setFieldsDisplayRules($rules = [])
+    public function setControlDisplayRules($rules = [])
```
mvorisek: did you miss `s` after `Control`? (here and for prop name)

later, we should probably move this to control directly, but no to mix this up for now in this PR...

georgehristov: Removing "s" was intentional (avoid double plural https://english.stackexchange.com/questions/227378/usage-of-double-plural)

mvorisek: I understand, but in computer science it seems wrong to me - we set (multiple) rules fo (multiple) controls

tests/FormTest.php
mvorisek: @georgehristov wdyt? `TextArea`, `DropDown`, ... Probably everywhere expect `\atk4\ui\DropDown` which should be fixed right after we merge this in another PR

mvorisek:
two cosmetic changes to fix, then we can merge it this afternoon because it is almost BC

src/Form.php
georgehristov: The code as we see it is for humans:) I would keep it that way

mvorisek: I disagree quite strongly. Here we have a bad design (rule set of rules, rules not set to child controls) and we need to keep the plurality to underline this

georgehristov: No arbiter so one has to concede. Changed now. Moving on ...

PhilippGrashoff:
Ah, Control is nice :)

mvorisek:
@PhilippGrashoff thanks for feedback, I mean for both! :D

PR #1312, opened 2020-06-24
Throw standard exception on "callback never reached" error

mvorisek:
No need for late error - even if some output has been already sent, if will be detected by output writer

Why it is important: to be able to pass/display exception in json as any other to/in browser

image

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -516,11 +516,7 @@ class App
             $this->hook(self::HOOK_BEFORE_OUTPUT);
 
             if (isset($_GET['__atk_callback']) && $this->catch_runaway_callbacks) {
-                $this->setResponseStatusCode(500);
-                $this->terminate(
-                    $this->buildLateErrorStr('Callback requested, but never reached. You may be missing some arguments in request URL.'),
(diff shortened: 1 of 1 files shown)


src/App.php
```diff
@@ -1091,8 +1082,8 @@ protected function outputResponse(string $data, array $headers): void
             }
         }
 
-        if ($lateErrorStr !== null) {
-            echo $lateErrorStr;
+        if ($lateError !== null) {
+            echo "\n" . '!! ATK4 UI ERROR: ' . $lateError . ' !!' . "\n";
```
georgehristov, line 1086: I would change "ATK4 UI" to `$app->title` if possible here to enable white labeling of application or just remove it

georgehristov (approved):
Idea is very good, only one comment (unrelated but can be taken care of now)

src/App.php
mvorisek: This is fatal error that should never apper in production.

mvorisek: We already have simillar error in core for exception renderer error.

PR #1313, opened 2020-06-24
Rename TableColumn NS to Table\Column

georgehristov:
merge after: https://github.com/atk4/ui/pull/1311

**Introduce Table namespace**

- Elegant and clear syntax when importing the `atk4\ui\Table` namespace only

```
use atk4\ui\Table;

new Table\Column\Checkbox();
```

**BC break:** 

Autoload of old classes will trigger `DEPRECATED` warning

**Migration:**

Simple regex: `(?<!\w)TableColumn(?!\w)`

Search/replace in files

1. class `TableColumn\Generic` to `Table\Column`
2. namespace `atk4\ui\TableColumn` to `atk4\ui\Table\Column`
3. class `TableColumn\CheckBox` to `Table\Column\Checkbox` _(change of class name case only)_
4. class `TableColumn\HTML` to `Table\Column\Html` _(change of class name case only)_
5. class `TableColumn\jsHeader` to `Table\Column\JsHeader` _(change of class name case only)_
5. class `TableColumn\FilterModel\Generic` to `TableColumn\FilterModel`


mvorisek:
tests are failing...

georgehristov:
> 
> 
> tests are failing...

That is why it is in draft :)

mvorisek:
but already requested a review...

src/Table/Column/Filter/Model.php
```diff
@@ -0,0 +1,152 @@
...
+
+use atk4\core\NameTrait;
+use atk4\core\SessionTrait;
+use atk4\data\Field;
+use atk4\data\Persistence;
+
+/**
+ * Implement a generic Type model for filtering data.
+ */
+class Model extends \atk4\data\Model
```
mvorisek: This must be changed. Model is used a lot and any confusion with it is highly unwanted. Previously, this was named "FilterModel\Generic".

"FilterPopup" remained, we can probably keep FilterModel as well.

src/TableColumn/Image.php
```diff
@@ -4,29 +4,13 @@
 
 namespace atk4\ui\TableColumn;
 
+if (!class_exists(\SebastianBergmann\CodeCoverage\CodeCoverage::class, false)) {
+    'trigger_error'('Use atk4\ui\Table\Column\Image instead', E_USER_DEPRECATED);
+}
+
 /**
- * Column for formatting image.
+ * @deprecated will be removed jun-2021
```
mvorisek: I think we should target for 1/2 year instead - this is to allow enought releases in between for smooth transition, otherwise we want to remove it asap

same for prev PR

mvorisek (changes requested):
otherwise seems good

mvorisek (approved):
LGTM, but I still think the removal of derecations should be planned sooner - ok with you? Are there any other reason than BC? Once people upgrade, I do not see any point to keep them...

georgehristov:
No reason, I will change to dec-2020.
We'll see when this gets released

src/TableColumn/Status.php
```diff
@@ -4,91 +4,13 @@
...
 namespace atk4\ui\TableColumn;
 
-use atk4\data\Model;
+if (!class_exists(\SebastianBergmann\CodeCoverage\CodeCoverage::class, false)) {
+    'trigger_error'('Use atk4\ui\Table\Column\Status instead', E_USER_DEPRECATED);
+}
 
 /**
- * Implements Column helper for grid.
+ * @deprecated will be removed jun-2021
```
mvorisek: this needs to be updated, PR rebased and at least from my side LGTM

georgehristov: updated

georgehristov:
Shall we merge this and then take up change of case on classes?

mvorisek:
Yes, wait 5 minutes, I am doing small finalization changes.

mvorisek:
@georgehristov review changes,

two more matches to be resolved:
- `\w(?<!add|get|Save|Load|Form)Field` (like AboveFields)
- `class.*field`

mvorisek:
@georgehristov did you checked these regexes, wdyt? I think we should fix at least some to be consistent.

mvorisek:
> 
> 
> Shall we merge this and then take up change of case on classes?

after the fixes for the past two PRs done, yes, and - not only class names, but also method names. (case strictly, no refactoring from snake case etc.)

PR #1315, opened 2020-06-24
Use strict array seed

mvorisek:
merge after: https://github.com/atk4/data/pull/639

no functional change, related with https://github.com/atk4/core/pull/249

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Create "index.php" file with:
 require_once __DIR__ . '/vendor/autoload.php';
 
 $app = new \atk4\ui\App();   // That's your UI application
-$app->initLayout(\atk4\ui\Layout\Centered::class);
+$app->initLayout([\atk4\ui\Layout\Centered::class]);
 
 $form = \atk4\ui\Form::addTo($app); // Yeah, that's a form!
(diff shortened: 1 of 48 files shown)


georgehristov:
Why you decided to switch to array only from "class/object or array"?

mvorisek:
to be exact to "object or array" (i.e. the only required change is to wrap class name/string seeds to an array - if something needs to be changed more, it is more or less bug)

- Because the current design is mixing almost everything and something even the author does not know what to put where.
- These are hidden bugs
- As the seeds are not in a normalized form, a lot of code (maybe 10% of all code) preprocess them "somehow working", but:
  - the behaviours is very inconsistent
  - it is not/can not be tested
  - there are hidden bugs (in preprocessing as well in calls)
  - this extra code can not be maintaned

in the future, there the code can be cleaned up and behaviour can be normalized to clear seed definition (see core/Factory methods)

basically, there should be no specific preprocessing except merging seed with some defaults/forced overrides (passive vs. active merge) and then straight factory call

DarkSide666 (approved):
Another breaking change, but oh well... LGTM

PR #1318, opened 2020-06-25
[fix] - Table Generic Hook const

ibelar:
From private to public

### DIFF
--- a/src/TableColumn/Generic.php
+++ b/src/TableColumn/Generic.php
@@ -25,8 +25,8 @@ class Generic
 
     /** @const string */
     public const HOOK_GET_HTML_TAGS = self::class . '@getHTMLTags';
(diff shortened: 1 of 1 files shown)


src/TableColumn/Generic.php
```diff
@@ -84,6 +84,7 @@ class Generic
      */
     public function __construct($defaults = [])
     {
+        $t = self::HOOK_GET_HTML_TAGS;
```
mvorisek: remove?

mvorisek (approved):
LGTM except some extra staged/committed code

DarkSide666 (approved):
LGTM, but do not need that line

src/TableColumn/Generic.php
ibelar: oops! done.

PR #1320, opened 2020-06-25
Hotfix string seeds

mvorisek:
missed in https://github.com/atk4/ui/pull/1315

### DIFF
--- a/src/ActionExecutor/UserAction.php
+++ b/src/ActionExecutor/UserAction.php
@@ -81,9 +81,9 @@ class UserAction extends Modal implements Interface_, jsInterface_
     /**
      * A form for action argument and fields user entry.
      *
-     * @var string
+     * @var array
      */
-    public $form = Form::class;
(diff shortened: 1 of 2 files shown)


mvorisek:
@georgehristov fix is here for 6 hours already :), can you approve?

georgehristov (approved):

src/ActionExecutor/UserAction.php
```diff
@@ -81,9 +81,9 @@ class UserAction extends Modal implements Interface_, jsInterface_
     /**
      * A form for action argument and fields user entry.
      *
-     * @var string
+     * @var array
      */
-    public $form = Form::class;
+    public $form = [Form::class];
```
georgehristov, line 86: At least in the code comments we should mention it is a seed ...

mvorisek: I expect much more inconsistencies, worth to fix consistently at once.

PR #1324, opened 2020-06-26
[fix] Form Conditional Control

ibelar:
Reset css class name for jsConditionalForm. 
see: https://ui.agiletoolkit.org/demos/form/jscondform.php


georgehristov (approved):
Thanks @ibelar

PR #1325, opened 2020-06-26
[fix] Form Section

ibelar:
Small fix form Accordion Form section.

### DIFF
--- a/src/Form/Layout/Section/Accordion.php
+++ b/src/Form/Layout/Section/Accordion.php
@@ -27,7 +27,7 @@ class Accordion extends \atk4\ui\Accordion
             // default behavior
             $jsError = [$form->js()->form('add prompt', $fieldName, $str)];
 
-            // if control is part of an accordion section, it will open that section.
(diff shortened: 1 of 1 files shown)


src/Form/Layout/Section/Accordion.php
```diff
@@ -27,8 +27,8 @@ public function init(): void
             // default behavior
             $jsError = [$form->js()->form('add prompt', $fieldName, $str)];
 
-            // if field is part of an accordion section, will open that section.
-            $section = $form->getClosestOwner($form->getElement($fieldName), AccordionSection::class);
+            // if control is part of an accordion section, it will open that section.
```
georgehristov: Maybe replace with " if a form control ..."

georgehristov (approved):
Thanks

src/Form/Layout/Section/Accordion.php
ibelar: done!

PR #1328, opened 2020-06-26
Do not handle exception in UserAction executor

mvorisek:
Opening this, because:

1. when modal is opened like with "Add xxx" button in CRUD for new record popup
2. and error is thrown during the action execution like when

like when model to add is defined like:
```
class ZInboundParcel extends \AppAtk4\MyModel {
...
    public function init(): void {
        parent::init();

        ...
        $this->addField('xxx', ['ui' => ['form' => [TextArea::class]]]);
        ...
    }
}
```

to replicate this issue like descibed, you need:
- unix OS (otherwise `TextArea.php` source will be loaded)
- to not use/load `Textarea` class before it
- maybe also no autoload optimalization in composer (not sure, it may normalize class names thus making it CI)

This issue is not about refactoring, it is about supressing great error trace with absolutely unhelpful text

Before:
image

After:
image

Also notice, that the form was previously rendered, but partly one - WE SHOULD NEVER ALLOW PARTIAL RENDER (and partial submit)

Need to be probably solved a little bit differently - handlers/steps should be reset, but exception handler/renderer from App should be used.

### DIFF
--- a/src/UserAction/ModalExecutor.php
+++ b/src/UserAction/ModalExecutor.php
@@ -187,27 +187,23 @@ class ModalExecutor extends Modal implements JsExecutorInterface
         $this->loader->set(function ($modal) {
             $this->jsSetBtnState($modal, $this->step);
 
-            try {
-                switch ($this->step) {
-                    case 'args':
-                        $this->doArgs($modal);
+            switch ($this->step) {
+                case 'args':
+                    $this->doArgs($modal);
 
-                        break;
-                    case 'fields':
-                        $this->doFields($modal);
+                    break;
+                case 'fields':
+                    $this->doFields($modal);
 
-                        break;
-                    case 'preview':
-                        $this->doPreview($modal);
+                    break;
+                case 'preview':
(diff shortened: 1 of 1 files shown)


ibelar:
@mvorisek - The original idea was for the user to be able to step back and give him a chance to change the argument value if the error was caused by the argument. But I agree that it does not make sense when an error like this is thrown. 
Maybe we can keep it but only for error that would make sense, i.e. that a user can correct with a step back. I am not really sure how this would work either.

Maybe @romaninsh can comment further as he is the one who asks for the functionality at first.

mvorisek:
@ibelar I also pointed this (to step back), ok, for @romaninsh

I think this is related with other bugs related with "user can not return back on error" and "popup cleared but not closed on error" I already experienced.

Unhandled errors needs to be probably solved in browser level (as server/response might fail in any state)

UPDATE:
I hit this bug for another case.

I think we have to never step forward without success action execution result, so then we do not have to roll ui back.

mvorisek:
merged in #2008

PR #1326, opened 2020-06-26
Fix FormField refactoring

mvorisek:
@georgehristov check https://github.com/atk4/ui/commit/26bc53f40d884b3fc06fada54f44a40c47a75f05 more

https://github.com/atk4/ui/commit/314e0799d630ba9b6a6a3a1290a28600d33141ae#diff-2c77116739ece89c91385095c838f093R127 seems good hotfix, but if `field` is not required by FUI, we should refactor:
- `\w(?<!add|get|Save|Load|Form)Field` (like AboveFields)
- `class.*field` (probably to `form-control` class)
as well

also fix:
- `TextArea` (do CS search, one result also in `atk4/core`)
- check other classes with updated case and create a regex `(DropDown|TextArea|...)` from them to easy identify the locations to update

PR #1329, opened 2020-06-26
Fix FormField/FormLayout rename refactoring

georgehristov:


demos/form/templates/input.html
```diff
@@ -1,15 +1,15 @@
...
   <div class="{$width} {$class} fields">{$Content}</div>{$Hint}
 </div>{/}
 {NoLabelGroup}
-<div class="{$field_class} field atk-form-group">
+<div class="{$control_class} field atk-form-group">
   <div class="{$width} {$class} fields">{$Content}</div>
 </div>{/}
-{InputField}
-<div class="{$field_class} field">
+{LabeledInput}
```
mvorisek: very good!

src/Form/Layout.php
```diff
@@ -199,22 +199,25 @@ public function recursiveRender()
...
 
             // Fields get extra pampering
             $template->setHTML('Input', $el->getHTML());
             $template->trySet('label', $label);
             $template->trySet('label_for', $el->id . '_input');
-            $template->set('field_class', $el->getControlClass());
+            $template->set('control_class', $el->getControlClass());
+
+            // backward compatibility - will be removed in dec-2020
+            $template->trySet('field_class', $el->getControlClass());
```
mvorisek: BC without deprecation warning is good as a release will be made, but users will not notice until the support is dropped

I this is is not used that much/on so many places in custom apps, so I would remove it from templates and throw an exception here if a value was/is set (but with note to remove that if/exception later)

mvorisek:
@georgehristov see original feedback/description in https://github.com/atk4/ui/pull/1326 next time feel free to commit there directly, to minimalize PRs count/history

src/Form/Layout.php
```diff
@@ -209,8 +210,10 @@ public function recursiveRender()
             $template->trySet('label_for', $el->id . '_input');
             $template->set('control_class', $el->getControlClass());
 
-            // backward compatibility - will be removed in dec-2020
-            $template->trySet('field_class', $el->getControlClass());
+            // BC-break exception
```
mvorisek: missing clear date when it should be removed (I think 1/2 year is good here too, good in general)

src/Form/Layout.php
```diff
@@ -209,8 +210,10 @@ public function recursiveRender()
...
+            if ($template->hasTag('field_class')) {
+                throw new Exception('field_class region has be deprecated. Use control_class instead');
```
mvorisek, line 215: the same for `AboveFields` when used

georgehristov: the problem is AboveFields is set directly in dev custom code. I kept the region in the form template for BC

src/Form/Layout.php
georgehristov: I just added it but problem is later if we remove and the dev template contains region `field_class` it will just not work without warning

src/Form/Layout.php
mvorisek: But we can check if set and throw. There should be not that many usages and search/replace is quick...

georgehristov: For this I need to remove the region in the template (I kept it for now)

mvorisek: If absolutely needed, I this we should map the old region in render in PHP and use only the new one in template.

BC is good thing, but oneday everyone will have to upgrade and leaving deprecated things in template etc. might confuse others.

georgehristov: done now

demos/form/templates/input.html
```diff
@@ -7,9 +7,9 @@
 <div class="{$control_class} field atk-form-group">
   <div class="{$width} {$class} fields">{$Content}</div>
 </div>{/}
-{LabeledInput}
+{LabeledControl}
 <div class="{$control_class} field">
   <label for="{$label_for}">{$label}<span style="float: right">{$Hint}</span></label>{$Input}
 </div>{/}
-{NoLabelInput}
+{NoLabelControl}
```
mvorisek, line 14: isn't this for input tag solely?

mvorisek:
@georgehristov is it possible to address the points from https://github.com/atk4/ui/pull/1326 PR desc?

demos/form/templates/input.pug
```diff
@@ -1,25 +1,25 @@
...
 | {NoLabelGroup}
-div(class="{$field_class} field atk-form-group")
+div(class="{$control_class} field atk-form-group")
     div(class="{$width} {$class} fields")
         | {$Content}
 | {/}
 
-| {InputField}
-div(class="{$field_class} field")
+| {LabeledControl}
```
mvorisek, line 15: is this for general control or input only?

georgehristov: This is for the control so I named it accordingly.
Form control (in ATK) = label + input + hint + validation message

mvorisek: I agree that input is subpart of control (in Atk). But if this is related strictly with input, I think it is more logical to keep input name here.

georgehristov:
@mvorisek I have addressed all issues from #1326 in atk4\ui. I have not done the TextArea you mention in atk4\core yet

mvorisek:
@georgehristov what about thsi https://github.com/atk4/ui/commit/314e0799d630ba9b6a6a3a1290a28600d33141ae#diff-2c77116739ece89c91385095c838f093R127 ? Can we use `.control` class here, or is it required by FUI?

mvorisek (changes requested):
seems LGTM expect:
- local/input only region `NoLabelControl` should be named with input instead of control term
- I think `field` / `fields` classes should be renamed to `control` / `controls`. If required by FUI, keep it, but it seems that it is not, see https://fomantic-ui.com/elements/input.html#input (expand the code)

demos/form/templates/input.pug
georgehristov: It is not strictly with input as I wrote.
LabelledControl = label + input + hint + validation message
NoLabelControl = input + hint + validation message
Input = input (this is the region for the pure input)

georgehristov:
`field` class is required by Fomantic UI behaviors and CSS as well
https://fomantic-ui.com/behaviors/form.html#/examples

mvorisek:
Then only one feedback left - `NoLabelControl` and `LabeledControl`

georgehristov:
The Control is the "group" of regions that the control class adds as below. It is not strictly with input as I wrote.
LabelledControl = label + input + hint + validation message
NoLabelControl = input + hint + validation message

Input = input (this is the region for the pure input)

I believe "control" is the right word to use here. Otherwise, what do we define as a region/contents of a control?

mvorisek:
But the input is labeled, right? This is not that critical, so I will approve. I tend more to use `input` there, because it does not refer to control component but instead to input in that component.

mvorisek (approved):

mvorisek:
@georgehristov do you plan to move https://github.com/atk4/ui/blob/develop/src/Layout/Generic.php#L10 up as well?

georgehristov:
Yes, if we want to follow uniform naming of NS this would be atk4\ui\Layout.

mvorisek:
I am for it, it has to be consistent.

georgehristov:
Okay, I can do it tonight. But for data\Field refactor I will postpone. I will propose only the Persistence NS refactor

PR #1331, opened 2020-06-29
Fix Action button icon

DarkSide666:
`Button` class already takes care of icons so no need to (wrongly) do that again in ActionButtons class.

Consider example:
```
// create ActionButtons column without title
$column = $this->crud->table->addColumn(null, [ActionButtons::class, 'caption' => '']);

// see 1st parameter - seed for Button class, icon name passed as string
$column->addModal(['icon' => 'key'], 'Change Password', function ($v, $id) {...});
```


DarkSide666:
@mvorisek approved in chat with some comments

PR #1332, opened 2020-06-29
fix property names

DarkSide666:


mvorisek:
I think actions were renamed to userActions. for @georgehristov

DarkSide666:
No they are not, just method names have changed.
This is just a typo fix for *current* develop version because property is named `action` not `actions`.

mvorisek:
seems so, but I have nothing to say to this - @georgehristov may, or just be sure and merge it on your own credit :)

georgehristov (approved):
Thanks!

PR #1334, opened 2020-06-30
Optimize Travis composer install

mvorisek:
It reduces install time from about 35 to 20 seconds.

But still prefer to move Behat to GH actions... :)


georgehristov (approved):
LGTM

PR #1337, opened 2020-07-01
Move safe "@return bool" phpdoc to code

mvorisek:


mvorisek:
@ibelar can you look at this? Only bool should be passed as the last arg... https://github.com/atk4/ui/blob/fbf193d908e756c3ff0c63a3c8d0d13dd912e2ae/src/CardDeck.php#L179

georgehristov (approved):
Looks better!

ibelar:
> @ibelar can you look at this? Only bool should be passed as the last arg...
> 
> https://github.com/atk4/ui/blob/fbf193d908e756c3ff0c63a3c8d0d13dd912e2ae/src/CardDeck.php#L179

Fix!

PR #1338, opened 2020-07-02
No echo and output buffering in App::run()

mvorisek:
No phpunit warnings:
image

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -1081,7 +1081,7 @@ class App
         }
 
         if ($lateError !== null) {
-            echo "\n" . '!! ATK4 UI ERROR: ' . $lateError . ' !!' . "\n";
+            echo "\n" . '!! FATAL UI ERROR: ' . $lateError . ' !!' . "\n";
             exit(1);
         }
 
--- a/src/App.php
+++ b/src/App.php
@@ -497,7 +497,6 @@ class App
         $isExitException = false;
 
         try {
-            ob_start();
             $this->run_called = true;
             $this->hook(self::HOOK_BEFORE_RENDER);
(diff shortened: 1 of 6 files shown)


georgehristov (approved):
LGTM

src/App.php
```diff
@@ -516,12 +515,12 @@ public function run()
             if (isset($_GET['__atk_callback']) && $this->catch_runaway_callbacks) {
                 throw new Exception('Callback requested, but never reached. You may be missing some arguments in request URL.');
             }
-            echo $this->html->template->render();
+
+            $output = $this->html->template->render();
```
georgehristov, line 519: Probably good to initialize `$output` for cases this line is not reached

georgehristov (approved):

src/App.php
mvorisek: You are right! But still working on other unclosed buffers, see warnings in unit tests.

src/App.php
```diff
@@ -1012,8 +1012,6 @@ function () {
                 if (!$this->run_called) {
                     try {
                         $this->run();
-                    } catch (ExitApplicationException $e) {
```
mvorisek, line 1015: catched in `App::run()`, @georgehristov @abbadon1334 @ibelar wdyt?

tests/DemosTest.php
```diff
@@ -144,8 +144,7 @@ protected function getClient(): Client
                     throw $e;
                 }
             } finally {
-                $body = ob_get_contents();
-                ob_end_clean();
+                $body = ob_get_clean();
```
mvorisek, line 167: the same, one line only...

PR #1339, opened 2020-07-02
[update] use explicit argument names

georgehristov:
Simple change name of arguments

### DIFF
--- a/demos/_includes/Persistence_Faker.php
+++ b/demos/_includes/Persistence_Faker.php
@@ -19,17 +19,17 @@ class Persistence_Faker extends \atk4\data\Persistence
         }
     }
 
-    public function prepareIterator($m)
+    public function prepareIterator($model)
     {
-        foreach ($this->export($m) as $row) {
+        foreach ($this->export($model) as $row) {
             yield $row;
         }
     }
 
-    public function export($m, $fields = [])
+    public function export($model, $fields = [])
     {
         if (!$fields) {
-            foreach ($m->getFields() as $name => $e) {
+            foreach ($model->getFields() as $name => $e) {
                 $fields[] = $name;
             }
         }
@@ -40,13 +40,13 @@ class Persistence_Faker extends \atk4\data\Persistence
             foreach ($fields as $field) {
                 $type = $field;
 
-                if ($field === $m->id_field) {
+                if ($field === $model->id_field) {
                     $row[$field] = $i + 1;
 
                     continue;
                 }
 
-                $actual = $m->getField($field)->actual;
+                $actual = $model->getField($field)->actual;
                 if ($actual) {
                     $type = $actual;
                 }
@@ -60,8 +60,8 @@ class Persistence_Faker extends \atk4\data\Persistence
             $data[] = $row;
         }
 
-        return array_map(function ($r) use ($m) {
-            return $this->typecastLoadRow($m, $r);
+        return array_map(function ($r) use ($model) {
+            return $this->typecastLoadRow($model, $r);
         }, $data);
     }
 }
--- a/demos/_includes/Persistence_Faker.php
+++ b/demos/_includes/Persistence_Faker.php
@@ -60,8 +60,8 @@ class Persistence_Faker extends \atk4\data\Persistence
             $data[] = $row;
         }
 
-        return array_map(function ($r) use ($model) {
-            return $this->typecastLoadRow($model, $r);
+        return array_map(function ($row) use ($model) {
+            return $this->typecastLoadRow($model, $row);
         }, $data);
     }
 }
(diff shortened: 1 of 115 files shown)


mvorisek:
There are still a lot of other similar names (search for `\$m(?!\w)`, even as function params).

I think `$m` is ok as long as it is not used for anything else. But if it will be normalized everywhere, I think `$model` can be ok, but I would still prefer `$m` in anonymous functions etc.

What is the rule?

georgehristov:
We all are setting the rules but using explicit names makes code clear to read for newcomers and also for cooperating or returning to your own code.
I prefer avoiding single letter vars unless `$k=>$v` in some small loop.

mvorisek:
> I prefer avoiding single letter vars unless `$k => $v` in some small loop.

I agree. I think we should follow about this, but then it needs to be really consistent, not 50 PRs to fix something that should be fixed in one.

The same for `$f` which is even more ambiguous as used for `Form` and `Field`, maybe even for some `Control` still...

So at least all these should be fixed in this PR.

georgehristov:
Okay, I will search for all and add in this PR

PhilippGrashoff:
nice PR you're working on. I agree on not liking the 1 letter var names. $model is so much more meaningful than $m, as is $field instead of $f.

src/View.php
```diff
@@ -215,11 +215,9 @@ public function __construct($label = null, $class = null)
      *
      * @return Model
      */
-    public function setModel(Model $m)
+    public function setModel(Model $model)
     {
-        $this->model = $m;
-
-        return $m;
+        return $this->model = $model;
```
mvorisek: I think 2 lines are much better to read - you assign and then you return.

demos/_includes/Persistence_Faker.php
```diff
@@ -60,8 +60,8 @@ public function export($m, $fields = [])
             $data[] = $row;
         }
 
-        return array_map(function ($r) use ($m) {
-            return $this->typecastLoadRow($m, $r);
+        return array_map(function ($r) use ($model) {
```
mvorisek: maybe `$row`

demos/_includes/SomeData.php
```diff
@@ -8,21 +8,21 @@ class SomeData extends \atk4\data\Model
 {
     public function __construct()
     {
-        $p = new Persistence_Faker();
+        $persistence = new Persistence_Faker();
```
mvorisek: `$fakerPersistence`

src/View.php
georgehristov: reverted

demos/interactive/wizard.php
```diff
@@ -27,38 +27,38 @@
...
 // and set a custom js action or even set a different link. You can use recall()
 // to access some values that were recorded on another steps.
-$t->addStep(['Select Model', 'description' => '"Country" or "Stat"', 'icon' => 'table'], function (Wizard $w) {
+$wizard->addStep(['Select Model', 'description' => '"Country" or "Stat"', 'icon' => 'table'], function (Wizard $w) {
     if (isset($_GET['name'])) {
         $w->memorize('model', $_GET['name']);
         $w->app->redirect($w->urlNext());
     }
 
     $c = \atk4\ui\Columns::addTo($w);
```
mvorisek: ideally fix all one char variables related with components like this

demos/interactive/wizard.php
```diff
@@ -27,38 +27,38 @@
...
 
-    $t = \atk4\ui\Grid::addTo($c->addColumn(), ['paginator' => false, 'menu' => false]);
+    $wizard = \atk4\ui\Grid::addTo($c->addColumn(), ['paginator' => false, 'menu' => false]);
```
mvorisek: hmm, it is Grid... Wizar is defined above (line 46) and not renamed

demos/interactive/wizard.php
georgehristov: fixed

demos/interactive/wizard.php
georgehristov: fixed

demos/_includes/Persistence_Faker.php
georgehristov: fixed

demos/_includes/SomeData.php
georgehristov: fixed

mvorisek:
Still more names to refactor, example:
image

mvorisek (approved):
LGTM, but more renames can be done to make this PR really complete one

mvorisek:
All snake_case variable names should be fixed to camelCase - regex `(?<! (public|protected|private) )\$\w+_\w+`

georgehristov:
We can work on it further but let's merge this now to avoid conflicts.

mvorisek:
already approved, but I think at least another similar usecases in demos should be fixed

PR #1342, opened 2020-07-02
Fix type check

mkrecek234:
The type check returns a control, not a field, and will always fail. It has to be checked on the field object belonging to the control instead. Tested and working.

### DIFF
--- a/src/Form/Control/DropdownCascade.php
+++ b/src/Form/Control/DropdownCascade.php
@@ -43,7 +43,7 @@ class DropdownCascade extends Dropdown
 
         $this->cascadeInput = is_string($this->cascadeFrom) ? $this->form->getControl($this->cascadeFrom) : $this->cascadeFrom;
 
-        if (!$this->cascadeInput instanceof Field) {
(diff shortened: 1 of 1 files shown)


src/Form/Control/DropdownCascade.php
```diff
@@ -43,7 +43,7 @@ public function init(): void
 
         $this->cascadeInput = is_string($this->cascadeFrom) ? $this->form->getControl($this->cascadeFrom) : $this->cascadeFrom;
 
-        if (!$this->cascadeInput instanceof Field) {
+        if (!$this->cascadeInput->field instanceof Field) {
             throw new Exception('cascadeFrom property should be an instance of ' . Field::class);
```
georgehristov, line 47: I believe the lines should be changed to `instance of Form\Control::class`

There are more changes to do in this class property names to keep consistency so I will take it up later today.

mkrecek234:
```suggestion
            if (!$this->cascadeInput instanceof Control) {
```

georgehristov:
Replaced by #1343

PR #1348, opened 2020-07-05
Fix not freed objects/references

### LINKED ISSUE #1347: Something is not releasing memory, at least in tests
patch phpunit like:
image

I tested to change it to 5000x and all other repos except this passed - this repo failed with memory exhausted fatal error

mvorisek:
fixes https://github.com/atk4/ui/issues/1347

the issue was related with `App` which was not GCed as it remained referenced because of `register_shutdown_function` and other global handlers.

example after unit tests were run 5000x in the same process with this PR:
image
(originally failed after a few runs...)


georgehristov (approved):
Good catch!

PR #1349, opened 2020-07-06
Refactor Template and improve performance and type safety

mvorisek:
marking as `BC break` as the internal structure is no longer accessible from outside the class

otherwise fully BC, this feature is not used much in a custom code

even with additional checks the whole app (unit tests) are overall 11% faster than before

before:
image

after:
image


mvorisek:
@ibelar can you review this PR

src/Template.php
```diff
@@ -120,24 +102,19 @@ public function isTopTag($tag): bool
...
-        $tag = array_shift($a);
-        //$ref = array_shift($a); // unused
+        $tag = explode('#', $tag, 2)[0];
         if (!isset($this->tags[$tag])) {
-            throw $this->exception('Tag not found in Template')
-                ->addMoreInfo('tag', $tag)
-                ->addMoreInfo('tags', implode(', ', array_keys($this->tags)));
+            throw $this->exceptionAddMoreInfo(
+                (new Exception('Tag not found in Template'))
+                    ->addMoreInfo('tag', $tag)
```
romaninsh: listing existing tags is very helpful in the error message. Could we keep that?

src/Template.php
```diff
@@ -154,51 +131,35 @@ public function &getTagRef($tag)
...
-        $ref = array_shift($a);
-        if (!$ref) {
+        if (strpos($tag, '#') === false) {
             if (!isset($this->tags[$tag])) {
-                throw $this->exception('Tag not found in Template')
-                    ->addMoreInfo('tag', $tag)
-                    ->addMoreInfo('tags', implode(', ', array_keys($this->tags)));
+                throw $this->exceptionAddMoreInfo(
+                    (new Exception('Tag not found in Template'))
+                        ->addMoreInfo('tag', $tag)
```
romaninsh: same as above - tags.

src/Template.php
```diff
@@ -662,23 +610,18 @@ public function tryLoad(string $filename)
...
-        $this->template = $this->tags = [];
-        if (!$str) {
-            return;
-        }
+        $this->tags = [];
+        $this->template = [];
         $this->tag_cnt = [];
 
-        // First expand self-closing tags {$tag} -> {tag}{/tag}
-        $str = preg_replace('/{\$([-_:\w]+)}/', '{\1}{/\1}', $str);
```
romaninsh, line 679: moved to parseTemplate.

src/Template.php
```diff
@@ -689,112 +632,86 @@ public function loadTemplateFromString($str)
...
-     *
-     * @return string unique tag name
+     * Register tag and return unique tag name.
      */
-    protected function regTag($tag)
+    protected function regTag(string $tag): string
     {
         if (!isset($this->tag_cnt[$tag])) {
-            $this->tag_cnt[$tag] = 0;
+            $this->tag_cnt[$tag] = -1;
```
romaninsh: could we keep those at 0 and start tags with #1? may introduce a minor compatibility issue. I have used positional tag references in the past (although long time ago).

src/Template.php
```diff
@@ -689,112 +632,86 @@ public function loadTemplateFromString($str)
...
      * Deploys parse recursion.
-     *
-     * @param string $str
      */
-    protected function parseTemplate($str)
+    protected function parseTemplate(string $str): void
     {
-        $tag = '/{([\/$]?[-_:\w]*[\?]?)}/';
+        $strHash = hash('sha256', $str);
+        if (!isset(self::$_parseTemplateCache[$strHash])) {
```
romaninsh: clever ;)

romaninsh (changes requested):
lets keep tag numbering as-is. Which part of this is BC break?

src/Template.php
mvorisek: `exceptionAddMoreInfo` does that... minor refactoring only.

src/Template.php
mvorisek: `exceptionAddMoreInfo` does that... minor refactoring only.

src/Template.php
mvorisek: yes...

src/Template.php
mvorisek: rebuildTags started on #0 already and we normalize it now to the more programmer way - the numbers after # are just like indexes! @romaninsh please rereview and approve

mvorisek: closing this as this is now internal representaton only

src/Template.php
```diff
@@ -18,191 +18,131 @@ class Template implements \ArrayAccess
...
             return [&$this->template];
         }
 
-        $a = explode('#', $tag);
-        $tag = array_shift($a);
-        $ref = array_shift($a);
-        if (!$ref) {
-            if (!isset($this->tags[$tag])) {
-                throw $this->exception('Tag not found in Template')
+        $tag = explode('#', $tag, 2)[0];
```
DarkSide666, line 132: This code below slightly duplicate code in `getTagRef` method.

src/Template.php
```diff
@@ -521,7 +466,7 @@ public function del($tag)
     public function tryDel($tag)
     {
         if (is_array($tag)) {
-            return $this->del($tag);
+            return $this->tryDel($tag);
```
DarkSide666: This is wrong and will create infinite loop.
Probably should be:
```
if (is_array($tag)) {
    foreach ($tag as $t) {
        $this->tryDel($t);
    }

    return $this;
}
```

src/Template.php
mvorisek: Yes, but we need to access `tagsIndex` first. Later, we should probably store values of template structure in 1D array to fix the rest performance issues. (I expect like +200% faster code)

src/Template.php
```diff
@@ -594,24 +527,22 @@ public function eachTag($tag, $callable)
...
-     * @return self
+     * @return static
      */
-    public function cloneRegion($tag)
+    public function cloneRegion(string $tag)
     {
-        if ($this->isTopTag($tag)) {
-            return clone $this;
+        $template = new static();
+        $template->app = $this->app;
```
DarkSide666, line 539: Is App required for template to work? If it's not mandatory, then this line should be optional.

DarkSide666 (changes requested):
Found 1 bug, but besides that LGTM.

src/Template.php
mvorisek: good catch, fixed

src/Template.php
mvorisek: see Skype, was needed and is needed, if we want to not change other core things (Model or how templates can be used)

romaninsh:
@mvorisek are you planning to revert on the numbering from 0 for better BC?

src/Template.php
```diff
@@ -689,112 +614,87 @@ public function loadTemplateFromString($str)
...
-     * @return string unique tag name
+     * Register tag and return unique tag name.
      */
-    protected function regTag($tag)
+    protected function regTag(string $tag): string
     {
-        if (!isset($this->tag_cnt[$tag])) {
-            $this->tag_cnt[$tag] = 0;
+        if (!isset($this->tagCnt[$tag])) {
+            $this->tagCnt[$tag] = -1;
```
romaninsh, line 630: can we keep this as 0 for BC?

romaninsh (changes requested):

src/Template.php
mvorisek: no longer avaiable to outside, thus irrelevant

DarkSide666 (approved):

PR #1358, opened 2020-07-08
Refactor render methods

mvorisek:
See commit messages for more

Historically, these render methods were not properly renamed and there is a lot of mess in it.

This PR fixes the major issues with minimal breaking impact.

Later, we should merge `render` and `getHtml` methods together.

typical action needed to update your code:
image


DarkSide666 (approved):
LGTM

PR #1359, opened 2020-07-09
[fix] Virtual page rendering

ibelar:
Prevent from Layout element to be output twice.
- The left menu was rendered twice using: https://ui.agiletoolkit.org/demos/interactive/virtual.php?in=callback&__atk_callback=1&p_id=Car


DarkSide666 (approved):
LGTM

PR #1360, opened 2020-07-12
Rename template dirs to match class NS

mvorisek:
No breaking change if you do not use own templates.

for consistent naming, seems like missed in https://github.com/atk4/ui/pull/1311


georgehristov (approved):

georgehristov:
I missed it intentionally but good to line up everything

PR #1361, opened 2020-07-13
Drop App::dbConnect() method

mvorisek:
related with https://github.com/atk4/data/pull/654

`App::dbConnect() method` was not even used anywhere in ui/demos, DB should be created by/added to App like:
```
$app->db = Persistence::connect($dsn, $user, $password, $args);
```


DarkSide666 (approved):
Agree

romaninsh:
Damn - this will blow up ...

PR #1370, opened 2020-07-16
Fix form selector for Lookup with dependency support

mvorisek:
this PR fixes the original selector that was relying on jQuery

bug introduced in https://github.com/atk4/ui/pull/1275

### DIFF
--- a/src/Form/Control/Lookup.php
+++ b/src/Form/Control/Lookup.php
@@ -408,7 +408,7 @@ class Lookup extends Input
 
         if ($this->dependency) {
             $this->apiConfig['data'] = array_merge([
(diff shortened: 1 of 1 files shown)


mvorisek:
@georgehristov there is also another problem - the typed data are not typecasted, I think, we should pass them typecasted, right?

georgehristov (approved):

georgehristov:
Can you give example?

mvorisek:
> 
> 
> Can you give example?

sure:

image

mvorisek:
> 
> 
> Can you give example?

sure:

image

@georgehristov this fix merged, but the typecasting should be solved, in ui/php, we should never operate on raw input data except if we process them raw. This issue seems almost as bug to me, as if `$_GET['form']` data are not presented, the passed data ARE typecasted.

PR #1372, opened 2020-07-21
Add App::jsDirectReload()

mvorisek:
see Discord, do we have enought usecases for this or should it stay per custom usecase only?


mvorisek:
maybe later, but not strong enough usecase for now

PR #1374, opened 2020-07-21
[fix] Modal Element reload

ibelar:
Fix issue when an element inside a modal needs to be reloaded. 
This fix will prevent the `View::getHtml()` method to run into an infinite loop.


src/View.php
```diff
@@ -790,6 +790,8 @@ public function renderToJsonArr(bool $forceReturn = true, $region = null): array
     public function getHtml()
     {
         if (isset($_GET['__atk_reload']) && $_GET['__atk_reload'] === $this->name) {
+            // prevent from looping if this get render again, i.e. part of a Modal View.
+            $_GET['__atk_reload'] = null;
```
mvorisek: `$_GET`, `$_POST`, ... should be always a string by definition and we may rely on this property - fix otherwise ok, but unset it with `unset()` instead

mvorisek: also, can we cover this in tests?

mvorisek: why do you mention `Modal View`, is it Modal specific only?

ibelar: As of now - only Modals are render in app-terminate() if type === 'application/json'. So this should only apply to Modal basically. 
I was not sure if it was better to clear the $_GET argument in App:getRenderedModals() instead, because this is where the issue originated from. wdyt? I have done it in View::getHtml() just in case another use case come up but I doubt it.

mvorisek: > I was not sure if it was better to clear the $_GET argument in App:getRenderedModals() instead, because this is where the issue originated from. wdyt? I have done it in View::getHtml() just in case another use case come up but I doubt it.

Very good point, it seems like related with App specific impl. (use of terminateJson() method in View), so I think moving it there is much cleaner. I moved it and fixed unset.

At least one Behat test should test this.

ibelar: Behat test is added.

mvorisek (approved):
tested, fix fixes and test tests the original issue

PR #1377, opened 2020-07-24
Callback/sticky args refactor

ibelar:
Refactor Callback
 
Post trigger has been remove! Triggering callback is done only via a $_GET argument.

Small BC break:
- No more POST trigger;
- `Callback::triggered()` method was used for two purposes: get callback mode (type) or check if callback was triggered. These functionalities are now split in two separate methods: `Callback::getTriggeredValue()` and `Callback::isTriggered()`


src/Callback.php
```diff
@@ -105,43 +98,29 @@ public function init(): void
...
-    public function terminate()
+    public function terminateJson(View $view = null)
     {
+        if (!$view) {
+            $view = $this->owner;
+        }
+
         if ($this->canTerminate()) {
-            $this->app->terminateJson($this->owner);
+            $this->app->terminateJson($view);
```
georgehristov: Why not
`$this->app->terminateJson($view ?: $this->owner)`?

src/Callback.php
```diff
@@ -155,12 +134,25 @@ protected function canTerminate(): bool
         return !$reload || $this->owner->name === $reload;
     }
 
+    public function isTriggered(): bool
```
mvorisek: 👍 the new api is now much better

src/Callback.php
```diff
@@ -155,12 +134,25 @@ protected function canTerminate(): bool
...
+    {
+        if ($this->isPostTriggered) {
```
mvorisek: I think POST trigger is wrong as we allow nesting for GET only. Can we drop it?

src/ItemsPerPageSelector.php
```diff
@@ -78,7 +78,6 @@ public function jsSetLabel($ipp)
     public function onPageLengthSelect($fx = null)
     {
         if (is_callable($fx)) {
-            if ($this->cb->triggered()) {
                 $this->cb->set(function () use ($fx) {
```
mvorisek: very good!

src/Callback.php
```diff
@@ -155,12 +134,25 @@ protected function canTerminate(): bool
...
+
+        return isset($_GET[$this->urlTrigger]);
+    }
+
     /**
-     * Is callback triggered?
+     * Return this callback mode.
      */
-    public function triggered()
+    public function getMode(): string
```
mvorisek: `getTriggeredMode` or `getTriggeredValue` may be more contextual name

src/Callback.php
mvorisek: yes, but `??`, @georgehristov, please, be strict

georgehristov: Above check is `if (!$view)` therefore my comment for equivalence

src/Modal.php
```diff
@@ -64,6 +61,8 @@ class Modal extends View
 
     /**
      * Set callback function for this modal.
+     * $fx is set as an array in order to comply with View::set().
+     * TODO Rename this function and break BC?
```
georgehristov, line 65: You can rename the method to new name and have the old method calling the new one with deprecation notice. Then it is fully BC

src/Callback.php
```diff
@@ -88,11 +74,18 @@ public function init(): void
...
-            $this->postTrigger = $this->name;
-        }
+        $this->app->stickyGet($this->urlTrigger);
+    }
 
-        $this->appSticky ? $this->app->stickyGet($this->urlTrigger) : $this->owner->stickyGet($this->urlTrigger);
+    public function setUrlTrigger(string $trigger)
+    {
+        $this->urlTrigger = $trigger;
+        $this->app->stickyGet($this->urlTrigger);
```
mvorisek: I belive this concept is wrong, here is why, imagine the following structure:
```
- View elem
- - Form A elem with callback inside (this sets stickyGet)
- Form B elem with callback inside
```

if the last element does something with the URL, it will still add the sticky parameter

I think we should instead introduce sticky parameters per View element and then when we build URL for View, collect sticky parameters from the parents/owners.

src/Callback.php
ibelar: Post triggered should be kept. Form submit rely on it. In theory, we should be able to submit any form, using javascript for example, without worrying about the proper GET param that it's needed. 
What is wrong was including the get parameter when Post triggered is required. I have already changed that.
 It should be the user's responsibility to properly set the Post argument when using post triggering.

src/Callback.php
ibelar: Good point. I have changed it for getTriggeredValue.

src/Callback.php
ibelar: Not sure I follow what you mean. Settings stickyGet to App vs View will only ensure that the get argument will be set in URL, no matter which View required it. Thus ensuring the callback will run.

src/Callback.php
mvorisek: parent callback must always be kept active if child element/View created another one

- with pure POST, this can not be satisfied (as child request may be GET)
- we can keep this but we need to set sticky GET value because of possible nesting (see another comment how)
- then, what is the advantage? Any POST request is GET request + body --> if we use GET only, then we do not need any special/extra logic, just generate right callback url (like for GET) for POST form

src/Callback.php
mvorisek: I named elements in my example.

Now consider this:
- Form A callback is triggered
- and wants to return rendered Form B

now form B will have sticky GET from Form A

That is definitelly wrong - sticky GET should not be global, but per tree item and only self and parent items should contribute to the sticky GET params used for a new URL

src/Modal.php
mvorisek: It is more question if set is good in View...

I would rename `arg2` here now to `$ignore`

and test if instanceof Closure here only

src/Callback.php
mvorisek: double improvement :)

src/JsSortable.php
```diff
@@ -81,15 +81,13 @@ public function init(): void
...
-                    return call_user_func_array($fx, [$sortOrders, $source, $newIdx, $orgIdx]);
-                });
-            }
+            $sortOrders = explode(',', $_POST['order'] ?? '');
+            $source = $_POST['source'] ?? null;
+            $newIdx = $_POST['new_idx'] ?? null;
+            $orgIdx = $_POST['org_idx'] ?? null;
+            $this->set(function () use ($fx, $sortOrders, $source, $newIdx, $orgIdx) {
+                return call_user_func_array($fx, [$sortOrders, $source, $newIdx, $orgIdx]);
+            });
```
georgehristov: To be cleaner here I would propose:
```
    $args = [
            'sortOrders' => explode(',', $_POST['order'] ?? ''),
            'source' => $_POST['source'] ?? null,
            'newIdx' => $_POST['new_idx'] ?? null,
            'orgIdx' => $_POST['org_idx'] ?? null,	
        ];

        $this->set(function () use ($fx, $args) {
            return call_user_func_array($fx, $args);
	});
```
The $args array keys are just for providing extra clarity

I also suggest using uniformly `$args` for the arguments array (instead of `$param` for instance)

src/Form/Control/Multiline.php
```diff
@@ -679,17 +679,15 @@ protected function renderView(): void
...
-                    $this->app->terminateJson(['success' => false, 'error' => $e->getMessage()]);
-                } catch (\Error $e) {
-                    $this->app->terminateJson(['success' => false, 'error' => $e->getMessage()]);
-                }
-            });
-        }
+        $this->cb->set(function () {
+            try {
+                return $this->renderCallback();
+            } catch (\atk4\Core\Exception $e) {
```
mvorisek: catch in one block https://stackoverflow.com/questions/8439581/catching-multiple-exception-types-in-one-catch-block/37522012

src/JsSortable.php
mvorisek: It will not work, because arguments are unpacked in `call_user_func_array`, but I support this idea to update this:
simply declare vars like before, but in the function, no need to declare and use use :) I prefer to keep vars, because it is clear what is passed - and use `call_user_func($v1, $v2, ...)`, no need to pack nad unpack

src/Form/Control/Upload.php
```diff
@@ -230,9 +228,10 @@ protected function renderView(): void
         }
         parent::renderView();
 
-        if (!$this->_isCbRunning && (!$this->hasUploadCb || !$this->hasDeleteCb)) {
+        if (!$this->_isCbRunning && !isset($_GET['__atk_reload']) && (!$this->hasUploadCb || !$this->hasDeleteCb)) {
```
mvorisek: this check is still triggering/throwing in the nested CRUD

- nested CRUD testing demo (you should be able to open/nest CRUD unlimited times and submit it):
https://github.com/atk4/ui/tree/reload_not_working_after_upload
- we should test 3x (main + 2x) nested CRUD in tests

- CRUD with upload (when you select a file and in uploads, then CRUD can not be submitted (it reloads and does not close))
https://github.com/atk4/ui/tree/reload_not_working_after_upload__add_submit_issue

ibelar:
@georgehristov - @mvorisek - Any idea why Behat Crud test is failing since last Data update? 
When in CRUD and search for "united kingdom", then edit the record, Normally after edit the CRUD reload and you should still see the United Kingdom record but it does not. The only change is that the query (_q)  argument is changed from "united kingdom" to "united+kingdom" but this was working before...

src/Callback.php
ibelar: I am not against the idea of keeping GET only. Let's ask @romaninsh what he thinks, as he is the one that created post triggered callback. Maybe it was for a specific reason.

mvorisek:
> last Data update

@ibelar you mean atk4/data update? If yes, than this is not true, the issue is not present on ui develop, this issue is solely in this PR.

ibelar:
> > last Data update
> 
> @ibelar you mean atk4/data update? If yes, than this is not true, the issue is not present on ui develop, this issue is solely in this PR.

Issue is in develop

> > last Data update
> 
> @ibelar you mean atk4/data update? If yes, than this is not true, the issue is not present on ui develop, this issue is solely in this PR.

Nevermind, I see where the error comes from.

src/Form.php
```diff
@@ -681,7 +681,7 @@ public function setFormConfig($config)
      */
     public function ajaxSubmit()
     {
-        $cb = JsCallback::addTo($this);
+        $this->_add($cb = new JsCallback(), ['desired_name' => 'submit']);
```
mvorisek: `add`/`addTo` can not be used?

src/Callback.php
```diff
@@ -105,43 +88,26 @@ public function init(): void
...
-                $this->app->run_called = true;
-                $this->owner->stickyGet($this->urlTrigger);
-                $ret = call_user_func_array($callback, $args);
-                //$this->app->stickyForget($this->name);
-                $this->app->run_called = $t;
-
-                return $ret;
-            }
+        if ($this->isTriggered()) {
+            $this->owner->stickyGet($this->urlTrigger);
```
mvorisek: I think we should not registed sticky GET even inside owner, this can/will be issue if the owner contains two Callbacks and at least one directly

I think we do not need this at all - sticky GET set in self::getUrlArguments()

ibelar: Yes this is need, otherwise triggering element inside a callback will not run, i.e. callback in callback will not work without it.

mvorisek: then we need to move everything related with Url inside View inside a trait or make Callback class extends View :)

> Yes this is need, otherwise triggering element inside a callback will not run, i.e. callback in callback will not work without it.

I looked at the code now and I think it is currently not even possible as URL method names in Callback vs. View are different, I think Callback now expect owner to be an instance of View strictly

src/Callback.php
```diff
@@ -64,13 +63,18 @@ public function init(): void
...
     }
 
     public function setUrlTrigger(string $trigger)
     {
         $this->urlTrigger = $trigger;
+        if ($this->isSticky) {
+            $this->app->stickyGet($this->urlTrigger);
+        }
+
+        $this->owner->stickyGet($this->urlTrigger);
```
mvorisek: probably typo - remove this line

src/Loader.php
```diff
@@ -112,7 +112,7 @@ protected function renderView(): void
     public function jsLoad($args = [], $apiConfig = [], $storeName = null)
     {
         return $this->js()->atkReloadView([
-            'uri' => $this->cb->getJsUrl(),
+            'uri' => $this->cb->getUrl(),
```
mvorisek, line 116: not loaded with JS?

src/Callback.php
```diff
@@ -35,13 +35,12 @@ class Callback
...
-    /**
-     * Specify a custom GET trigger here.
-     *
-     * @var string|null
-     */
+    /** @var string Specify a custom GET trigger. */
     protected $urlTrigger;
 
+    /** @var bool Create app sticky trigger. */
+    public $isSticky = false;
```
mvorisek: I think we should not use this flag and make everything sticky by default

ibelar: I have try with default to all sticky but then other problems come up. For instance, when running crud test, delete is not trigger after edit because of that.

src/Loader.php
ibelar: Terminate with js but uses Callback and not JsCallback.

src/Callback.php
ibelar: thks. done!

src/Form.php
ibelar: Done!

src/Loader.php
mvorisek: you are right

src/Callback.php
mvorisek: Hmm, I think this will not allow proper/really abstract nesting - we should never set an app/global sticky GET from Callback. This is where the
> I have try with default to all sticky but then other problems come up. For instance, when running crud test, delete is not trigger after edit because of that.

issues come from.

Not even parent/owner sticky GET.

src/Callback.php
```diff
@@ -105,88 +91,85 @@ public function init(): void
...
+    public function canTerminate(): bool
+    {
+        return $this->urlTrigger === ($_GET['__atk_callback'] ?? null);
+    }
+
+    /**
+     * Allow callback to be triggered or not.
      */
-    protected function canTerminate(): bool
+    public function canTrigger(): bool
```
mvorisek: very hard to understand what is the difference between isTrigger, from it's name

I highly prefer no `canTrigger` method - can we move it to `isTriggered`?

src/Form/Control/Upload.php
```diff
@@ -228,8 +228,14 @@ protected function renderView(): void
...
         parent::renderView();
 
-        if (!$this->_isCbRunning && !isset($_GET['__atk_reload']) && (!$this->hasUploadCb || !$this->hasDeleteCb)) {
-            throw new Exception('onUpload and onDelete callback must be called to use file upload. Missing one or both of them.');
+        if ($this->cb->canTerminate()) {
+            $action = $_POST['action'] ?? null;
+            if (!$this->hasUploadCb && ($action === self::UPLOAD_ACTION)) {
+                throw new Exception('Missing onUpload callback.');
+            }
+            if (!$this->hasDeleteCb && ($action === self::DELETE_ACTION)) {
```
mvorisek: elseif, otherwise I like this

src/Wizard.php
```diff
@@ -67,21 +67,21 @@ public function init(): void
             $this->stepCallback = Callback::addTo($this, ['urlTrigger' => $this->name]);
         }
 
-        $this->currentStep = (int) $this->stepCallback->triggered() ?: 0;
+        $this->currentStep = (int) $this->stepCallback->getTriggeredValue() ?: 0;
```
mvorisek: it should be more like `(int) ($this->stepCallback->getTriggeredValue() ?: 0)`

src/Callback.php
```diff
@@ -105,88 +94,88 @@ public function init(): void
...
+        return $this->owner->url($this->getUrlArguments($value));
+    }
+
+    /**
+     * Return proper url argument for this callback.
+     */
+    private function getUrlArguments(string $value): array
     {
-        return $this->owner->url([$this->urlTrigger => $mode, '__atk_callback' => 1], (bool) $this->postTrigger);
+        return ['__atk_callback' => $this->urlTrigger, $this->urlTrigger => $value];
```
mvorisek, line 175: Setting value of `__atk_callback` implies last callback in the render tree, right?

It is used in `canTerminate()`. I think it will fail to reload callback containin another callback.

Previously, there was:
```
        $reload = $_GET['__atk_reload'] ?? null;
        return $this->urlTrigger === ($_GET['__atk_callback'] ?? null);
```
I think we need to keep the `__atk_reload` check in `canTerminate()`.

src/Modal.php
mvorisek: resolved

src/JsSortable.php
mvorisek: resolved

src/Form/Control/Upload.php
mvorisek: resolved

src/Wizard.php
mvorisek: resolved

src/Callback.php
mvorisek: `canTrigger` is now excl. `isTriggered()`, resolved

ibelar:
@mvorisek - Since your last changed, File upload demo (/demos/form-control/upload.php) and Wizard demo step 2 (/demos/interactive/wizard.php) are not working anymore. Please fix.

src/Callback.php
```diff
@@ -81,18 +65,40 @@ public function init(): void
...
+        $this->urlTrigger = $trigger ?? $this->name;
+    }
 
-        $this->appSticky ? $this->app->stickyGet($this->urlTrigger) : $this->owner->stickyGet($this->urlTrigger);
+    public function getUrlTrigger(): string
+    {
+        return $this->urlTrigger;
+    }
+
+    private function callWithAppStickyTrigger(\Closure $fx)
```
ibelar: Why do we need this? All was working perfectly fine before...

mvorisek: not, it was not - the described reload issue on Discord is not resolved (and even this does not solve it), I belive, it never will with app/global sticky args approach

This should be the same as before, but only the app/global sticky args are set when the callback is executing - once it is done, it is no longer app/global sticky - we need to converge to a solution without app/global sticky at all

mvorisek:
- /demos/form-control/upload.php

fixed, support of multiple files already in https://github.com/atk4/ui/pull/1380

- /demos/interactive/wizard.php (Wizard demo step 2)

yes, this issue is related with https://github.com/atk4/ui/pull/1377/commits/b459488ed880c4f3da8a35700cf459086f82cb8b , can you look at it?

src/Callback.php
ibelar: I do not see any issue using your branch describe on Discord. All was working fine for me locally. I can even click 'Add Country" three time and Add will still work using the third modal.

mvorisek: https://discord.com/channels/657649422280425523/664534590160896006/737766184044199947

how to reproduce https://discord.com/channels/657649422280425523/664534590160896006/738128513412104303

ibelar:
@mvorisek - I revert Callback to the last working state. Let's merge this and we can look into your issue after.

mvorisek:
> 
> 
> @mvorisek - I revert Callback to the last working state. Let's merge this and we can look into your issue after.

I fixed the revert to revert only the too strict code

I belive these global sticky args must be removed, so I opened https://github.com/atk4/ui/pull/1377#issuecomment-666393890 .

> we should test 3x (main + 2x) nested CRUD in tests

- please add some nested tests to this PR and then let's merge it
- please also add test for Wizard demo (which will fail if we revert the revert)

demos/_unit-test/callback_2.php
```diff
@@ -0,0 +1,46 @@
...
+/** @var \atk4\ui\App $app */
+require_once __DIR__ . '/../init-app.php';
+
+$m = (new CountryLock($app->db))->setLimit(5);
+$m->getUserAction('edit')->ui['button'] = new Button(['Edit', ['ui' => 'atk-test button']]);
+
+$loader = Loader::addTo($app);
+$loader->loadEvent = false;
+
+$loader->set(function($p) use ($m) {
```
mvorisek: should be CRUD (as it uses Popup) and probably it should be defined with function, which can be called unlimited times

ibelar: This test is failing using the current develop branch. So should be enough for now.
But feel free to supply your own sample.

mvorisek: basically in `reload_not_working_after_upload__add_submit_issue` branch, `addCrudDemo`, can be very simple, I even do not think we need upload, but it will be nice if crud (like there) can be used for input for field, this is quite complex test which will be needed if we will drop sticky global later...

the field should be used twice and both fields should be edited twice (to simulate issues with with sticky...)

mvorisek (approved):

PR #1380, opened 2020-07-25
Allow upload multiple

mvorisek:
this PR adds support to multiple files to be uploaded, but only uploaded - the Upload control still remains for one file only

original fx signatures remains - now only more files can be passed

marking as BC breaks as no special "error" string param is passed anymore

### DIFF
--- a/demos/form-control/upload.php
+++ b/demos/form-control/upload.php
@@ -30,13 +30,13 @@ $img->onDelete(function ($fileId) use ($img) {
     ]);
 });
 
-$img->onUpload(function ($files) use ($form, $img) {
-    if ($files === 'error') {
+$img->onUpload(function ($postFile) use ($form, $img) {
+    if ($postFile['error'] !== 0) {
         return $form->error('img', 'Error uploading image.');
     }
 
     $img->setThumbnailSrc('./images/logo.png');
-    $img->set('123456', $files['name'] . ' (token: 123456)');
+    $img->set('123456', $postFile['name'] . ' (token: 123456)');
 
     //Do file processing here...
 
--- a/src/Form/Control/Upload.php
+++ b/src/Form/Control/Upload.php
@@ -171,7 +171,7 @@ class Upload extends Input
                 $this->_isCbRunning = true;
 
                 $postFiles = [];
-                for ($i = 0;; $i++) {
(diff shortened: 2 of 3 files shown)


src/Form/Control/Upload.php
```diff
@@ -167,28 +167,36 @@ public function onUpload(\Closure $fx)
...
+                $files = [];
+                for ($i = 0;; $i++) {
+                    $k = 'file' . ($i > 0 ? '-' . $i : '');
+                    if (!isset($_FILES[$k])) {
+                        break;
+                    }
+                    $files[] = $_FILES[$k];
+                }
+
+                if (count($files) > 0) {
```
mvorisek: remains for 1 file only like before and the whole control component

demos/form-control/upload.php
```diff
@@ -30,13 +30,13 @@
     ]);
 });
 
-$img->onUpload(function ($files) use ($form, $img) {
-    if ($files === 'error') {
+$img->onUpload(function ($postFile) use ($form, $img) {
+    if ($postFile['error'] !== 0) {
```
mvorisek, line 34: we pass now in native php FILES format, also if error

src/Form/Control/Upload.php
```diff
@@ -161,54 +161,65 @@ public function addJsAction($action)
...
+    public function onUpload(\Closure $fx)
     {
-        $this->hasDeleteCb = true;
-        $action = $_POST['action'] ?? null;
-        if ($this->cb->triggered() && $action === 'delete') {
-            $this->_isCbRunning = true;
-            $fileName = $_POST['f_name'] ?? null;
-            $this->cb->set(function () use ($fx, $fileName) {
-                $this->addJsAction($fx($fileName));
+        $this->hasUploadCb = true;
```
georgehristov, line 168: $this->hasUploadFx ?
But only if we can refactor $this->cb to $this->fx (BC)
To keep all naming consistent

georgehristov (approved):

src/Form/Control/Upload.php
mvorisek: it is related with Callback, lets fix this in https://github.com/atk4/ui/pull/1377

PR #1381, opened 2020-07-26
[fix] Build Query function

ibelar:
- fix http_build_query function
   now return %20 instead of + when a space is part of the argument


DarkSide666 (approved):
LGTM

PR #1384, opened 2020-07-30
Callable to closure refactor

mvorisek:
You must always pass a proper Closure instead of callable

### DIFF
--- a/src/Component/InlineEdit.php
+++ b/src/Component/InlineEdit.php
@@ -120,7 +120,7 @@ class InlineEdit extends View
                         $this->app->terminateJson([
                             'success' => true,
                             'hasValidationError' => true,
-                            'atkjs' => $this->jsError(call_user_func($this->formatErrorMsg, $e, $value))->jsRender(),
+                            'atkjs' => $this->jsError(($this->formatErrorMsg)($e, $value))->jsRender(),
                         ]);
                     }
                 });
@@ -141,7 +141,7 @@ class InlineEdit extends View
         if (!$this->autoSave) {
             $value = $_POST['value'] ?? null;
             $this->cb->set(function () use ($fx, $value) {
-                return call_user_func($fx, $value);
+                return $fx($value);
             });
         }
     }
--- a/src/Component/InlineEdit.php
+++ b/src/Component/InlineEdit.php
@@ -76,7 +76,7 @@ class InlineEdit extends View
      * A default one is supply if this is null.
      * It receive the error ($e) as parameter.
      *
-     * @var callable|null
+     * @var \Closure|null
      */
     public $formatErrorMsg;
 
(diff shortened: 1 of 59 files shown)


src/Component/InlineEdit.php
```diff
@@ -134,19 +134,15 @@ public function setModel(\atk4\data\Model $model)
...
-            if ($this->cb->triggered()) {
-                $value = $_POST['value'] ? $_POST['value'] : null;
-                $this->cb->set(function () use ($fx, $value) {
-                    return call_user_func($fx, $value);
-                });
-            }
+        if (!$this->autoSave) {
+            $value = $_POST['value'] ?? null;
+            $this->cb->set(function () use ($fx, $value) {
+                return call_user_func($fx, $value);
```
georgehristov: `return $fx($value);`?

src/Form/Control/DropdownCascade.php
```diff
@@ -65,7 +65,7 @@ public function getNewValues(string $id): array
         $model = $this->cascadeControl->model->load($id)->ref($this->reference);
         $values = [];
         foreach ($model as $k => $row) {
-            if ($this->renderRowFunction && is_callable($this->renderRowFunction)) {
+            if ($this->renderRowFunction) {
                 $res = call_user_func($this->renderRowFunction, $row, $k);
```
georgehristov: `$res = $this->renderRowFunction($row, $k);`?

src/Form/Control/Lookup.php
```diff
@@ -202,7 +202,7 @@ public function getData($limit = true): array
      */
     public function renderRow(\atk4\data\Model $row): array
     {
-        $renderRowFunction = is_callable($this->renderRowFunction) ? $this->renderRowFunction : [__CLASS__, 'defaultRenderRow'];
+        $renderRowFunction = $this->renderRowFunction ?? \Closure::fromCallable([static::class, 'defaultRenderRow']);
 
         return call_user_func($renderRowFunction, $this, $row);
```
georgehristov: `return $renderRowFunction($this, $row);`?

georgehristov (approved):
👍

src/Component/InlineEdit.php
mvorisek: I would normally write it like that, but on the other side `call_user_func*` is easier to search. But yes, lets never use call_user_func :)

src/Form/Control/Lookup.php
mvorisek: I prefer to pass args only once to be consistent

src/Form/Control/DropdownCascade.php
mvorisek: https://github.com/atk4/ui/pull/1384#discussion_r462926547

src/Component/InlineEdit.php
mvorisek: done, yes, it is better

PR #1386, opened 2020-07-30
Use native and/or in Vue Scope builder internally

mvorisek:
also, if possible, internal representation (even in browser) should never be stored in text (like "equals" vs "=")


georgehristov (approved):
LGTM

PR #1388, opened 2020-07-31
Fix code for Model::action to Model::toQuery refactor

mvorisek:
merge after https://github.com/atk4/data/pull/688


demos/collection/multitable.php
```diff
@@ -78,7 +78,7 @@ public function setModel(\atk4\data\Model $model, $route = [])
 \atk4\ui\Header::addTo($app, ['MacOS File Finder', 'subHeader' => 'Component built around Table, Columns and JsReload']);
 
 $vp = \atk4\ui\VirtualPage::addTo($app)->set(function ($vp) use ($model) {
-    $model->action('delete')->execute();
+    $model->toQuery('delete')->execute();
```
mvorisek, line 81: is this even correct use?

PR #1389, opened 2020-07-31
Add Behat testing with Firefox

mvorisek:


.travis.yml
```diff
@@ -65,7 +65,7 @@ script:
     #- mkdir -p coverage
     #- chmod 777 coverage
     #- cp tools/coverage.php demos/coverage.php
-    - vendor/bin/behat --config behat-travis.yml || exit -1
+    - vendor/bin/behat --config tests/behat-travis.yml || exit -1
```
georgehristov: tests/behat/travis.yml?

mvorisek: I do not like too much files in root directory, but I think it should not go to tests/behat, it is config... Let's clear this on Discord.

.travis.yml
```diff
@@ -65,7 +65,7 @@ script:
     #- mkdir -p coverage
     #- chmod 777 coverage
     #- cp tools/coverage.php demos/coverage.php
-    - vendor/bin/behat --config behat-travis.yml || exit -1
+    - vendor/bin/behat --config behat.yml.travis || exit -1
```
mvorisek, line 68: so you can/should create `behat.yml.local`

georgehristov (approved):
👍

PR #1392, opened 2020-08-03
Improve Behat testing and handle all waits using hooks

mvorisek:
we should disable all CSS/JS animation for Behat testing


mvorisek:
we can/should probably disable all animation for testing:

- https://api.jquery.com/jQuery.fx.off/
- https://stackoverflow.com/questions/11131875/what-is-the-cleanest-way-to-disable-css-transition-effects-temporarily

georgehristov (approved):
super!

DarkSide666 (approved):

PR #1396, opened 2020-08-05
Add AbstractView and use it as a base class for Callback

mvorisek:
small break: `View::add()` now checks instanceof AbstractView

### DIFF
--- a/src/Callback.php
+++ b/src/Callback.php
@@ -147,4 +147,12 @@ class Callback extends View
     {
         return ['__atk_callback' => $this->urlTrigger, $this->urlTrigger => $value];
     }
+
+    protected function _getStickyArgs($triggerBy): array
+    {
+        // DEV NOTE:
+        // - getUrlArguments $value used only in https://github.com/atk4/ui/blob/08644a685a9ee07b4e94d1e35e3bd3c95b7a013d/src/VirtualPage.php#L134
+        // - $_GET['__atk_callback'] from getUrlArguments seems to control terminating behaviour!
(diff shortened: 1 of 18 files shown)


src/Callback.php
```diff
@@ -174,4 +147,12 @@ private function getUrlArguments(string $value): array
     {
         return ['__atk_callback' => $this->urlTrigger, $this->urlTrigger => $value];
     }
+
+    protected function _getStickyArgs($triggerBy): array
+    {
+        // DEV NOTE:
```
mvorisek, line 153: will be solved in another/breaking PR later...

src/CallbackLater.php
```diff
@@ -21,7 +21,7 @@ class CallbackLater extends Callback
      *
      * @return mixed|null
      */
-    public function set($callback, $args = [])
+    public function set($callback = null, $args = null)
```
georgehristov: $fx

georgehristov (commented):
I like the idea but:
- what is the breaking change?
- what is the benefit?

src/CallbackLater.php
mvorisek: >     * what is the breaking change?

image


>     * what is the benefit?

strict item tree type validation and later abstract nesting for Callback

mvorisek: `$fx` renamed

georgehristov (approved):
Fix coding style and I would say we can merge

Will this solve the problem of repeated reloads when using callback to reload a view?

mvorisek:
> Will this solve the problem of repeated reloads when using callback to reload a view?

Quite a long jorney to it, this PR is for about 20% work needed :)

PR #1397, opened 2020-08-05
No app/global sticky Callback

### LINKED ISSUE #1385: Sticky global must apply to self and subtree only
issues to fix described in already merged https://github.com/atk4/ui/pull/1377

I find out that https://github.com/atk4/ui/pull/1377#discussion_r463019668 issue depends somehow if the field is required or not.

When field is required - CRUD form does reload (/wo error), when not required, form submits as expected.

mvorisek:
fixes: #1385

BC break: GET sticky args are no longer globally sticky, they are sticky only within the element where defined and its subtree

still semi-global, as cb sticky args are usually used for owning component

### DIFF
--- a/src/View.php
+++ b/src/View.php
@@ -1061,7 +1061,9 @@ class View extends AbstractView implements JsExpressionable
             if ($ex instanceof self && $ex instanceof UserAction\JsExecutorInterface) {
                 //Executor may already had been add to layout. Like in CardDeck.
                 if (!isset($this->app->html->elements[$ex->short_name])) {
-                    $ex->viewForUrl = $this; // very dirty hack
+                    // very dirty hack, @TODO, attach modals in the standard render tree
+                    // but only render the result to a different place/html DOM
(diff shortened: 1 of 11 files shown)


georgehristov (approved):
LGTM

Please list in description the BC-break

src/View.php
```diff
@@ -1061,6 +1061,9 @@ public function on($event, $selector = null, $action = null, $defaults = null)
             if ($ex instanceof self && $ex instanceof UserAction\JsExecutorInterface) {
                 //Executor may already had been add to layout. Like in CardDeck.
                 if (!isset($this->app->html->elements[$ex->short_name])) {
+                    // very dirty hack, @TODO, attach modals in the standard render tree
+                    // but only render the result to a different place/html DOM
+                    $ex->viewForUrl = $this;
```
mvorisek, line 1066: @ibelar I was able to finish this PR with this hack, but do you think we can/should add Modal like any other View to a standard render tree and then render like now?

PR #1398, opened 2020-08-05
[fix] check if any data recalled by FilterModel

georgehristov:


src/Table/Column/FilterModel/TypeEnum.php
```diff
@@ -30,15 +30,16 @@ public function init(): void
 
     public function setConditionForModel($model)
     {
-        $filter = $this->recallData();
-        $values = [];
-        foreach ($filter as $key => $isSet) {
-            if ($isSet === true) {
-                $values[] = $key;
+        if ($filter = $this->recallData()) {
```
mvorisek, line 33: `[]` as default in FIlterMode::recallData?

georgehristov: I was considering both options but this seems more robust I believe

mvorisek: it seems fine, other checks looks like:
```

        $filter = $this->recallData();
        if (isset($filter['id'])) {
```

why can not we use the same?

mvorisek: FIlterMode::recallData should definitely be declared with return type `?array`

mvorisek (approved):
see comments

src/Table/Column/FilterModel/TypeEnum.php
georgehristov: Maybe we do all 3, would not hurt.
And declare :array as return type.
Then it is robust :)

mvorisek: if [] is good value for "not set" :) (not sure about this, probably not)

georgehristov: `[]` is okay for this `isset($filter['id'])`

PR #1401, opened 2020-08-06
Replace table column Link useLabel with label

georgehristov:
allow to set custom label template

### DIFF
--- a/src/Table/Column/Link.php
+++ b/src/Table/Column/Link.php
@@ -61,7 +61,7 @@ class Link extends Table\Column
      */
     public $args = [];
 
-    /** @var bool use value as label of the link */
+    /** @var bool|string use value as label of the link */
     public $useLabel = true;
 
     /** @var string|null set element class. */
@@ -128,7 +128,11 @@ class Link extends Table\Column
 
         $label = '';
         if ($this->useLabel) {
-            $label = $field ? ('{$' . $field->shortName . '}') : '[Link]';
+            if ($this->useLabel === true) {
+                $label = $field ? ('{$' . $field->shortName . '}') : '[Link]';
+            } else {
+                $label = $this->useLabel;
+            }
         }
 
         if ($this->class) {
--- a/src/Table/Column/Link.php
+++ b/src/Table/Column/Link.php
@@ -61,8 +61,19 @@ class Link extends Table\Column
      */
     public $args = [];
 
-    /** @var bool|string use value as label of the link */
+    /** @deprecated use $label instead - will be removed dec-2020 */
     public $useLabel = true;
+
+    /**
+     * Label template to use on the link.
+     *
+     * $label = true - use field value as label
+     * $label = false - no label
+     * $label = 'string {$field}' - use custom template provided
+     *
+     * @var bool|string
+     * */
+    public $label = true;
 
     /** @var string|null set element class. */
     public $class;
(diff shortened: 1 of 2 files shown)


src/Table/Column/Link.php
```diff
@@ -55,7 +55,7 @@ class Link extends Table\Column
      */
     public $args = [];
 
-    /** @var bool use value as label of the link */
+    /** @var bool|string use value as label of the link */
     public $use_label = true;
```
mvorisek: no string for "use_xxx"

georgehristov: I would normally propose using $label.
`$label = false` - no label
`$label = true` - default label template
`$label = 'string'` - use it as template

_(It works pretty nice in the new filestore Upload Control having such feature)_

mvorisek: yes

src/Table/Column/Link.php
```diff
@@ -136,7 +136,11 @@ public function getDataCellTemplate(\atk4\data\Field $field = null)
 
         $label = '';
         if ($this->use_label) {
-            $label = $field ? ('{$' . $field->short_name . '}') : '[Link]';
+            if ($this->use_label === true) {
+                $label = $field ? ('{$' . $field->short_name . '}') : '[Link]';
```
mvorisek: `{$` is there for some templating?

georgehristov: yes, $label is a template string

src/Table/Column/Link.php
```diff
@@ -55,9 +55,20 @@ class Link extends Table\Column
...
     public $args = [];
 
-    /** @var bool|string use value as label of the link */
+    /** @deprecated use $label instead - will be removed dec-2020 */
     public $use_label = true;
 
+    /**
+     * Label template to use on the link.
+     *
+     * $label = true - use field value as label
```
mvorisek, line 64: implied from type!

src/Table/Column/Link.php
mvorisek: then it is set wrong - label can be be used as template...

georgehristov: I didn't get your comment?

mvorisek: imagine label with `{$xx}`

georgehristov: This is true for any decorator, then you need to escape the template like syntax

mvorisek: and html? `<script>alert("hi, my Bitcoin wallet is...");</script>`

georgehristov: scripts gets directly executed on each row render

mvorisek: I really hate these single purpose solution and even more if they are not complete/secure at all.

image

can we solve it better? is this for some core atk4 feature?

georgehristov: The additional code is due to BC notices, etc. Actually additional code is 1 line.
It provides much more flexible functionality for the Link where dev can set whatever template he likes, not just the field value

mvorisek: but implemented as single purpose if - should we copy paste that 100x times?!?

mvorisek:
In `Card::useLabel` and `CardDeck::useLabel` are simillar properties. I am still not fully sold on this, especially as the (newly introduced) `Link::label` requires a template, which requires a `$fieldShortName` - do we need/want to allow to link across other field values?

@georgehristov please provide some usecases

mvorisek:
Currently this PR is missing a test, not encoding HTML in the label and it is a questionable if this is even wanted, as the label currently represents the value itself.

As no feedback/usecases were provided, I have to close this PR.

PR #1402, opened 2020-08-07
[fix] closure variable name

georgehristov:


mvorisek (approved):
good catch, probably wrongly resolved conflict...

PR #1404, opened 2020-08-07
Improve Behat testing - always fail if page contains uncaught exception

mvorisek:
some tests test for text but can currently succeed even if exception is displayed


georgehristov (approved):
👍

PR #1412, opened 2020-08-17
Render full exception details on ModalExecutor error

mvorisek:
Fix missing info error info

We still want probably fix https://github.com/atk4/ui/pull/1328 as the Modal can still render in uncontrolled/buggy fashion, very dangerous to existing data as the modal can end up half initialized


georgehristov (approved):
👍

PR #1413, opened 2020-08-17
Improve demo setup docs

bedengler:
The atk4/ui demo now runs also if you position it outside the atk4 repo and rename the "demos" directory to something else. This should add a layer of positive user experience, if someone tries that out and it works flawlessly.

In addition I have added documentation for & link to the atk4/ui demos.


docs/overview.rst
```diff
@@ -451,3 +451,51 @@ infrastructure, Linux and SSH. A good place to start is Heroku (https://www.hero
 We reference Heroku in our tutorials, but Agile Toolkit can work with any cloud
 hosting that runs PHP apps.
 
+
+Online-Demo showing many functions
```
mvorisek: this is ok if not written somewhere else alredy

docs/overview.rst
```diff
@@ -451,3 +451,51 @@ infrastructure, Linux and SSH. A good place to start is Heroku (https://www.hero
...
+There's a demo available of atk4/ui which shows many of the modules and functions available
+in atk4.
+You can watch & use the source code of each example to find best practice examples and to see
+how to use atk4 in certain application cases.
+
+You can find the demo here:
+https://agile-ui.readthedocs.io/en/latest/index.html
+
+
+Demo contained in the atk4 repository
```
mvorisek: text below is quite specifically worded...

there should be:
- advise demos need clone/fork
- how sqlite database can be created: `php demos/_demo-data/create-sqlite-db.php`

bedengler: According to @georgehristov they are adviced to don't move the original repos and use them as they are.
What do we advice for consistency?  

1. clone just the "demos" directory or
2. clone the entire atk4/ui directory or
3. don't clone and use the files in /vendors/atk4/ui...

???

bedengler: @mvorisek, I added SQLITE but didn't commit. I'm waiting for feedback for the cloning issue...

mvorisek: fork whole atk4/ui + composer install + init sqlite db, then make whole atk4/ui accessible by webserver

bedengler: What about moving just the "demos" directory in a subdirectory called "atk4/ui" on the webserver that's publicly available instead of the whole atk4/ui directory? 
Advantage: when updating atk4/ui, you have the demos separated and they aren't overwritten.

mvorisek: impossible - you need to fork the whole ui

for separate vendor files, you need to understand composer more deeply and I think this should be not documented as too advanced for most users...

docs/overview.rst
bedengler: I didn't find it anywhere. Even if it's anywhere else, it should be placed in the overview (which is the first place when  someone is new to atk4 and opens the documentation).
And it should be removed from "anywhere else", if someone could find it.

docs/overview.rst
bedengler: adapted, please check

demos/init-app.php
```diff
@@ -51,8 +51,13 @@
     throw new \atk4\ui\Exception('Database error: ' . $e->getMessage());
 }
 
-[$rootUrl, $relUrl] = preg_split('~(?<=/)(?=demos(/|\?|$))|\?~s', $_SERVER['REQUEST_URI'], 3);
-$demosUrl = $rootUrl . 'demos/';
+// Current sub directory of webserver - makes it independent where to run the demo from
+// basically we want to run it just from within the ui repo, but this adds a layer of
+// better user experience, when a user tries to copy it elsewhere and it works
+$currDir = substr(str_replace($_SERVER['DOCUMENT_ROOT'], "", dirname(__FILE__)), 1);
```
mvorisek: I belive should be not supported, we should actually not distribute tests/demos in release at all for security reasons.

docs/overview.rst
```diff
@@ -451,3 +451,71 @@ infrastructure, Linux and SSH. A good place to start is Heroku (https://www.hero
...
+ - Setup the database using the file "dump.sql" from the subdirectory "atk4/ui/demos/_demo-data" of the "demos" directory. You can also setup an SQLITE database using a provided script (see below)
+ - Open the demos from your browser (e.g. with https://localhost/vendors/atk4/ui/demos or if you choose to clone it, with e.g. https://localhost/atk4/ui/demos)
+
+Setup SQLITE database
+---------------------
+The demo also includes a script that let's you setup a SQLITE database. You will find this script in the subdirectory
+"atk4/ui/demos/_demo-data/".
+Call it as follows (you may have to adapt the paths to fit your webserver structure)::
+
+    php atk4/ui/demos/_demo-data/create-sqlite-db.php;  // when using the CLI approach
```
mvorisek: we should allow to create from CLI only

docs/overview.rst
```diff
@@ -451,3 +451,71 @@ infrastructure, Linux and SSH. A good place to start is Heroku (https://www.hero
...
+
+or::
+
+    https://localhost/atk4/ui/demos/_demo-data/create-sqlite-db.php;  // when calling in the browser
+
+
+Additional tips
+---------------
+
+ - If you are using Eclipse, import the previously cloned "atk4" directory into Eclipse (then it finds all sub-projects)
```
mvorisek: no specific doc for Eclipse

demos/init-app.php
bedengler: I'm seeing it from the perspective as a user (I for myself joined atk4 a year ago). 
The learning curve is ridiculous steep as there's a big lack of documentation. Reading everything out of the code without having examples makes everything even more time consuming. Which is unnecessary imo, if we just 
1) offer more examples
2) write documentations
3) make it as easy as possible to get results (raise usability)

Currently you have to try out a lot. 
If you just spend hours or days of getting a demo running, users jump off again before they even were able to see a welcome screen...

Therefore I'd be interested in opinions of other team members first

mvorisek: please post usecase (step by step) that needs this change

mvorisek:
@bedengler revert `demos/init-app.php`, then we @ibelar or @georgehristov can review or we can close completely

demos/init-app.php
romaninsh: I think this is a positive change and it does not break anything. Sound reasoning on moving things outside.

docs/overview.rst
```diff
@@ -451,3 +451,71 @@ infrastructure, Linux and SSH. A good place to start is Heroku (https://www.hero
...
+https://agile-ui.readthedocs.io/en/latest/index.html
+
+
+Demo contained in the atk4 repository
+=====================================
+When you download and install atk4 you will find a subdirectory called "demos" in the atk4
+repository which also could be locally executed.
+This folder is basically intended to run under the UI package only.
+That means the demo may not work properly, if you copy the "demos" directory out of this location
+(to run it from a different location) or if the "vendors" directory isn't accessible through the
```
romaninsh: I thought the above fix addresses that? 

Could you rephrase this text in the form of step-by-step instruction on how to actually run demo elsewhere?

docs/overview.rst
romaninsh: it might be more confusing to offer alternative implementation, but if your opinion as someone new to ATK that both instruction should be provided -that's OK.. 

Just consider file permission implications. 12-factor app requires applications to be runnable in read-only containers.

Same can be solved with docker-compose and mysql container, btw.

docs/overview.rst
romaninsh: in my case directory is always called "ui", not atk4.

romaninsh:
@bedengler - thanks so much for not only identifying documentation issue, but also be willing to improve the situation. Framework relies on contributors and we need more people focused specifically on the docs, so if you are serious about this - you would be able to refine documentation gradually.

Overall - your instruction seem very "specific" to your personal use case, can require a bit more work, but having this much is better than having nothing at all, so well done!

demos/init-app.php
mvorisek: > 
> 
> please post usecase (step by step) that needs this change

as written, I need usecase that needs this change, otherwise `str_replace($_SERVER['DOCUMENT_ROOT'], "", dirname(__FILE__)` seems super hacky to me and it will not work on even slighly badly configured servers!

mvorisek (changes requested):
need changes

mvorisek:
@bedengler Please proof read the changes, remove what is not needed/up-to-date and we can merge then.

abbadon1334 (approved):
LGTM

mvorisek (approved):

PR #1424, opened 2020-08-23
Build and diff CSS in GH CI + fix sync

mvorisek:
CSS vs. LESS was not in sync, LESS was not compiled after https://github.com/atk4/ui/pull/1269


georgehristov (approved):
LGTM

PR #1425, opened 2020-08-23
Add possibility to change Template class in loadTemplate; clean up App

PhilippGrashoff:
Introduced $app->templateClass property which defaults to \atk4\ui\Template::class. Change this to make App::loadTemplate return a different class instance, e.g. \Your\Extended\Template

cleaned up App class a bit:
- move two properties which were in between methods to the property block
- changed default template dir construction
- moved code from __construct() to new method setupTemplateDirs() to make __construct less long.


src/App.php
```diff
@@ -214,6 +226,20 @@ static function ($severity, $msg, $file, $line) {
         }
     }
 
+    /**
+     *
```
mvorisek: remove this empty comment

src/App.php
```diff
@@ -147,6 +147,25 @@ class App
...
+    /**
+     * @var array global sticky arguments
+     */
+    protected $sticky_get_arguments = [
+        '__atk_json' => false,
+        '__atk_tab' => false,
+    ];
+
+    /**
+     * @var string The template class to use in loadTemplate
```
mvorisek: no need for comment, user can find usage in IDE very quickly

src/App.php
```diff
@@ -214,6 +226,20 @@ static function ($severity, $msg, $file, $line) {
...
+     */
+    protected function setupTemplateDirs() {
+        // Set up template folder
+        if ($this->template_dir === null) {
+            $this->template_dir = [];
+        } elseif (!is_array($this->template_dir)) {
+            $this->template_dir = [$this->template_dir];
+        }
+
+        $this->template_dir[] = dirname(__DIR__) .  DIRECTORY_SEPARATOR . 'template' . DIRECTORY_SEPARATOR . $this->skin;
```
mvorisek: no need for dir separator, `/` is fine

tests/AppUsesTemplateClassPropertyTest.php
```diff
@@ -0,0 +1,44 @@
...
+
+use atk4\core\AtkPhpunit;
+use atk4\ui\App;
+use atk4\ui\Template;
+
+
+/**
+ * Test if property templateClass is used correctly in loadTemplate()
+ */
+class AppUsesTemplateClassPropertyTest extends AtkPhpunit\TestCase
```
mvorisek: I prefer this test in App tests

tests/AppUsesTemplateClassPropertyTest.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    /**
+     * Test if default setting works correctly
```
mvorisek: comment not needed

tests/AppUsesTemplateClassPropertyTest.php
PhilippGrashoff: There is no AppTest class yet, so I created this one. Was quite surprised to not find any AppTest class.

mvorisek: I renamed your test class, so everyone can add additional tests easily

mvorisek (approved):

PR #1426, opened 2020-08-24
Enforce industry standard Airbnb JS Coding Standards

mvorisek:


tests-behat/basicexecutor.feature
```diff
@@ -1,3 +1,4 @@
+@basic-executor
```
mvorisek: just curious, what is this for?

js/src/services/api.service.js
```diff
@@ -27,11 +27,10 @@ class ApiService {
    * will work just fine, even if $ is not assign globally.
    *
    * @param code // javascript to be eval.
-   * @param jQueryRef // reference to jQuery.
+   * @param $ // reference to jQuery.
    */
-    evalResponse(code, jQueryRef) {
-        const $ = jQueryRef; //eslint-disable-line
-        window.eval(code);
+    evalResponse(code, $) { //eslint-disable-line
```
mvorisek: would be good to add fixer for comments, to always normalize to `//<SP>text` and `/*<SP>text<SP>*/`

tests-behat/basicexecutor.feature
ibelar: allow to run a specific test instead of the whole suite. usage: bin/behat --tags basic-executor

I use to be able to run specific test before using bin/behat features/basicexecutor.feature but this is not working anymore and could not find a way using the new structure. So I just use tags instead

mvorisek: features are now moved to `tests-behat/`, if that works, I think it is better to no duplicate file names

js/src/components/multiline.component.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-bitwise */
```
mvorisek: we can disable rule only per specific line, see https://stackoverflow.com/questions/27732209/turning-off-eslint-rule-for-a-specific-line

tests-behat/basicexecutor.feature
ibelar: try using tests-behat but did not work. Also using tags is quicker...

js/src/services/api.service.js
ibelar: I added spaced-comment rule. It will at least catch them next time.

ibelar:
@mvorisek - Is there anything else? If not, I can approve and merge.

mvorisek:
> 
> 
> @mvorisek - Is there anything else? If not, I can approve and merge.

- import/no-unresolved
- import/prefer-default-export

are these two really needed? They look not ideal to me...
<br>
https://github.com/atk4/ui/pull/1426/files#diff-36b7a9a0205df583edf23575837c8b39R35-R36

120 chars would be much better, as it is the absolute maximum that can fix on 1/2 of a screen when doing diffs
<br>
https://github.com/atk4/ui/pull/1426/files#diff-36b7a9a0205df583edf23575837c8b39R19-R20

I would prefer to import one canonical name

otherwise LGTM, we can always adjust on the fly
<br>
Also, I would prefer to drop the Behat tags if possible, I think we should prevent to duplicate the names, because:
- names can be autocompleted by shell
- if we need more files per one functionlity group, we should group them in a folder

(I will look into this)

ibelar:
> > @mvorisek - Is there anything else? If not, I can approve and merge.
> 
> * import/no-unresolved
> * import/prefer-default-export
> 
> are these two really needed? They look not ideal to me...

Not need for us, so I turn them off - they were activate by airbnb-base.

> 
> https://github.com/atk4/ui/pull/1426/files#diff-36b7a9a0205df583edf23575837c8b39R35-R36
> 
> 120 chars would be much better, as it is the absolute maximum that can fix on 1/2 of a screen when doing diffs

- fixed

> https://github.com/atk4/ui/pull/1426/files#diff-36b7a9a0205df583edf23575837c8b39R19-R20
> 
> I would prefer to import one canonical name

- not sure what you mean here. It is only to indicated wich global var we are using so it does not throw errors...

> 
> otherwise LGTM, we can always adjust on the fly
> 
> 
> Also, I would prefer to drop the Behat tags if possible, I think we should prevent to duplicate the names, because:
> 
> * names can be autocompleted by shell
> * if we need more files per one functionlity group, we should group them in a folder
> 
> (I will look into this)

I have updated the name using xx-tag until we find a better solution. I could not run single test using behat tests-behat/crud.feature as before. I do not care which method we use, as long as we can run a single test 😄 

I will approve and merge!

ibelar (approved):
LGTM

mvorisek:
> > 
> > * import/no-unresolved
> > * import/prefer-default-export
> > 
> > are these two really needed? They look not ideal to me...
> 
> Not need for us, so I turn them off - they were activate by airbnb-base.

yes, keep them turned on if possible

> 
> > https://github.com/atk4/ui/pull/1426/files#diff-36b7a9a0205df583edf23575837c8b39R19-R20
> > I would prefer to import one canonical name
> 
> not sure what you mean here. It is only to indicated wich global var we are using so it does not throw errors...
> 

to use and import only one, probably only `$`

> 
> I have updated the name using xx-tag until we find a better solution. I could not run single test using behat tests-behat/crud.feature as before. I do not care which method we use, as long as we can run a single test 😄

yes, 🚀 🚀 🚀

mvorisek:
@ibelar filtering by path is working when path is absolute or relative to project dir when cwd is that dir too, see https://github.com/atk4/ui/pull/1427

also, can we we solve that two issues above (turn `import` rules back on and `jQuery` vs `$`)?

PR #1427, opened 2020-08-26
Revert Behat filename tags

mvorisek:
How to run Behat on Windows:

- copy `behat.yml.dist` to `behat.yml` locally and modify it to match your Selenium enviroment

- run Behat tests from CLI from root directory like:
```
php ../../behat/behat/bin/behat
```

- you may also append absolute or relative path to CWD to run only one specific test
```
php ../../behat/behat/bin/behat tests-behat/callback.feature
```

- on Windows, you can run the tests from different directory from batch like:
```
cmd /c "cd "%~dp0../vendor/atk4/ui" && php ../../behat/behat/bin/behat"
```

/cc @ibelar @georgehristov


ibelar:
@mvorisek - still can not run tests separately, even though I use my own behat configuration file.
This work:   `php ../../behat/behat/bin/behat` but run all tests.
This does not work: `php ../../behat/behat/bin/behat tests-behat/callback.feature`

Maybe because my environment is different I do not know.

Beside using tag works like a charm: `php ../../behat/behat/bin/behat --tags callback-tag`

So, What is wrong with using tag? This is a behat feature, so why can't we use it?

mvorisek:
I had simillar issue in past - do you have your configuration in ui root like `behat.yml`?

> So, What is wrong with using tag? This is a behat feature, so why can't we use it?

because now, they duplicate file names, that is the same like adding tags to every phpunit file. Possible, but not wanted :)

ibelar:
> I had simillar issue in past - do you have your configuration in ui root like `behat.yml`?

Yes - I do, like I said, running the complete suite is not a problem


> because now, they duplicate file names, that is the same like adding tags to every phpunit file. Possible, but not wanted :)

If only the name is a problem, we can name them anything we want. Besides if we change structure later, running a test using tag name would always be the same.

mvorisek:
working also in CI: https://github.com/atk4/ui/commit/9e26ba9ce78e30a02385e4645ed585fd51c78644

copy `behat.yml.dist` to `behat.yml` locally and modify it to match your Selenium enviroment, then using the latest Behat one test is run when path is given, both relative and absolute work

mvorisek:
@ibelar I noticed now when Behat tests fail, Behat output something like:
```
--- Failed scenarios:
    tests-behat\callback.feature:5
```

this is very strong argument to run single test preferably via filename as it can be easily copied

If you recreate your config from `behat.yml.dist` into ui root and run with the latest Behat from ui root directory, simply appending the feature path relative to the ui root should work for you too.

ibelar:
The more I read about tags, the more I found them useful. 
https://www.axelerant.com/resources/team-blog/how-to-speed-things-up-with-behat-commands

Benefits are there, plus you can still use the old syntax if you prefer.

Let's close this PR and move on, please!

mvorisek:
no, this is totally unneeded hack

ibelar:
> no, this is totally unneeded hack

Hack? Seriously? - Sorry @mvorisek but the tag will stay.

mvorisek:
yes, hack, copypasting filename to tag, you also described the reason for it to be that :)

ibelar:
We change the tag name, so all is good then. 😉

mvorisek:
@ibelar nothing personal, I simple did not get the point of using tags like filenames. I belive, it polute clean design. I also shown, that you should prefer using filenames because when test fails, filename is shown and the easiest way to start only that test is to copy that name.

My objective comment on this is how others use tags, for ex. Prestashop:
https://github.com/PrestaShop/PrestaShop/blob/develop/tests/Integration/Behaviour/Features/Scenario/Cart/CartToOrder/cart_to_order.feature
they do not use tags like you present it.

ibelar:
@mvorisek - The point of using tag is too get quicker access to run the test. Plus, you can group them, add them to scenarios for running the scenario only, etc. When creating/testing the test, it really helps!!!

Personally - I do not see any problem using the same name as file name but you do. That is why I put the xxx-tag mention on it. Should we find a better naming structure? Sure, why not, I am not against it (it does not have to be complicated too). But I definitely want to keep the tag name.

mvorisek:
You are repeating solved problem. Running single test is possible, grouping should be done by a folder level.

Appending `-tag` does not solve of course anything, do not take my words 1:1, understand why I am saying them.

My question is, why do others not use them you like want?

ibelar:
Who cares about how others are using them? Tag is part of Behat so Why not use them? 
It makes creating/debugging test quicker because you can target a specific one. There are probably a thousand ways of doing so but using tag is one of them, so WHY not?

Does the tag impact test performance? no;
Does the tag interfere with the build? no;
Does the tag help in creating/debugging features/scenarios? yes

Please give me one GOOD reason not to have them? ( using the same name of the file is  not ) Then I can reconsider. Otherwise let's move on. We already have lost way too much time with this.

mvorisek:
My feedback was posted, I belive, you should fix your local issue :)

ibelar:
There is no reason not to have tags. Closing for now.

mvorisek:
fixed in merged https://github.com/atk4/ui/pull/1430

PR #1429, opened 2020-08-27
Fix code for updated Model class (prevent reload with a different record

mvorisek:
related with https://github.com/atk4/data/pull/697

### DIFF
--- a/demos/collection/jsactionsgrid.php
+++ b/demos/collection/jsactionsgrid.php
@@ -10,7 +10,6 @@ require_once __DIR__ . '/../init-app.php';
 // Demo for Model action in Grid
 
 $country = new CountryLock($app->db);
-$country->tryLoadAny();
 DemoActionsUtil::setupDemoActions($country);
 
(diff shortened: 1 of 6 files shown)


demos/collection/jsactionsgrid.php
```diff
@@ -10,7 +10,6 @@
 // Demo for Model action in Grid
 
 $country = new CountryLock($app->db);
-$country->tryLoadAny();
```
mvorisek, line 13: @ibelar is this (and the same in `demos/layout/layout-panel.php`) load needed?

ibelar: Sure is. The model record needs to be loaded in order to use certains model action.

ibelar: > Sure is. The model record needs to be loaded in order to use certains model action.

My bad. I have read too fast the first time and thought you were refering another page. 😄 Yes, it can be remove in both files.

mvorisek: thanks

georgehristov (approved):
👍

PR #1430, opened 2020-08-27
[fix] - Js Package bundle

### LINKED ISSUE #1393: Disable debouce for testing, or remove it completely
related with https://github.com/atk4/ui/pull/1392

see https://github.com/atk4/ui/search?q=debounce

### LINKED ISSUE #1420: Form\Control\ScopeBuilder Rule on Date Field
When adding a rule on date field the current date is set as rule value by default in the UI component but not in the resulting JSON.

ibelar:
- Fix issue where QueryBuilder style set in .vue file was not loading
   - cause by css-loader v 4.2.1 (this fix revert package version to 3.6.0)
- remove unused clean package from dependencies;
- Set proper debounce value in js;

fixes #1393 
fixes #1420


js/src/services/panel.service.js
```diff
@@ -303,7 +303,7 @@ class PanelService {
         // clicking anywhere in main tag will close panel.
         $('main').on('click.atkPanel', debounce((evt) => {
             that.closePanel(id);
-        }, 100));
+        }, 250));
```
PhilippGrashoff, line 305: Is it on purpose that 2 times 350 is used and 250 here?

ibelar: yes - 250 here is when user clicking outside panel and is not related to user typing something. Shorter value is ok.

mvorisek: were there any server performance issues?

debounce feature slows tests - can we declare about the same debounce function and allow to override the debounce time for testing to like 5 ms

I already fixed all explicit waitings for AJAX requests, but this debounce is the last feature that need manual waiting now.

tests-behat/vue.feature
```diff
@@ -9,6 +9,5 @@ Feature: Vue
     Then I hide js modal
 
     When I fill in "atk-vue-search" with "united kingdom"
-# needed because debouce is used
-    Then I sleep 250 ms
+    Then I sleep 350 ms
```
mvorisek: there should be no need for waiting in testing, see https://github.com/atk4/ui/pull/1430/files#r478697880

js/webpack.config.js
```diff
@@ -80,6 +80,8 @@ module.exports = (env) => {
                 },
                 // this will apply to both plain `.css` files
                 // AND `<style>` blocks in `.vue` files
+                // NOTE: css-loader version 4.2.1 fail to load query builder style.
+                //      Revert to v. 3.6.0 until issue is found.
```
mvorisek: should be easy to find the issues in "query builder style" - delete all styling and add it back in chunks until issue is present

mvorisek (changes requested):

js/src/services/panel.service.js
ibelar: Debounce is there for a reason. Please never downgrade user experience in order to satisfy test performance...

tests-behat/vue.feature
ibelar: I add an option for settings debounce time in component and set proper test for it.

js/webpack.config.js
ibelar: Would be good if it was that simple. 😄 The problem is that component style is not loaded at all within the bundle, thus not available at all. Downgrade to version 3 fixes the issue.

js/src/components/item-search.component.js
```diff
@@ -41,13 +48,19 @@ export default {
...
     },
     methods: {
-        onChange: debounce(function (e) {
-            if (this.query !== this.temp) {
-                if (this.query === '') this.query = null;
-                this.sendQuery();
-                this.temp = this.query;
-            }
-        }, 350),
+        onChange: function () {
```
mvorisek, line 50: no, this is not good, we should not have an option for it in search compoment, instead we should have one global function like debounce, with with an global possibility to override

mvorisek: better to revert, we can not define so much code everywhere

js/webpack.config.js
mvorisek: we should aim to support the latest technologies, otherwise old technologies may be incompatible with another/user packages

so better to find a functional example for 4.x and mimic it

js/src/components/item-search.component.js
ibelar: This is what it does...

js/webpack.config.js
ibelar: agree but not now!

js/src/services/panel.service.js
mvorisek: > 
> 
> Debounce is there for a reason. Please never downgrade user experience in order to satisfy test performance...

I never do, I also want fast override for testing and also fast debounce for people that value every subsecond :)

js/webpack.config.js
mvorisek: ok with opened issue to be fixed soon

js/src/components/item-search.component.js
ibelar: Done! You can now built a package using `npm run build-behat `which will return a bebounce value of 0 for all debounce function.

mvorisek: @ibelar do not be ignorant, listen why, revert this and please produce less code in favor of maintaineable code, your productivity is very good, but you often come up with hacky solutions which you then defend like a champion, but why?

ibelar: @mvorisek - did you ever work in Vue? Reason why It need to be like this is that it allow customization of debounce value. But in order to do so, you can just pass the timeout value inside the vue method because it is out of context. 
In order to be in context and be able to use this.timeout props we need to be inside the method, thus the code you see now. But hey you can try for yourself if you want...

ibelar: I have changed to another way of setting proper `this` context. Does the same as before but cleaner.

js/webpack.config.js
ibelar: I always look up for using the latest package release in the bundle. No need to open an issue for this.

tests-behat/useraction.feature
```diff
@@ -1,14 +1,14 @@
-@user-action-tag
+@exec-user-action
```
mvorisek: we really need a CI test for tags if they are 1:1, otherwise we might confuse users why one name used one way work and not in another

js/src/agile-toolkit-ui.js
```diff
@@ -18,16 +18,18 @@ atk.version = () => _ATKVERSION_;
...
-    // Value for debounce time out (in ms) that will be apply globally when set.
-    let debounceTimeout = null;
+    const options = {
+        // Value for debounce time out (in ms) that will be apply globally when set.
+        debounceTimeout: null,
+    };
     return {
-        setDebounceTimeout: (value) => { debounceTimeout = parseInt(value, 10); },
-        getDebounceTimeout: () => debounceTimeout,
+        setOption: (name, value) => { options[name] = value; },
```
mvorisek: then simply get/set :)

ibelar: ok!

ibelar:
I will re have a look at css-loader package v4 and try to make it work.

mvorisek (approved):

PR #1433, opened 2020-08-28
[fix] Dropdown using multiple in read only mode

### LINKED ISSUE #1382: Dropdown readonly not working (with isMultiple)
If you add a Dropdown control with parameter 'readonly'=> true and isMultiple => true, the control cannot be "edited" in terms of selecting additional entries, BUT: you can still click on the 'X' for each entry and remove it.

Solution. Disable the "'X' delete item" functionality if readonly is set to true.

ibelar:
fix: #1382 
- prevent option to be remove in readonly mode when using multiple.

Screen Shot 2020-08-28 at 3 09 39 PM

### DIFF
--- a/src/Form/Control/Dropdown.php
+++ b/src/Form/Control/Dropdown.php
@@ -284,6 +284,9 @@ class Dropdown extends Input
             $this->setDropdownOption('showOnFocus', false);
             $this->setDropdownOption('allowTab', false);
             $this->removeClass('search');
+            if ($this->isMultiple) {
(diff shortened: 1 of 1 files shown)


mvorisek:
@ibelar are you sure, FUI has no single property readonly option? if no, merge

mkrecek234:
@mvorisek I could not find one either - implemented it similarly in the multiupload control

mvorisek (approved):

PR #1435, opened 2020-09-02
Refactor Js Atk Bundle eventBus

ibelar:
- Create eventBus under atk namespace;
 - remove deprecated Vue eventBus (Vue v3)
 - remove deprecated inline-template (Vue v3)
- Move options to atk-utils.js

Small BC- Break:

Emitting an event was done using a Vue Instance from the vue service. Support for emitting/responding to event from a Vue instance will be removed in V3.

Emitting/responding to event is now perform at the atk bundle level using mitt package.

Breaking part:

- JsVueService::emitEvent() method has been remove completelly;
- View::jsVueEmit() method has been replace by View::jsEmitEvent()

### DIFF
--- a/demos/javascript/vue-component.php
+++ b/demos/javascript/vue-component.php
@@ -86,14 +86,11 @@ $clock_script = "
           data: function() {
             return {style : this.clock, currentIdx : 0}
           },
-          created: function() {
+          mounted: function() {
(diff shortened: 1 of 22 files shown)


DarkSide666 (approved):
LGTM

PR #1438, opened 2020-09-08
Fix doc syntax

mvorisek:


docs/form-control.rst
```diff
@@ -94,7 +94,7 @@ properly:
 Hints
 -----
 
-.. php:attr: hint
+.. php:attr:: hint
```
mvorisek, line 97: example how rendered: https://agile-ui.readthedocs.io/en/latest/form-control.html#hints

(not actually rendered)

georgehristov (approved):

PR #1439, opened 2020-09-08
[fix] class declarations in docs

georgehristov:


mvorisek (approved):
I have no idea about the format, I trust you have a good itentions with it 8-)

PR #1443, opened 2020-09-08
Fix/doc attributes and typos

georgehristov:


mvorisek (approved):
very good, how you do this long doc/text updates?

georgehristov:
unfortunately manually :)
I have auto-correction marking wrong words but opening file by file

PR #1448, opened 2020-09-15
[feature] Vue Component loaded dynamically

ibelar:
Vue components in Js bundle count for about 75% of the js bundle size but are used only within certain UI views only.  This means that most of the time, we do not need such a big js bundle file.

Vue components are loaded dynamically thus only loading js file needed by the page.

On page load:

Screen Shot 2020-09-16 at 11 36 19 AM

After modal is open via reload:

Screen Shot 2020-09-16 at 11 36 36 AM

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -115,6 +115,9 @@ class App
     /** @var Persistence|Persistence\Sql */
     public $db;
 
+    /** @var bool Use Vue js bundle */
+    public $useVueJs = false;
+
     /** @var string[] Extra HTTP headers to send on exit. */
     protected $response_headers = [
         'cache-control' => 'no-store', // disable caching by default
@@ -447,8 +450,6 @@ class App
 
         $this->layout = $this->html->add($layout);
 
-        $this->initIncludes();
-
         return $this;
     }
 
@@ -468,7 +469,7 @@ class App
         $this->requireJs($this->cdn['serialize-object'] . '/jquery.serialize-object.min.js');
 
         // Agile UI
-        $this->requireJs($this->cdn['atk'] . '/atkjs-ui.min.js');
+        $this->useVueJs ? $this->requireJs($this->cdn['atk'] . '/atkjs-vue.min.js') : $this->requireJs($this->cdn['atk'] . '/atkjs-ui.min.js');
         $this->requireCss($this->cdn['atk'] . '/agileui.css');
     }
 
@@ -507,6 +508,8 @@ class App
      */
     public function run()
     {
+        $this->initIncludes();
+
         $isExitException = false;
         try {
             $this->run_called = true;
--- a/src/App.php
+++ b/src/App.php
@@ -115,9 +115,6 @@ class App
     /** @var Persistence|Persistence\Sql */
     public $db;
 
-    /** @var bool Use Vue js bundle */
-    public $useVueJs = false;
-
     /** @var string[] Extra HTTP headers to send on exit. */
     protected $response_headers = [
         'cache-control' => 'no-store', // disable caching by default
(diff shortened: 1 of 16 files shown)


src/Form/Control/TreeItemSelector.php
```diff
@@ -80,6 +80,9 @@ protected function init(): void
     {
         parent::init();
 
+        // TreeItemSelector use Vue components.
```
mvorisek: comments like this should be dropped, yes, this component use vue :)

src/App.php
```diff
@@ -468,7 +469,7 @@ public function initIncludes()
         $this->requireJs($this->cdn['serialize-object'] . '/jquery.serialize-object.min.js');
 
         // Agile UI
-        $this->requireJs($this->cdn['atk'] . '/atkjs-ui.min.js');
+        $this->requireJs($this->cdn['atk'] . ($this->useVueJs ? '/atkjs-vue.min.js' : '/atkjs-ui.min.js'));
```
mvorisek: not ideal, we should always load atk and optionally the 2nd js for vue, otherwise use might need to download more data than now (think like login without vue and app with vue)

if not possible, then maybe better even not to split... :(

mvorisek:
I think making this PR is quite challenging as reload might reload a new component that is the first one which require Vue, no, we can not assume, that user will load/flag to load vue manually

src/App.php
ibelar: Yes, can split, just thought it would be better to only have one request instead of two if Vue is needed.

ibelar:
> I think making this PR is quite challenging as reload might reload a new component that is the first one which require Vue, no, we can not assume, that user will load/flag to load vue manually

If split in two parts than I will check if we can load vue js dynamically. Otherwise, you are right, user will have to preload Vue in order for the reload to work. Then we need to decide if this is a good trade off for loading performance.

src/App.php
mvorisek: requests are not an issue, like css loads several fonts etc, 2, 3, requests for JS instead 1 is absolutelly fine

src/Form/Control/TreeItemSelector.php
ibelar: Property not use now

src/App.php
ibelar: Will load js per component no need for two separate bundles.

mvorisek: If I will need 20 components (I know, there is currently not that many), what will be the total size and what is the avg. size per component?

mvorisek: there will be at least one mroe RTT with this approach if we render new/unseen component often

ibelar: > If I will need 20 components (I know, there is currently not that many), what will be the total size and what is the avg. size per component?

This really depend on the component itself. For example, tree-item-selector size is only 10kb while query-builder does 400k. I also intend to use more of external Vue component, like v-calendar, which is bigger in size. it only make sense to load it when needed. Otherwise, js bundle keep increase in size while we are using only 10% of it most of the time.

> there will be at least one mroe RTT with this approach if we render new/unseen component often

Not sure what you mean here...

mvorisek: RTT - round time trip - basically one latency, which can be 100 - 500 ms if you are really far away from the server, but it should be only once per browser session if cached, so ok probably

10kb - good, but what will happen if there will be multiple simillar components? If there will be some dep with size like 500kb, I think, it will be packed everywhere... So I think sooner or later we will need more lib than component oriented approach.

ibelar: > 10kb - good, but what will happen if there will be multiple simillar components? If there will be some dep with size like 500kb, I think, it will be packed everywhere... So I think sooner or later we will need more lib than component oriented approach.

I think this is the right approach for now. When we create a Vue 'app' instance, we need to supply our root component and this is what this does. Then from the root, you can define/add other components that your need for your app instance.

mvorisek: I have nothing to merge this, seems well made (not too much added code :) )

DarkSide666:
Loading performance if one of the most important things for any webpage ...

DarkSide666 (approved):
LGTM

PR #1450, opened 2020-09-15
Move setting ScopeBuilder default scope

georgehristov:


DarkSide666 (approved):
This way it also work.
Only drawback is that if setModel() is not called, then query is not built. Probably not big issue, but will see...

georgehristov:
Actually setting model is required as that is how ScopeBuilder rules are built

mvorisek:
A wonder if this helps @DarkSide666 as field value must be instance of Scope this way... But if yes, then it is perfect.

PR #1456, opened 2020-09-20
Add smoke and burn CI testing

mvorisek:
merge after https://github.com/atk4/core/pull/275


DarkSide666 (approved):
nice

mvorisek:
something is definitely leaking, see https://github.com/atk4/ui/runs/1145743024#step:9:35 , the memory is increasing during testing...

mvorisek:
replicateable on Windows too, at least php require of file containing anonymous class is leaking

https://bugs.php.net/bug.php?id=80140

fixing now with low burn iteration count

Issue #1399, opened 2020-08-06
onChange for Form\Lookup and Form\Control inconsistent to \atk4\ui\DropDown


mkrecek234:
This does not work in Form\Control\Lookup or Form\Control\Dropdown (but it does in \atk4\ui\Dropdown:

```
     *      $dropdown->onChange(function($item) {
     *          return 'New seleced item: '.$item;
     *      });.
```

How to do the same in Form\Control\Lookup? I need to use the selected value to search a model and then pre-fill further fields in the form. Thx!

PR #1457, opened 2020-09-22
Dropdown refactor

ibelar:
- Now use model id instead of model name for onChange event;
```
$dropdown = \atk4\ui\Dropdown::addTo($menu, ['With Callback', 'dropdownOptions' => ['on' => 'hover']]);
$dropdown->setSource(['a', 'b', 'c']);
$dropdown->onChange(function ($itemId) {
    return 'New seleced item id: ' . $itemId;
});
```   
- Allow to set argument name to send with the request;

Small BC break
 - $js property was rename to $dropdownOptions

### DIFF
--- a/src/Dropdown.php
+++ b/src/Dropdown.php
@@ -18,18 +18,18 @@ class Dropdown extends Lister
     public $cb;
 
     /**
-     * Supply an optional parameter to the drop-down.
+     * As per Fomantic-ui dropdown options.
      *
-     * @var array will be converted to json passed into dropdown()
+     * @var array
      */
-    public $js;
+    public $dropdownOptions = [];
 
     protected function init(): void
     {
         parent::init();
 
         if (!$this->cb) {
-            $this->cb = JsCallback::addTo($this, ['urlTrigger' => 'item']);
+            $this->cb = JsCallback::addTo($this);
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/Dropdown.php
```diff
@@ -18,54 +18,50 @@ class Dropdown extends Lister
...
      *      $dropdown->setModel($menuItems);
      *      $dropdown->onChange(function($item) {
-     *          return 'New seleced item: '.$item;
+     *          return 'New selected item: '.$item;
      *      });.
      *
      * @param \Closure $fx handler where new selected Item value is passed too
      */
-    public function onChange(\Closure $fx)
+    public function onChange(\Closure $fx, string $argumentName = 'item')
```
mvorisek: when a different argument name makes sense?

ibelar: Let's say you have two dropdowns in the same page. They will both use 'item' in the post request. This will help differentiate them.

mvorisek: you mean post request of dynamic dropdown? (which is always for one dropdown only)

ibelar: I was thinking more in term of handling multiple dropdown using one function, something like:
```
$menu_selected = function($id) {
  $menu = isset($_POST['menu_1']) ? 'menu 1': (isset($_POST['menu_2']) ? 'menu 2' : 'unknown');

  return 'you have selected item ' . $id . ' of ' . $menu;
};

$menu = \atk4\ui\Menu::addTo($app);
$dropdown = \atk4\ui\Dropdown::addTo($menu, ['Menu 1', 'dropdownOptions' => ['on' => 'hover']]);
$dropdown->setSource($items);
$dropdown->onChange($menu_selected, 'menu_1');

$dropdown = \atk4\ui\Dropdown::addTo($menu, ['Menu 2', 'dropdownOptions' => ['on' => 'hover']]);
$dropdown->setSource($items);
$dropdown->onChange($menu_selected, 'menu_2');
```

By default onChange handler is using POST request, may be we could change this for GET instead?

mvorisek: thanks for the code/example

we should definitely handle this differently, so drop it

what you want is to simply pass a different callback like:
```
$dropdown->onChange(function ($id) use ($menu_selected) { $menu_selected($id, 'menu_1'); });
```

correct `onChange` should be called based on Callback id/name...

ibelar: make sense!

mvorisek (approved):

mkrecek234:
This closes issue #1399, thank you @ibelar

PR #1458, opened 2020-09-23
[fix] Behat test when using chunk js file

ibelar:
- Make sure Vue dynamic js file are fully loaded prior to test them.
- Will show loading state while bigger component are loading or error message if fail.

Note: also rename bundlePublicPath var to a more specific name: __atkBundlePublicPath in order to avoid possible conflict.

### DIFF
--- a/js/src/agile-toolkit-ui.js
+++ b/js/src/agile-toolkit-ui.js
@@ -1,6 +1,6 @@
 /* eslint-disable */
 /* global _ATKVERSION_:true, __webpack_public_path__:true */
-__webpack_public_path__ = window.bundlePublicPath === undefined ? '/public/' :  window.bundlePublicPath + '/';
+__webpack_public_path__ = window.__atkBundlePublicPath === undefined ? '/public/' :  window.__atkBundlePublicPath + '/';
 
 import debounce from 'debounce';
 import 'core-js/stable';
--- a/js/src/services/vue.service.js
+++ b/js/src/services/vue.service.js
@@ -1,17 +1,36 @@
 import Vue from 'vue';
 import SuiVue from 'semantic-ui-vue';
-
 import atkClickOutside from '../directives/click-outside.directive';
 import { focus } from '../directives/commons.directive';
 
 Vue.use(SuiVue);
 
+// Vue loader component to display while dynamic component is loading.
+const atkVueLoader = {
+    name: 'atk-vue-loader',
+    template: '<div><div class="ui active centered inline loader"></div></div>',
+};
+
+// Vue error component to display when dynamic component loading fail.
+const atkVueError = {
+    name: 'atk-vue-error',
+    template: '<div class="ui negative message"><p>Error: Unable to load Vue component</p></div>',
+};
+
+// Return async component that will load on demand.
+const componentFactory = (component) => () => ({
+    component: component(),
+    loading: atkVueLoader,
+    error: atkVueError,
+    delay: 200,
+});
+
 const atkComponents = {
-    'atk-inline-edit': () => import(/* webpackChunkName: "atk-vue-inline-edit" */'../components/inline-edit.component'),
-    'atk-item-search': () => import(/* webpackChunkName: "atk-vue-item-search" */'../components/item-search.component'),
-    'atk-multiline': () => import(/* webpackChunkName: "atk-vue-multiline" */'../components/multiline/multiline.component'),
-    'atk-tree-item-selector': () => import(/* webpackChunkName: "atk-vue-tree-item-selector" */'../components/tree-item-selector/tree-item-selector.component'),
-    'atk-query-builder': () => import(/* webpackChunkName: "atk-vue-query-builder" */'../components/query-builder/query-builder.component.vue'),
+    'atk-inline-edit': componentFactory(() => import(/* webpackChunkName: "atk-vue-inline-edit" */'../components/inline-edit.component')),
+    'atk-item-search': componentFactory(() => import(/* webpackChunkName: "atk-vue-item-search" */'../components/item-search.component')),
+    'atk-multiline': componentFactory(() => import(/* webpackChunkName: "atk-vue-multiline" */'../components/multiline/multiline.component')),
+    'atk-tree-item-selector': componentFactory(() => import(/* webpackChunkName: "atk-vue-tree-item-selector" */'../components/tree-item-selector/tree-item-selector.component')),
+    'atk-query-builder': componentFactory(() => import(/* webpackChunkName: "atk-vue-query-builder" */'../components/query-builder/query-builder.component.vue')),
 };
 
 // setup atk custom directives.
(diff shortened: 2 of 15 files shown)


tests-behat/vue.feature
```diff
@@ -1,8 +1,11 @@
 Feature: Vue
   Testing Vue component
 
-  Scenario: testing InlineEdit
+  Scenario: loading components
     Given I am on "javascript/vue-component.php"
+    Then I check for vue components to be load
```
mvorisek: this is basically a wait we want to avoid in any circumstances as it will be needed before any component without standard JS

we may want to tweak the Context::jqueryWait and/or Context::getFinishedScript methods instead

can we somehow monitor loading JS?

I think we can:
1. set some global/app flag with that JS name
2. unset this flag right at the end? in that JS file
3. in Context::getFinishedScript check if no JS loading flag is set

ibelar: Not sure what you have in mind here. To me, this is what it does.

Vue component js file is now loaded dynamically when required.

This is the cycle of operation:
- createAtkVue method is creating the Vue instance;
- When It does, it set the flag for this component 'isLoaded' to false;
- Then js file need for this component start loading and when the file is fully loaded, 'isLoaded' for that component is set to true;

calling `atk.vueService.areComponentsLoaded()` will return true only when all components file requests are loaded, otherwise false if still trying to load some files.

What this test does is only checking if all file are loaded using:
 `$loaded = $this->getSession()->wait(2000, 'atk.vueService.areComponentsLoaded()');`

It will wait until the javascript expression above return true, or until the max time of 2sec is reached. 
In most of the case, there is no waiting time, cause test pass most of the time. If not, it will probably wait just a few ms or until all files are loaded. Which is needed to run the test.

mvorisek: I see `atk.vueService.areComponentsLoaded()` - can we simply add it into `Context::getFinishedScript`?

this seems only vue related, the JS split is for vue only?

js/src/services/vue.service.js
```diff
@@ -1,17 +1,36 @@
...
-
 import atkClickOutside from '../directives/click-outside.directive';
 import { focus } from '../directives/commons.directive';
 
 Vue.use(SuiVue);
 
+// Vue loader component to display while dynamic component is loading.
+const atkVueLoader = {
+    name: 'atk-vue-loader',
+    template: '<div><div class="ui active centered inline loader"></div></div>',
```
mvorisek, line 13: why this template, should it display/flash some loader icon shortly?

tests-behat/vue.feature
ibelar: yes, for vue only

js/src/services/vue.service.js
ibelar: Yes, in case a component is too large and take some time to load.
We can see it sometime using query builder

tests-behat/vue.feature
mvorisek: good, I pushed the change to my biggest concern - I do not want to see any wait in tests - all loading conditions must be checked on the fly, not manually

js/src/agile-toolkit-ui.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable */
 /* global _ATKVERSION_:true, __webpack_public_path__:true */
-__webpack_public_path__ = window.bundlePublicPath === undefined ? '/public/' :  window.bundlePublicPath + '/';
+__webpack_public_path__ = window.__atkBundlePublicPath === undefined ? '/public/' :  window.__atkBundlePublicPath + '/';
```
mvorisek, line 3: is there any reason why this must be under `windows` instead under `atk`?

js/src/services/vue.service.js
```diff
@@ -112,6 +150,24 @@ class VueService {
...
+    componentIsLoaded(name) {
+        this.vues.forEach((component) => {
+            if (component.name === name) {
+                component.isLoaded = true;
+            }
+        });
+    }
+
+    /**
+     * Check if all components on page are ready or fully loaded.
```
mvorisek:
```suggestion
     * Check if all components on page are ready and fully loaded.
```

js/src/services/vue.service.js
```diff
@@ -112,6 +150,24 @@ class VueService {
     getVue() {
         return Vue;
     }
+
+    /*
+    * Mark a component as loaded.
+    */
+    componentIsLoaded(name) {
```
mvorisek: better name `markComponentLoaded` - this is not getter

js/src/services/vue.service.js
```diff
@@ -79,26 +103,40 @@ class VueService {
...
         });
     }
 
+    /*
+    *  Add component to vues container.
+    *  Group ids that are using the same component.
+     */
+    registerComponent(component) {
+        // check if that component is already registered
+        const registered = this.vues.filter((comp) => comp.name === component.name);
```
mvorisek, line 126: like we have below `areComponentsLoaded` we can add `isComponentLoaded(name)`

js/src/agile-toolkit-ui.js
ibelar: yes, because atk is not loaded yet and webpack will produce a proper import statement base on this value inside atk.

js/src/services/vue.service.js
ibelar: sure, I can add it. But I just do not see any good reason why you would check for a specific component only. Do you?

js/src/services/vue.service.js
ibelar: I will change

js/src/services/vue.service.js
mvorisek: Some JS code may need to check if some component is loaded or not, `areComponentsLoaded` can/should use this function plus this check where commetend.

js/src/agile-toolkit-ui.js
mvorisek: hmm, I find it messy, but if browsers or some analysers do not complain about it, ok to keep...

js/src/services/vue.service.js
ibelar: I really do not think this is necessary. Since we are talking about Vue components only, the best way to determine if a component is ready or not is via the component itself, from the 'mounted' hook method for example. areComponentsLoaded is really only use for Behat.

mvorisek: ok, if needed, we can change later... :)

js/src/agile-toolkit-ui.js
ibelar: I know it looks messy but this needs to be set, in the main bundle entry file and prior to any import statement in order to work properly. And it does work well actually. It is needed for Saasty actually for it to properly load atk bundle file.

ibelar: Is everything ok?

mvorisek: LGTM

ibelar: an approval would be nice then 😄

mvorisek (approved):
🚀

PR #1463, opened 2020-09-30
Refactor ScopeBuilder date picker

ibelar:
- Create atk-date-picker component wrapper for v-date-picker;

- Allow customizing date picker in the query builder via v-date-picker props.

Example: Starting calendar on a Monday by setting the first-day-of-week option and changing language

```
    public $datePickerProps = [
        'locale' => 'fr-Fr',
        'masks' => ['input' => 'YYYY-MM-DD'],
        'first-day-of-week' => 2
    ];
```

Screen Shot 2020-09-30 at 10 59 14 AM

### DIFF
--- a/js/src/atk-utils.js
+++ b/js/src/atk-utils.js
@@ -39,7 +39,7 @@ const atkEventBus = (function () {
 
 /*
 * Utilities function that you can execute
-* from atk context. Usage: atk.exec().date().parse('string');
+* from atk context. Usage: atk.utils.date().parse('string');
 */
 const atkUtilites = (function () {
     return {
--- a/js/src/atk-utils.js
+++ b/js/src/atk-utils.js
@@ -37,4 +37,24 @@ const atkEventBus = (function () {
     };
 }());
 
-export { atkOptions, atkEventBus };
+/*
+* Utilities function that you can execute
+* from atk context. Usage: atk.exec().date().parse('string');
+*/
+const atkUtilites = (function () {
+    return {
+        date: function() {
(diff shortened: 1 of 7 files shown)


js/src/atk-utils.js
```diff
@@ -37,4 +37,24 @@ const atkEventBus = (function () {
     };
 }());
 
-export { atkOptions, atkEventBus };
+/*
+* Utilities function that you can execute
+* from atk context. Usage: atk.utils.date().parse('string');
+*/
+const atkUtilites = (function () {
```
mvorisek: atkUtils to be consistent with `atk.utils.`

ibelar: done!

DarkSide666 (approved):
What if we set `'masks' => ['input' => 'DD-MM-YYYY']` or `phpDateFormat = 'd-m-Y'`?
Are these related to that regex and if yes then it will probably fail.

Besides that LGTM.

ibelar:
> What if we set `'masks' => ['input' => 'DD-MM-YYYY']` or `phpDateFormat = 'd-m-Y'`?
> Are these related to that regex and if yes then it will probably fail.

 `'masks' => ['input' => 'DD-MM-YYYY']` is for how v-date-picker will display date inside the input to user, while `phpDateFormat = 'd-m-Y'` is the format that will be set in query. They are are not regex related. This is another reason why I want to switch to flatpickr (see #1467 ) Flatpickr use PHP date format so no need to translate.

PR #1465, opened 2020-10-01
Fix JS bundle URL escape

mvorisek:
we must escape for JS/JSON, not using HTML

### DIFF
--- a/demos/javascript/vue-component.php
+++ b/demos/javascript/vue-component.php
@@ -61,7 +61,7 @@ $lister->setModel($search->setModelCondition($model))->setLimit(50);
 // ****** CREATING CUSTOM VUE USING EXTERNAL COMPONENT *****************************
 \atk4\ui\Header::addTo($app, ['External Component', 'subHeader' => 'Creating component using an external component definition.']);
 
-$app->requireJs('https://unpkg.com/vue-clock2@1.1.5/dist/vue-clock.min');
+$app->requireJs('https://unpkg.com/vue-clock2@1.1.5/dist/vue-clock.min.js');
 
 // Injecting template but normally you would create a template file.
 $clock_template = new \atk4\ui\Template('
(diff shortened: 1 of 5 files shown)


demos/javascript/vue-component.php
```diff
@@ -61,7 +61,7 @@
 // ****** CREATING CUSTOM VUE USING EXTERNAL COMPONENT *****************************
 \atk4\ui\Header::addTo($app, ['External Component', 'subHeader' => 'Creating component using an external component definition.']);
 
-$app->requireJs('https://unpkg.com/vue-clock2@1.1.5/dist/vue-clock.min');
+$app->requireJs('https://unpkg.com/vue-clock2@1.1.5/dist/vue-clock.min.js');
```
mvorisek, line 64: just prevent unneeded redirect...

src/App.php
```diff
@@ -465,7 +465,7 @@ public function initIncludes()
         $this->requireCss($this->cdn['atk'] . '/agileui.css');
 
         // Set js bundle dynamic loading path.
-        $this->html->template->trySet('PublicPath', $this->cdn['atk']);
+        $this->html->template->trySet('InitJsBundle', (new JsExpression('window.__atkBundlePublicPath=[]', [$this->cdn['atk']]))->jsRender(), false);
```
mvorisek: jsRender can not close script tag, thus 100% safe to not escape by definition

src/JsExpression.php
```diff
@@ -21,13 +21,7 @@ class JsExpression implements JsExpressionable
...
     public $args = [];
 
-    /**
-     * Constructor.
-     *
-     * @param string $template
-     * @param array  $args
-     */
-    public function __construct($template = '', $args = [])
+    public function __construct(string $template = '', array $args = [])
```
mvorisek, line 24: array tag type is super important - otherwise string can be passed without any warning!

ibelar (approved):
LGTM

PR #1466, opened 2020-10-01
Cleanup JsExpression to string/JS

mvorisek:
breaking as `jsRender()` method signature is changed and needs to be adjusted if overrided

also use App JSON encode/decode as possible, but keep native json_encode/json_decode in View, so it can be (at least for now) used without an App


georgehristov (approved):
LGTM

PR #1469, opened 2020-10-03
refactor - Calendar input

### LINKED ISSUE #1467: WIP - DatePicker consistency
### Problem:
- Calendar input uses Fomantic-UI calendar in order to display date selection;
- Vue component like QueryBuilder or Multiline uses v-date-picker for date selection because there are no vue equivalent to Fomantic-ui calendar module;

Although Vue v-date-picker is similar in style to Fomantic-Ui module, they are not the same (see image below).
 - it is possible to use date-time format in Fomantic-Ui module but not in the v-date-picker Vue compnent. 
 - v-date-picker is localizable while Fomantic-Ui is not (unless you provide your own)

### The propose solution:

Drop Fomantic-Ui calendar module and v-date-picker and replace them using Flatpickr : https://flatpickr.js.org/examples/
Flatpickr is highly customizable and has a Vue wrapper equivalent, which would bring consistency within the framework (regular input vs Vue component input) 
Flatpickr also supports more features then Fomantic-Ui calendar module or v-date-picker.

Fomantic-UI calendar module:

Screen Shot 2020-10-02 at 11 15 56 AM

v-date-picker:
Screen Shot 2020-10-02 at 11 16 12 AM

ibelar:
Use flatPickr js.

#### BC-Break
Small BC-Break if using Fomantic-UI specific options. You may have to set them using flatPickr
see: https://flatpickr.js.org/options/

Fix #1467 

Screen Shot 2020-10-03 at 9 38 42 AM

Screen Shot 2020-10-03 at 9 38 52 AM


PhilippGrashoff:
@ibelar Whats your reason for not using FUI calendar input any more? In the last year, it was extended a lot and has a lot of functionality now.

Update: I just read #1467, this seems sensible reasoning. However, not having 2 different JS libs for the same inputs (FUI and SUI-Vue which are both used for input type="text" depending on Form Field for example) would be better I think. But we discussed this already and I understood that this would mean loooots of work.

src/Form/Control/Calendar.php
```diff
@@ -4,78 +4,102 @@
...
+        $this->options['dateFormat'] = $this->options['dateFormat'] ?? $this->app->ui_persistence->{$this->type . '_format'};
 
-        if ($dayOfWeek = $this->app->ui_persistence->firstDayOfWeek) {
-            $this->options['firstDayOfWeek'] = $dayOfWeek;
+        // set default according to type.
+        switch ($this->type) {
+            case 'time':
+                $this->options['enableTime'] = true;
+                $this->options['noCalendar'] = true;
+                $this->options['time_24hr'] = $this->use24HrTime;
```
mvorisek: no property for 24h - parse php date format like:

```
private function is24hrTimeFormat(): bool
{
    return !preg_match('~[gh]~', $this->options['dateFormat'] ?? $this->app->ui_persistence->{$this->type . '_format'});
}
```

ibelar: this is not for date parsing but rather an option for flatpickr on how to display/set time for user.

mvorisek: if we set somewhere in php a format "option on how to display/set time for user", then parse if 24hr format from it, that is all

ibelar: I think they are independent (setting vs output). But I will check and change accordingly. I will also need to add tests for this.

src/Form/Control/ScopeBuilder.php
```diff
@@ -261,8 +262,51 @@ class ScopeBuilder extends Control
...
+                self::OPERATOR_TIME_LESS_EQUAL,
+                self::OPERATOR_EMPTY,
+                self::OPERATOR_NOT_EMPTY,
+            ],
+            'props' => [
+                'dateFormat' => 'H:i',
+                'enableTime' => true,
+                'noCalendar' => true,
+                'time_24hr' => true,
+            ],
```
ibelar: @georgehristov - Can you take a look at how datetime and time rules are defined. Time and Datetime can now by supported via vue-flatpickr .
Question: 
Do we need datetime at all?
Flatpickr vue component needs to have proper set of properties define to switch from date, time or datetime picker. I add them via $rule['props'] property. Is that ok? Do we need to switch format based on ui persistence? if yes, then where do we set the property. If not, should we just set them in javascript not in php?

src/Form/Control/Calendar.php
```diff
@@ -146,4 +145,27 @@ public function jsGetFlatPickr(): JsExpression
...
+    {
+        // translate from php to flatpickr.
+        $format = preg_replace(['~[aA]~', '~[s]~', '~[g]~'], ['K', 'S', 'G'], $format);
+
+        return $format;
+    }
+
+    private function use24hrTimeFormat(): bool
+    {
+        return !(bool) preg_match('~[gGh]~', $this->options['altFormat'] ?? $this->options['dateFormat']);
```
mvorisek: no need for bool cast

src/Persistence/Type/Basic.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace atk4\ui\Persistence\Type;
+
+/**
+ * Base type class.
+ */
+class Basic
```
mvorisek: This dupliates a lot of data Type, not sure, if this is step into the right direction.

ibelar: The idea was to:
- clearly separate each field type value casting to it's own class, i.e. clearer of what each type value casting does; 
- easily add a new field type without having to change Persistence/UI class;
- override each type class easily to adapt for custom needs;

mvorisek: add/override is the only plus, but it strong argument to do it so.

I added some design comments directly

/cc @georgehristov

src/Persistence/Type/Castable.php
```diff
@@ -0,0 +1,20 @@
...
+namespace atk4\ui\Persistence\Type;
+
+use atk4\data\Field;
+
+interface Castable
+{
+    /**
+     * Cast value when loaded from POST request.
+     */
+    public static function castLoadValue(Field $field, $value);
```
mvorisek: not static!

src/Persistence/Ui.php
```diff
@@ -19,35 +23,26 @@
...
+        'boolean' => Boolean::class,
+        'date' => Date::class,
+        'time' => Date::class,
+        'datetime' => Date::class,
+        'money' => Money::class,
+        'array' => Serial::class,
+        'object' => Serial::class,
+    ];
+
+    public function getTypeClass(string $type): ?string
```
mvorisek: `getType(string $type): Castable`

src/Persistence/Ui.php
```diff
@@ -19,35 +23,26 @@
...
+    {
+        return $this->typeClass[$type] ?? null;
+    }
 
-    public $calendar_options = [];
+    public function setTypeClass(string $type, string $class)
```
mvorisek: registerTypeClass or addTypeClass

see how ORM/DBAL type registry works (yes, they are for php <->, but usage should be about the same)

src/Persistence/Type/Castable.php
ibelar: done!

src/Persistence/Ui.php
ibelar: done!

src/Persistence/Ui.php
ibelar: done! I used registerTypeClass

src/Form/Control/Calendar.php
ibelar: done!

demos/form-control/calendar.php
```diff
@@ -35,32 +35,32 @@
     Form\Control\Calendar::class,
     'type' => 'date',
     'caption' => 'Allow input (F d, Y)',
-    'options' => ['allowInput' => true]])
-     ->set(date(Date::getFormat('date')));
+    'options' => ['allowInput' => true], ])
```
mvorisek: end `]` on new line, if there is NL after `[`

src/Form/Control/Calendar.php
ibelar: option is now set according to php format and test is added.

demos/form-control/calendar.php
ibelar: fix

demos/form-control/calendar.php
```diff
@@ -0,0 +1,80 @@
...
+
+/** @var \atk4\ui\App $app */
+require_once __DIR__ . '/../init-app.php';
+
+$layout = GridLayout::addTo($app, ['rows' => 1, 'columns' => 2]);
+
+$form = Form::addTo($layout, [], ['r1c1']);
+
+Date::setFormat('date', 'Y-m-d');
+$form->addControl('date_y_m_d', [Form\Control\Calendar::class, 'type' => 'date', 'caption' => 'Date (Y-m-d)'])
```
mvorisek, line 21: UI format types must be implemented in non-static way, probably once per App which we can then easily access with like `->frontFormatter` and conversion must be definitely handled in Control itself - in php/end user/programmer code, we must NEVER work on formatter data - these are only for user

mvorisek: example here: for date type, type will allow to set `\DateTimeInterface` only, never formatted string from php

```
->set(new \DateTime());
```

ibelar: Revert to previous ui_persistence for now.

src/Form/Control/ScopeBuilder.php
```diff
@@ -333,17 +335,43 @@ protected function buildQuery(Model $model)
...
             'id' => $field->short_name,
             'label' => $field->getCaption(),
             'options' => $this->options[strtolower((string) $type)] ?? [],
-        ], $field->ui['scopebuilder'] ?? []), $field);
+        ], $field->ui['scopebuilder'] ?? [], $typeOptions), $field);
 
         return $this;
     }
 
+    protected function getTypeOptions(string $type)
```
ibelar: @georgehristov - I refactored ScopeBuilder using this rule['options'] instead of settings props to date, datetime or time type. This way, if we decide to use other Vue components in QueryBuilder, then we could simply pass component props using rule.options. 

I have also set dateFormat accoding to ui_persistence, which means that value sent to the server is format accordingly. Let me know if you need to change, i.e. if value sent always need to be in the same format.

src/Form/Control/ScopeBuilder.php
ibelar: I remove props and now use rule['options'] see new comments.

src/Form/Control/Calendar.php
```diff
@@ -145,26 +145,26 @@ public function getJsInstance(): JsExpression
         return (new Jquery('#' . $this->id . '_input'))->get(0)->_flatpickr;
     }
 
-    private function translateFormat(string $format): string
+    public static function translateFormat(string $format): string
```
mvorisek: is there any reason for static except "it does not require `$this` and maybe is expected to not change"?

ibelar: SopeBuilder also uses flatpickr and I uses it to properly set options to it.

```
$options['dateComponent']['time_24hr'] = Calendar::use24hrTimeFormat($format);
$options['dateComponent']['noCalendar'] = ($type === 'time');
$options['dateComponent']['enableSeconds'] = Calendar::useSeconds($format);
```

mvorisek: instantiate dummy Calendar control class in ScopeBuilder or UI Persistence if needed

ibelar: and why not use static function? Options are not going to change. if they do, they will need to be change everywhere...

mvorisek: Static anything is not optimal from OOP design in general as DI is not easy then. I understand, that here you expect it will not be required to change, but what if user want to use different Calendar lib? I am very happy, that Atk does not use static, so I want to keep that clean code.

ibelar: make sense. I reverted it to public function instead.

src/Form/Control/ScopeBuilder.php
ibelar: @georgehristov - Just refactor (again) to use component type instead of field type. and change $rule['options'] to $rule['componentProps'] I think it make more sense this way.

ibelar: Now use altFormat for date display and dedicated format for setting string value.

tests-behat/bootstrap/Context.php
```diff
@@ -428,6 +429,86 @@ public function iClickIconUsingCss($arg1)
...
+     */
+    public function textRuleOperatorAndValue($arg1, $arg2, $arg3)
+    {
+        $rule = $this->assertRuleExist($arg1);
+        $this->assertRuleOperatorSelectedValue($rule, $arg2);
+        $this->assertRuleInputValue($rule, $arg3);
+    }
+
+    /**
+     * @Then /^hasRef rule "([^"]*)" operator is "([^"]*)" and value is "([^"]*)"$/
```
mvorisek: Then ScopeBuilder rule...

and name method like scopeBuilderHasRef...

and not need for comment `//// ScopeBuilder rule /////////` above

demos/_unit-test/scope-builder.php
```diff
@@ -15,20 +17,33 @@
 // Display: Project Name matches regular expression [a-zA-Z]
 $project = new Condition('project_name', Condition::OPERATOR_REGEXP, '[a-zA-Z]');
 // Display: Client Country Iso equals Brazil => ToWord: Client Country Iso is equal to 'Brazil'
-$brazil = new Condition('client_country_iso', '=', 'Brazil');
+$brazil = new Condition('client_country_iso', Condition::OPERATOR_EQUALS, 'Brazil');
```
mvorisek: well, I think `=` was invented for humans and well standardized, @georgehristov wdyt? Is there even any sense making this constant public?

ibelar: What is the status here? should we use condition const or symbol?

georgehristov: We should use symbols I would say

ibelar: thanks, I will change it back to symbol.

tests-behat/bootstrap/Context.php
```diff
@@ -500,7 +500,7 @@ public function iCheckIfWordMatch()
 
         $responseWord = $wordResponseContainer->getText();
 
-        if (preg_replace('/[^A-Za-z0-9\-]/', '', $expectedWord) !== preg_replace('/[^A-Za-z0-9\-]/', '', $responseWord)) {
+        if (preg_replace('/\s*/m', '', $expectedWord) !== preg_replace('/\s*/m', '', $responseWord)) {
```
mvorisek:
```suggestion
        if (preg_replace('~\s*~', '', $expectedWord) !== preg_replace('~\s*~', '', $responseWord)) {
```

I highly prefer `~` as regex delimiter. `m` flag completely not needed as `\s` match NL even then.

behat.yml.dist
```diff
@@ -4,7 +4,7 @@ default:
       paths:
         features: '%paths.base%/tests-behat'
       contexts:
-        - atk4\ui\behat\ContextDump
+        - atk4\ui\behat\Context
```
mvorisek, line 7: I am ok with this change as `ContextDump` is not much helpful with AJAX...

behat.yml.dist
```diff
@@ -20,4 +20,4 @@ default:
                 chrome:
                   args:
                     - '--headless'
-                    - '--window-size=1280,720'
+                    - '--window-size=1920,1080'
```
mvorisek: I think we should solve this in Behat testing (I think focus is needed), otherwise tests may fail when testing locally with smalled viewport.

mvorisek (changes requested):

tests-behat/bootstrap/Context.php
```diff
@@ -419,6 +430,139 @@ public function iClickIconUsingCss($arg1)
...
+            throw new \Exception('Data word does not match: ' . $responseWord . ' expected: ' . $expectedWord);
+        }
+    }
+
+    /**
+     * @Then /^I check if input "([^"]*)" match$/
+     */
+    public function iCheckIfInputMatch($name)
+    {
+        $expected = $this->getSession()->getPage()->find('css', '.atk-expected-input-result .content')->getText();
```
mvorisek: I do not see any advantage of this, we should just `I check if input XXX match YYY`

tests-behat/bootstrap/Context.php
```diff
@@ -419,6 +430,139 @@ public function iClickIconUsingCss($arg1)
...
+        $this->assertScopeBuilderRuleExist($name);
+        $idx = ($value === 'Yes') ? 0 : 1;
+        $isChecked = $this->getSession()->evaluateScript('return $(\'[data-name="' . $name . '"]\').find(\'input\')[' . $idx . '].checked');
+        if (!$isChecked) {
+            throw new \Exception('Radio value selected is not: ' . $value);
+        }
+    }
+
+    /**
+     * @Then /^I check if word from data scope match$/
```
mvorisek: The same here, just add the expected text to feature.

mvorisek: or add at least that two selectors as args, ie. check if elem(arg1) value === elem(arg2) value

behat.yml.dist
ibelar: ContextDump is useful when testing locally though.

behat.yml.dist
ibelar: From what I could see, --window-size switch is used when --headless mode is on only.
Then if window size is set too small in --headless mode, then we got 'button not find 'or 'content not find' errors.
Does not happen when --headless is off.

tests-behat/bootstrap/Context.php
ibelar: ok - will try.

tests-behat/bootstrap/Context.php
ibelar: Because it will test if ScopeBuilder is properly building Vue Query Builder (VQB) query. VQB json query object is set by ScopeBuilder using Model condition. It should match value predefine by condition set in test file.

tests-behat/bootstrap/Context.php
ibelar: I added it to the demo file because we can change it without changing the Context file.
Will try to set up using the argument.

tests-behat/bootstrap/Context.php
mvorisek: If you pass the target as selector (like in the next comment), I am ok with it, but Context class should not grow without significant reason for new API.

behat.yml.dist
mvorisek: In non-headless mode, we can not expected that the window will be full HD as well.

We want to test in smaller screen in general (to test if even focusable/scrollable in smaller devices).

tests-behat/bootstrap/Context.php
```diff
@@ -482,41 +482,41 @@ public function scopeBuilderBoolRule($name, $value)
...
+        if (!$compareToContainer) {
+            throw new \Exception('Unable to find compare to container: ' . $compareToSelector);
         }
 
-        $responseWord = $wordResponseContainer->getText();
+        $compareToText = $compareToContainer->getText();
 
-        if (preg_replace('/\s*/m', '', $expectedWord) !== preg_replace('/\s*/m', '', $responseWord)) {
-            throw new \Exception('Data word does not match: ' . $responseWord . ' expected: ' . $expectedWord);
+        if (preg_replace('~\s*~', '', $expectedText) !== preg_replace('~\s*~', '', $compareToText)) {
```
mvorisek: When expected test is generated, when it does not match binary? I think, we do not need this replace... 2x

behat.yml.dist
ibelar: In non-headless, the test pass, no matter what size you set --window-size.
In headless mode, size matters because it seems like window is not scrolling using that mode, which make the test failed. 

If you want to test focusable/scrollable, we should not use headless mode at all.

mvorisek: I just tried to resize non-headless Chrome local browser to size like 200x200 and 3 tests failed:
image

I will test this more...

tests-behat/bootstrap/Context.php
ibelar: done

behat.yml.dist
ibelar: If I am not using --headless (comment it out) and used window-size of 200,200 then callback feature test pass for me locally. It fail using --headless option.

Pass:
                switches:
                  - "--window-size=200,200"

Fail:
                switches:
                  - "--headless"
                  - "--window-size=200,200"

tests-behat/bootstrap/Context.php
mvorisek: also for `compareInputValueToElementText`

behat.yml.dist
mvorisek: https://github.com/atk4/ui/compare/behat_size_debug / https://github.com/atk4/ui/runs/1359529478 with reverted size, it seems to fail for another reason (empty expected text, not bebause something is not clickable)

ibelar: Not using window size option make test pass for headless or non-headless.
May be we should remove "--window-size" completely.

ibelar: > https://github.com/atk4/ui/compare/behat_size_debug / https://github.com/atk4/ui/runs/1359529478 with reverted size, it seems to fail for another reason (empty expected text, not bebause something is not clickable)

yes - exactly why I have increase size. Because the container using 'word' is below the window size if reverted.

mvorisek: because if may use larger size... Current settings are good I think, as it defines size for screenshotting etc.

behat.yml.dist
mvorisek: Yes, but static page issues are discovered even with phpunit... Marking as resolved.

behat.yml.dist
ibelar: Current settings is clearly an issue with headless mode. Besides there is no need for screenshots using headless mode. 
I suggest we remove, or increase, window-size value. 
Local setting can be overrides as well.

mvorisek: I am strictly againt it. This does not address real issue it only increases size.

The sandpit test above also shown, that very unrelated message is shown.

ibelar: there is definitely an issue in Selenium --headless mode with window size. My guess, only a guess, is when we reload content via callback then something is going with browser window an element text is not accessible after callback.
I have issue a workaround for it. See https://github.com/atk4/ui/pull/1469/commits/59f73d38dc0e905b41aa2e3c1102ecf6cb04a8dd

behat.yml.dist
```diff
@@ -20,4 +20,4 @@ default:
                 chrome:
                   args:
                     - '--headless'
-                    - '--window-size=1280,720'
+                    - "--window-size=1280,720"
```
mvorisek:
```suggestion
                    - '--window-size=1280,720'
```

mvorisek: no change, as before...

tests-behat/bootstrap/Context.php
ibelar: test fail without preg_replace for input...

behat.yml.dist
mvorisek: I also wonder why there in no Behat "Stack trace:"...

ibelar:
> no, definitely no, we need to address the real issue

Do you refer to the workaround? If so, then I will simply remove the test for word, because this is only the way I could find beside removing --window-size option...

ibelar:
BTW - why is Behat(latest, Chrome Lowest) take so much time?

mvorisek:
> 
> 
> BTW - why is Behat(latest, Chrome Lowest) take so much time?

because it resolves dependencies, maybe because of composer2, maybe because some dep "expends to large dep tree to explore", I have seen it, I might help later or/and check with composer guys

mvorisek:
> 
> 
> > no, definitely no, we need to address the real issue
> 
> Do you refer to the workaround? If so, then I will simply remove the test for word, because this is only the way I could find beside removing --window-size option...

The fix is horrible in sense of cleaness, so that fix defitely no, I still do not know the real issue, exception is unrelated - the real issue is silence probably as there is JS in between which does not throw...

tests-behat/bootstrap/Context.php
mvorisek: then we expect wrong input ;-) - must be fixed

mvorisek:
> 
> 
> BTW - why is Behat(latest, Chrome Lowest) take so much time?

let's wait until https://github.com/composer/composer/pull/9261 is released, it should solve exactly this issue

behat.yml.dist
ibelar: I have include #1538  configuration in this pr for the test to pass temporarily. Lets' resolve that issue over there.

tests-behat/bootstrap/Context.php
ibelar: Both, $expected and $input-getValue(), contains space char. $expected for readability and $get->input() for "matches regular expression" so we need to remove space in both place for test to pass.

DarkSide666 (approved):
LGTM. Thanks for a hard work!

PR #1470, opened 2020-10-05
Bundle repeatable build JS files

mvorisek:
the current/hosted JS files are very unreliable at least because of hosting issues

this PR also fixes release bundling with split JS build as the new filenames were ignored


.github/workflows/test-unit.yml
```diff
@@ -234,12 +235,16 @@ jobs:
...
+          rm public/agileui.css.orig
 
-      - name: Build JS files (only for Slow)
+      - name: Build/diff JS files (only for Slow)
         if: matrix.type == 'Chrome Slow'
         run: |
+          cp -r public public.orig && rm public/*.js
           (cd js && npm run build)
-          sed -i "s~'https://ui.agiletoolkit.org/public.*~'/demos',~" src/App.php
+          sed -i "s~'https://cdn.jsdelivr.net/gh/atk4/ui@develop/public.*~'/public',~" src/App.php
```
mvorisek, line 245: for some reasons, it worked with `/demos`

.github/workflows/test-unit.yml
```diff
@@ -234,12 +235,16 @@ jobs:
...
+          diff -qr public public.orig
```
mvorisek, line 246: tested, diff working

src/App.php
```diff
@@ -38,7 +38,7 @@ class App
 
     /** @var array|false Location where to load JS/CSS files */
     public $cdn = [
-        'atk' => 'https://ui.agiletoolkit.org/public', // develop branch
+        'atk' => 'https://cdn.jsdelivr.net/gh/atk4/ui@develop/public',
```
mvorisek, line 41: keeping CDN instead of something like `/public`, as it will not work if not inside root or when using rewrite (in the future, we may use Symfony path parser for it)

mvorisek:
@ibelar for develop (non-release), do you want to bundle minified or devel js files?

romaninsh (approved):

PR #1474, opened 2020-10-05
Fix app/owner getters

mvorisek:
for https://github.com/atk4/core/pull/277

as getApp and/or getOwner assert not null, this also highly improve error trace and allow assumptions if not nullable


mvorisek:
@ibelar can you look into the Behat errors - I think, the errors are valid, once parent has app defined, any child should not be created with default/different App first

mvorisek:
@romaninsh please approve this as well

DarkSide666 (approved):

PR #1475, opened 2020-10-06
Fix caching, use raw.githack.com instead of jsdelivr for develop

mvorisek:
raw.githack.com should cache files no more than a few minutes, jsdelivr.com supports purge, but it does not update the source resource until tagged (seems almost a bug there)


georgehristov (approved):
LGTM

PR #1483, opened 2020-10-19
Fix phpdoc - init layout string argument

PhilippGrashoff:
Currently, passing a string argument to App::initLayout() produces an error:
```
TypeError : array_key_exists() expects parameter 2 to be array, string given
 /var/www/html/atk/ui/vendor/atk4/core/src/Factory.php:57
 /var/www/html/atk/ui/vendor/atk4/core/src/Factory.php:105
 /var/www/html/atk/ui/vendor/atk4/core/src/Factory.php:239
 /var/www/html/atk/ui/vendor/atk4/core/src/DiContainerTrait.php:108
 /var/www/html/atk/ui/vendor/atk4/core/src/DiContainerTrait.php:145
 /var/www/html/atk/ui/src/App.php:432
 /var/www/html/atk/ui/tests/AppTest.php:51

```

This PR will fix it. At the moment, there is only a test as proof that it does not work.


tests/AppTest.php
```diff
@@ -6,16 +6,19 @@
...
 
 class AppTest extends AtkPhpunit\TestCase
 {
     protected function getApp()
     {
-        return new App([
-            'catch_exceptions' => false,
-            'always_run' => false,
-        ]);
+        return new App(
```
mvorisek: no :)

tests/AppTest.php
```diff
@@ -41,4 +44,12 @@ public function testTemplateClassCustom()
             $app->loadTemplate('html.html')
         );
     }
+
+    public function testInitLayoutWithStringArgument()
+    {
+        $app = $this->getApp();
+        $app->initLayout(Admin::class);
```
mvorisek: maybe we require standard seed, ie. `[Admin::class]`, thus wont'merge if so

tests/AppTest.php
PhilippGrashoff: ?

tests/AppTest.php
PhilippGrashoff: yes, then this PR would consist of dropping string param. Though, to be honest, a string param seems quite sensible. At least I didn't see code doing more that just passing class or object, e.g. passing extra parameters.

tests/AppTest.php
mvorisek: original CS is ok and better - we do not want code to be longer than needed when readability remains

tests/AppTest.php
mvorisek: no, we have strict convenction about seeds - always an array with 1st param class name

PhilippGrashoff: That makes life easy, just removed in comment that string is a valid parameter

src/App.php
```diff
@@ -423,7 +423,7 @@ public function terminateJson($output, array $headers = []): void
     /**
      * Initializes layout.
      *
-     * @param string|Layout|array $seed
+     * @param Layout|array $seed
```
mvorisek, line 426: 👍

tests/AppTest.php
mvorisek: if you correct this, we can merge the phpdoc change

PhilippGrashoff: Even php CS fixer does not complain about this. But if its so important to you, I will.

tests/AppTest.php
mvorisek: Happy now :)

mvorisek (approved):

PR #1485, opened 2020-10-20
[fix] ScopeBuilder operator for hasOne

ibelar:
Fix issue where hasOne condition is used. This fix will make sure that the proper operator is selected.

```
$model = new Stat($app->db, ['caption' => 'Demo Stat']);
$model->addCondition('client_country_iso', '=', 'Anguilla');
```

Will show in Ui like in picture below where operator dropdown is not set:

Screen Shot 2020-10-20 at 11 34 32 AM

@georgehristov - please have a look and see if this fix is enough or if something else is needed.


src/Form/Control/ScopeBuilder.php
```diff
@@ -218,7 +218,7 @@ class ScopeBuilder extends Control
         'enum' => [
             'type' => 'select',
             'operators' => [
-                self::OPERATOR_EQUALS,
+                self::OPERATOR_TEXT_EQUALS,
```
mvorisek: what is the difference and why the condition below is not updated as well?

ibelar: Was not sure myself, so this is why I asked @georgehristov to take a look.

ibelar: It does fix the issue but we may have to look further.

georgehristov:
I am looking into this. It needs a better fix...

mvorisek:
If this fixes... I would like to see Behat test for this, as ScopeBuilder is completely untested currently.

georgehristov:
Yes, a test is necessary ... will think and work on it

mvorisek (approved):
fix seems clean, but I do not know all internals of the ScopeBuilder class

PR #1486, opened 2020-10-20
[fix] Lookup

ibelar:
Fix Lookup callback url for dropdown menu items to be populated via callback.
Fix demo lookup.


demos/form-control/lookup-dep.php
```diff
@@ -36,13 +36,8 @@
...
     'model' => new Country($app->db),
     'dependency' => function ($model, $data) {
-        $conditions = [];
         foreach (explode(',', $data['starts_with'] ?? '') as $letter) {
-            $conditions[] = ['name', 'like', $letter . '%'];
-        }
-
-        if ($conditions) {
-            $model->addCondition($conditions);
+            $model->addCondition('name', 'like', $letter . '%');
```
mvorisek, line 40: why? this will create AND condition instead of OR

mvorisek, line 40: do we really want AND conditions?

src/Form/Control/Lookup.php
```diff
@@ -147,6 +148,9 @@ protected function init(): void
         $this->initQuickNewRecord();
 
         $this->settings['forceSelection'] = false;
+
+        $this->callback = \atk4\ui\Callback::addTo($this);
```
mvorisek, line 152: callback property is defined, is it set somewhere too?

src/Form/Control/Lookup.php
```diff
@@ -300,7 +304,7 @@ protected function applyLimit($limit = true)
      */
     protected function applySearchConditions()
     {
-        if (!isset($_GET['q'])) {
+        if (!isset($_GET['q']) || empty($_GET['q'])) {
```
mvorisek: only `empty` :)

src/Form/Control/Lookup.php
```diff
@@ -450,4 +451,9 @@ public function set($value = null, $junk = null)
 
         return parent::set($value, $junk);
     }
+
+    protected function mergeStickyArgsFromChildView(): ?AbstractView
```
mvorisek, line 455: do we need this?

src/Form/Control/Lookup.php
mvorisek, line 152: yes, this makes sense, check if any other callback is added outside init somewhere

demos/form-control/lookup-dep.php
ibelar: yes

src/Form/Control/Lookup.php
ibelar: Done!

src/Form/Control/Lookup.php
ibelar: for generating proper callback url, yes.

src/Form/Control/Lookup.php
ibelar: No. Move from renderView to init so url can be generated properly.

mvorisek: yes, but there might be other files with that bad code...

src/Form/Control/Lookup.php
mvorisek: URL of that particial callback (`Lookup->callback`) will always be ok - as this callback is used for API only, I do not think this correct - sticky args will be added from all parents to that callback - otherwise that Lookup will contain sticky args from that callback once called, which is unwanted

please provide usecase

src/Form/Control/Lookup.php
ibelar: no sure of what you mean here.... There is only one callback for Lookup.

mvorisek: yes, but other ui components using `Callback`?

ibelar: I think you had done most of them when you did #1397 - We can fix others, if so when needed.

ibelar: We need this fix ASAP for project....

mvorisek: this takes like 5 minutes and should be parts of every fix - check, if the same bug is not anywhere else, at least in the same project

src/Form/Control/Lookup.php
ibelar: It will be needed when using lookup inside CRUD edit/add form. 
Otherwise, dropdown is always empty because the modal callback terminate before the lookup callback. This is due to lookup callback not having proper url set. I have test it using another project and Lookup is working fine after this fix. 

You can also see for yourself in: https://ui.agiletoolkit.org/demos/collection/actions.php
In Crud, select Actions/Edit and click the Dropdown for "Parent Folder Id" control. Nothing shows. With this fix, parent folder are properly displayed in dropdown control.

src/Form/Control/Lookup.php
ibelar: I did and did not see any other, did you see others?

mvorisek: > I did and did not see any other

great

src/Form/Control/Lookup.php
mvorisek: > I have test it using another project and Lookup is working fine after this fix.

it always will, the question is is it needed? this method was introduced by me, it has a super power and I would like to use it as less as possible, otherwise we can be where we was

> This is due to lookup callback not having proper url set.

I think this is the issue - you probably want to fix your code or use `CallbackLater` (but I still think this is overkill), but not this method

> You can also see for yourself...

I see no difference between this PR and develop:
- add - `Parent Folder Id` dropdown in CRUD form is filled
- edit - `Parent Folder Id` dropdown in CRUD form is NOT filled

ibelar: CallbackLater will not do any good either. The problem is from generating the proper URL for Callback in order to execute it when it is triggered.

Thus, I do not see any other alternative than using mergeStickyArgsFromChildView(). Besides, there were other cases in a specific project when Callback was needed for a View and that I had to use this method too for generating proper callback URL. 

Should we revert #1397 and keep app sticky argument? The way I see it, any future View using a Callback where proper URL needs to be generated might has to need mergeStickyArgsFromChildView(). It is true for Modal, Loader, Popup, Panel/Content, Wizard and now Lookup...

mvorisek: > Should we revert #1397 and keep app sticky argument?

never! This PR is very critical to support true component tree.

If the `mergeStickyArgsFromChildView` method is really needed here, add the most simpletest Behat test for it and let's merge it. This way, we can be sure the functionality is fine and have still a testable option to remove this method in the future.

ibelar:
I can add the test but do we have a predictable set of data in demos that can return predictable model->hasOne result that we can use in Crud/edit?

Envoyé de mon iPhone

mvorisek:
> I can add the test but do we have a predictable set of data in demos that can return predictable model->hasOne result that we can use in Crud/edit?

if not, simply edit `demos/_demo-data/dump.sql` (in minimalistic way, as we transform this data with regex)

src/Form/Control/Lookup.php
ibelar: Test added!

mvorisek: ❤️! One CI job failed - https://github.com/atk4/ui/pull/1486/checks?check_run_id=1287604565 - probably some issue with detection if load finished or not. Once fixed, you can 🚀 from my side

mvorisek (approved):

PR #1489, opened 2020-10-20
Use proper boolean decorator for CardTable

DarkSide666:


src/CardTable.php
```diff
@@ -45,7 +45,13 @@ public function setModel(Model $model, $columndef = null)
         $mm = parent::setSource($data);
         $this->addDecorator('value', [Table\Column\Multiformat::class, function ($row, $field) use ($model) {
             $field = $model->getField($row->data['id']);
-            $ret = $this->decoratorFactory($field);
+            if ($field->type === 'boolean') {
+                $ret = $this->decoratorFactory($field,[
+                    Table\Column\Status::class, ['positive' => [true, 'Yes'], 'negative' => [false, 'No']]
+                ]);
+            } else {
+                $ret = $this->decoratorFactory($field);
```
mvorisek: I prefer
```
,
$field->type === 'boolean' ? [Table\Column\Status::class, ['positive' => [true, 'Yes'], 'negative' => [false, 'No']]] : []
```

ie. no code duplication

DarkSide666: ok

PR #1490, opened 2020-10-20
Add True Reload feature

mvorisek:
in the future, we should fix all components and allow true reload for all of them

help welcomed

How to test:
1. open `/demos/javascript/reloading.php`
2. click `Reload but keep custom changes`
3. notice, that the input value stay unchanges
4. if you do not belive, change it manually again and click reload (step 3)
5. if you do not belive even more, open inspector, change the number after the input field -  and click reload (step 3)


demos/javascript/reloading.php
```diff
@@ -19,6 +19,72 @@
 $v = View::addTo($app, ['ui' => 'segment'])->set((string) random_int(1, 100));
 Button::addTo($app, ['Reload random number'])->js('click', new JsReload($v, [], new JsExpression('console.log("Output with afterSuccess");')));
 
+// Reload but keep custom changes
+\Atk4\Ui\Header::addTo($app, ['Button reloading View without loosing original values']);
+$v = \Atk4\Ui\View::addTo($app)->set((string) random_int(1, 1000));
+$inputControl = \Atk4\Ui\Form\Control\Input::addTo($v);
+\Atk4\Ui\View::addTo($app)->js(true, null, $inputControl)->find('input')->val('test ' . (string) random_int(1, 1000)); // simulate change by user
+
+$app->requireJs($app->cdn['atk'] . '/external/diff-dom/browser/diffDOM.js'); // use probably versioned CDN link here
```
mvorisek, line 28: related https://github.com/fiduswriter/diffDOM/issues/84

mvorisek:
maybe one day in the future :)

PR #1492, opened 2020-10-21
Remove unused/old JsUniv

mvorisek:
maybe old, maybe I do not understand the purpose or used in some other projects

but there is no `univ.min.js` in this repo


ibelar (approved):
Yes - Very old!

PR #1493, opened 2020-10-21
Test using lowest deps

mvorisek:


ibelar (approved):
Not sure I understand everything but LGTM 😄

PR #1494, opened 2020-10-21
[fix] Behat Lookup test

ibelar:
Make sure the dropdown menu is fully hidden and not covering the EditMe button in the Lookup test.

### DIFF
--- a/demos/_unit-test/lookup.php
+++ b/demos/_unit-test/lookup.php
@@ -1,6 +1,10 @@
 <?php
 
 declare(strict_types=1);
+/**
+ * Behat testing.
+ * Test for hasOne Lookup as dropdown control.
+ */
 
 namespace atk4\ui\demo;
 
@@ -21,6 +25,7 @@ $edit->callback = function ($model) use ($app) {
 
 $crud = Crud::addTo($app);
 
+// Properly set control for Behat test.
 $crud->onFormEdit(function ($f) {
     $f->getControl('product_category_id')->settings['duration'] = 0;
     $f->getControl('product_category_id')->settings['delay'] = ['hide' => 0, 'search' => 0];
--- a/demos/_unit-test/lookup.php
+++ b/demos/_unit-test/lookup.php
@@ -20,4 +20,14 @@ $edit->callback = function ($model) use ($app) {
 };
 
 $crud = Crud::addTo($app);
+
+$crud->onFormEdit(function($f){
+    $f->getControl('product_category_id')->settings['duration'] = 0;
+    $f->getControl('product_category_id')->settings['delay'] = ['hide' => 0, 'search' => 0];
+
+    $f->getControl('product_sub_category_id')->settings['duration'] = 0;
+    $f->getControl('product_sub_category_id')->settings['delay'] = ['hide' => 0, 'search' => 0];
+
+});
+
 $crud->setModel($model, ['name']);
--- a/demos/_unit-test/lookup.php
+++ b/demos/_unit-test/lookup.php
@@ -29,10 +29,12 @@ $crud = Crud::addTo($app);
 $crud->onFormEdit(function ($f) {
     $f->getControl('product_category_id')->settings['duration'] = 0;
     $f->getControl('product_category_id')->settings['delay'] = ['hide' => 0, 'search' => 0];
+    // Important to reduce control width because it could cause Behat test to fail if over EditMe button.
     $f->getControl('product_category_id')->setStyle(['width' => '50%']);
 
     $f->getControl('product_sub_category_id')->settings['duration'] = 0;
     $f->getControl('product_sub_category_id')->settings['delay'] = ['hide' => 0, 'search' => 0];
+    // Important to reduce control width because it could cause Behat test to fail if over EditMe button.
     $f->getControl('product_sub_category_id')->setStyle(['width' => '50%']);
 });
 
--- a/demos/_unit-test/lookup.php
+++ b/demos/_unit-test/lookup.php
@@ -21,13 +21,12 @@ $edit->callback = function ($model) use ($app) {
 
 $crud = Crud::addTo($app);
 
-$crud->onFormEdit(function($f){
+$crud->onFormEdit(function ($f) {
     $f->getControl('product_category_id')->settings['duration'] = 0;
     $f->getControl('product_category_id')->settings['delay'] = ['hide' => 0, 'search' => 0];
 
     $f->getControl('product_sub_category_id')->settings['duration'] = 0;
     $f->getControl('product_sub_category_id')->settings['delay'] = ['hide' => 0, 'search' => 0];
-
 });
 
 $crud->setModel($model, ['name']);
--- a/demos/_unit-test/lookup.php
+++ b/demos/_unit-test/lookup.php
@@ -24,9 +24,11 @@ $crud = Crud::addTo($app);
 $crud->onFormEdit(function ($f) {
     $f->getControl('product_category_id')->settings['duration'] = 0;
     $f->getControl('product_category_id')->settings['delay'] = ['hide' => 0, 'search' => 0];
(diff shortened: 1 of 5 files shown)


mvorisek:
@ibelar the 20 ms animation delay seems not to be the issue, see `try_lower_animation_time` branch. I am mentioning this branch mainly as you can cherrypick the latest commit here for testing and verify, if the fix fixes consistently.

PhilippGrashoff:
I had issues like that a lot, too. What I did was wrapping all these things in wait()->until(), e.g.:
```


    /*
     * tries to click element until click succeeds
     */
    public function tryClick(string $css_selector) {
        self::$webDriver->wait($this->waitTimeOut, $this->waitInterval)->until(
            function() use ($css_selector) {
                try {
                    $this->findByCSS($css_selector)->click();
                    return true;
                }
                catch(Exception $e) {
                    return false;
                }
            }
        );
    }
```

This way, selenium tries to click the element until it either is clickable or a timeout occurs.

mvorisek:
> What I did was wrapping all these things in wait()->until()

we want to primary identify the issue as this will lead to unnessary delays in general

however for the dropdown this solution might be ok, but adjust the try/catch, so specific exception is caught and throw on timeout (important for debug)

PhilippGrashoff:
Hi,

yes that code is a bit old and can use some improvement: catching only the Exception type that is thrown when the element to click is overlaid by something.

Regarding the timeout, thats something the wait() function takes care of. No need to implement that in the function.

mvorisek:
every single call to browser costs some time, not mentioning that wait wait in 100ms increments (I fixed that in Mink some months ago, but it is not released yet)

PhilippGrashoff:
you can set the wait interval - at least in selenium. I lowered it to 25ms in my tests. Of course this solution takes some more time, but its an easy way to get things like dropdowns working. I ended up doing all my clicks using this function as I use a lot of dropdowns and frequently ran into "element not clickable" problems.

mvorisek:
> you can set the wait interval - at least in selenium. I lowered it to 25ms in my tests.

no, you can not - see https://github.com/minkphp/Mink/pull/801 and https://github.com/minkphp/MinkSelenium2Driver/pull/322 , if there is another setting we can lower, please let me know or send PR directly

> I ended up doing all my clicks using this function as I use a lot of dropdowns and frequently ran into "element not clickable" problems.

in past, I fixed everything without need for this, which is the correct way, the main issue are animations, which has been fixed, we do not want to destroy all this!

PhilippGrashoff:
> no, you can not - see [minkphp/Mink#801](https://github.com/minkphp/Mink/pull/801) and [minkphp/MinkSelenium2Driver#322](https://github.com/minkphp/MinkSelenium2Driver/pull/322) , if there is another setting we can lower, please let me know or send PR directly
No idea about mink, was talking about selenium only.

> in past, I fixed everything without need for this, which is the correct way, the main issue are animations, which has been fixed, we do not want to destroy all this!
I didn't mean that we should do this everywhere just because I did this. I just offered a solution I found for a similar problem I often faced already.

mvorisek:
@ibelar I tried to remove `mergeStickyArgsFromChildView` from Lookup and all Behat tests pass (incl. new `tests-behat/lookup.feature`) - do we really need that method or can you adjust the tests that when removed, the tests fail?

ibelar:
Just realized that Fomantic Dropdwown module was causing js error in v2.8.6 when selecting a value.

Screen Shot 2020-10-22 at 9 56 22 AM

I have update Fomantic to 2.8.7 and error disappear, so let's see if this was causing our test error.

demos/_unit-test/lookup.php
```diff
@@ -20,4 +20,13 @@
 };
 
 $crud = Crud::addTo($app);
+
+$crud->onFormEdit(function ($f) {
```
mvorisek, line 27: tests are still failing

I think we should not need this, as I override this for testing **globally** in:
https://github.com/atk4/ui/blob/72e19a60edd150141890e6631276f0925da3449c/tests-behat/bootstrap/Context.php#L62

PS: I already tried to set `none` - without any change

if dropdown is loaded from JS and you do more than one operation in step, you should probably call `jqueryWait` between them

think of this outside unit test - `iSelectValueInLookup` should work without **any** extra sleep/wait anywhere in features - ie. modifying the tested code manually like this for testing is not ok

ibelar: It is still possible that Dropdown module override delay via js.

tests-behat/bootstrap/Context.php
```diff
@@ -67,7 +67,7 @@ protected function disableAnimations(): void
             'animation-duration' => '5s',
             'transition-duration' => '5s',
         ]);
-        $script = '$(\'<style>' . $css . '</style>\').appendTo(\'head\'); $.fx.off = true;';
+        $script = 'if (Array.prototype.filter.call(document.getElementsByTagName("Style"), e => e.getAttribute("about") === "atk-test-behat").length === 0) {$(\'<style about="atk-test-behat">' . $css . '</style>\').appendTo(\'head\'); $.fx.off = true;}';
```
mvorisek: https://stackoverflow.com/questions/14017864/how-to-check-if-a-tag-exists-using-javascript-without-getting-an-error

`if () {}` seems much simpler, also `getElementsByTagName("style")` (style, not Style)

otherwise 👍

demos/_unit-test/lookup.php
mvorisek: yes, but try locally (where you can see) - for me, it seems this disables the animations

tests-behat/bootstrap/Context.php
mvorisek: was this the issue or is it just randomly fixed?

running https://github.com/atk4/ui/tree/test_behat_fix more tests...

UPDATE: still failing, tests passed only randomly...

(but still of course keep your change, we should add style only once)

ibelar: yeah, need more testing. 

getElementsByTagName("style").length === 0 will always be false. What we need to check if there is a style tag using atk-test-behat attribute. I will fix typo. 

does https://github.com/atk4/ui/tree/test_behat_fix has the latest changes?

mvorisek: yes, it has - see git tree, feel free to use this branch for testing, ie. merge this branch there, push, and recheck the CI results for that branch, we will delete it once fixed

tests-behat/bootstrap/Context.php
```diff
@@ -450,7 +450,7 @@ protected function getFinishedScript(): string
         return 'document.readyState === \'complete\''
             . ' && typeof $ !== \'undefined\' && jQuery.active === 0'
             . ' && typeof atk !== \'undefined\' && atk.vueService.areComponentsLoaded()'
-            . ' && 0 === jQuery(\':animated\').length)';
+            . ' && 0 === jQuery(\':animated\').length';
```
mvorisek: I think this is not needed with `$.fx.off = true;`, as JQuery JS animations are disabled

ibelar: understand but there is also no guarantee that a Fomantic module is not setting it back to false when initializing.

mvorisek: unless it does - remove, because it is very unexpected it do so

ibelar: why remove? this is just a safety test. besides, all seem to be working well now. The reason I added this test was that when some test was failing, it was because it could not find one of the dropdown control, sometimes it was the first one, sometimes it was the second one. Adding this test seems to fix the issue.

mvorisek: > Adding this test seems to fix the issue.

I am doing testing now, if this is needed, of course, it will stay. Otherwise I think it is duplicated to the jQuery setting as libs that set it basically all use jQuery :)

mvorisek (approved):

PR #1498, opened 2020-10-22
Disable Lookup test until fixed reliably

mvorisek:
https://github.com/atk4/ui/pull/1494 did not fixed

see https://github.com/atk4/ui/tree/behat_debug


ibelar (approved):
Ok

PR #1499, opened 2020-10-22
Fix/enable Lookup Behat test

mvorisek:
merge after https://github.com/atk4/ui/pull/1525 and https://github.com/atk4/ui/pull/1526

Lookup is sending AJAX requests and after every operation with it we must wait.


tests-behat/bootstrap/Context.php
```diff
@@ -380,7 +380,7 @@ public function iSelectValueInLookup($arg1, $arg2)
         $script = '$("#' . $lookup->getAttribute('id') . '").dropdown("show")';
         $this->getSession()->executeScript($script);
         // Wait till dropdown is visible
-        // Cannot call jqueryWait because calling it will return prior from dropdown to fire ajax request.
+        $this->jqueryWait();
```
mvorisek: removed comment no longer valid - `jqueryWait` checks multiple times, thus AJAX should be already fired at that time

tests-behat/bootstrap/Context.php
```diff
@@ -452,8 +452,7 @@ protected function getFinishedScript(): string
     {
         return 'document.readyState === \'complete\''
             . ' && typeof jQuery !== \'undefined\' && jQuery.active === 0'
-            . ' && typeof atk !== \'undefined\' && atk.vueService.areComponentsLoaded()'
-            . ' && jQuery(\':animated\').length === 0';
```
mvorisek: everything is tweaked to not need it, so we want to keep that clean state if possible

demos/_unit-test/lookup.php
```diff
@@ -26,8 +26,6 @@
 // Important for Behat testing
 $crud->onFormEdit(function ($f) {
     foreach (['product_category_id', 'product_sub_category_id'] as $controlName) {
-        $f->getControl($controlName)->settings['duration'] = 0;
```
mvorisek, line 29: seems also not helful

demos/_unit-test/lookup.php
```diff
@@ -26,8 +26,6 @@
...
-        $f->getControl($controlName)->settings['delay'] = ['hide' => 0, 'search' => 0];
         // reduce control width because it could cause Behat test to fail if over EditMe button
         $f->getControl($controlName)->setStyle(['width' => '50%']);
```
mvorisek, line 30: I am ok with this to stay for now, I think the issue is it can not fit completely into the viewport at least I already faced simillar issue

demos/_unit-test/lookup.php
```diff
@@ -22,15 +22,4 @@
 };
 
 $crud = Crud::addTo($app);
-
-// Important for Behat testing
-$crud->onFormEdit(function ($f) {
```
mvorisek, line 27: completely not needed...

tests-behat/bootstrap/Context.php
```diff
@@ -242,7 +242,7 @@ public function iDontSee($arg1)
      */
     public function labelChangesToNumber()
     {
-        $this->getSession()->wait(5000, '!$("#' . $this->buttonId . '").hasClass("loading")');
+        $this->jqueryWait('!$("#' . $this->buttonId . '").hasClass("loading")');
```
mvorisek: `jqueryWait` not needed here stictly, but this way, we wait + some delay for AJAX to stabilize (ie. all requests complete)

even better would be to modify the Behat driver, so we do not have to care about AJAX waits at all, but that is quote complex to do...

tests-behat/bootstrap/Context.php
```diff
@@ -426,19 +425,6 @@ public function iClickIconUsingCss($arg1)
...
-     * Wait for an element, usually an auto trigger element, to show that loading has start"
-     * Example, when entering value in JsSearch for grid. We need to auto trigger to fire before
-     * doing waiting for callback.
-     * $arg1 should represent the element selector for jQuery.
-     *
-     * @Then I wait for loading to start in :arg1
-     */
-    public function iWaitForLoadingToStartIn($arg1)
-    {
-        $this->getSession()->wait(2000, '$("' . $arg1 . '").hasClass("loading")');
```
mvorisek, line 439: very badly designed check - be must wat until loading is complete, not started (which we may miss completely)

tests-behat/grid.feature
```diff
@@ -5,7 +5,6 @@ Feature: Grid
     Given I am on "collection/grid.php"
     Then I search grid for "kingdom"
 #    make sure auto query trigger
-    And I wait for loading to start in "button.atk-search-button"
```
mvorisek, line 8: @ibelar this is failing, is debounce really disabled?

tests-behat/bootstrap/Context.php
```diff
@@ -386,6 +386,9 @@ public function iSelectValueInLookup($arg1, $arg2)
 
         // hide dropdown and wait till fully closed
         $this->getSession()->executeScript('$("#' . $lookupElem->getAttribute('id') . '").dropdown("hide");');
+        $this->jqueryWait();
+        // for unknown reasons, dropdown very often remains visible in CI, so hide twice
+        $this->getSession()->executeScript('$("#' . $lookupElem->getAttribute('id') . '").dropdown("hide");');
```
mvorisek, line 392: This was the final CI issue - dropdown keeps open/reopen after closed from JS - @ibelar please report to FUI if not related with https://github.com/atk4/ui/issues/1521 , ie. fixed by https://github.com/fomantic/Fomantic-UI/pull/1725

tests-behat/bootstrap/Context.php
```diff
@@ -242,7 +241,6 @@ public function iDontSee($arg1)
      */
     public function labelChangesToNumber()
     {
-        $this->getSession()->wait(5000, '!$("#' . $this->buttonId . '").hasClass("loading")');
```
mvorisek, line 240: not needed, because we call `jQueryWait` after each step, ie. the first statement in a step should never check load - it should be always loaded

tests-behat/bootstrap/Context.php
```diff
@@ -452,23 +444,20 @@ protected function getFinishedScript(): string
     {
         return 'document.readyState === \'complete\''
             . ' && typeof jQuery !== \'undefined\' && jQuery.active === 0'
-            . ' && typeof atk !== \'undefined\' && atk.vueService.areComponentsLoaded()'
-            . ' && jQuery(\':animated\').length === 0'; // needed for SUI dropdown
```
mvorisek, line 446: no `:animated` or `.loading` class condition needed, which is great, as these can slow down this check

tests-behat/bootstrap/Context.php
```diff
@@ -480,7 +469,7 @@ protected function jqueryWait($duration = 5000)
             }
         }
 
-        throw new \Exception('JQuery did not finished within a given time limit');
+        throw new \Exception('jQuery did not finished within a time limit');
```
georgehristov, line 464: Same here: jQuery did not finish within the time limit

georgehristov (approved):

tests-behat/bootstrap/Context.php
mvorisek: this PR has that PR merged in, so will be resolved once merged into develop later itself...

PR #1500, opened 2020-10-23
Trigger Bundler CI using better filter

mvorisek:
currently:
image

Bundler is run, only some jobs are filtered

we should prevent Bundler to run at all and trigger only on specific branch


ibelar (approved):
LGTM

PR #1501, opened 2020-10-23
Patch Behat until behat/mink is released to improve testing speed

mvorisek:
|         | before | after |
|---------|--------|-------|
| Chrome  | 1m 26s | 51s (with 1 ms wait about 44s)   |
| Firefox | 1m 40s | 1m 7s |

stats based on https://github.com/atk4/ui/runs/1296933654 and https://github.com/atk4/ui/runs/1296942796 runs

tested even with 100us wait: https://github.com/atk4/ui/runs/1297037856

### DIFF
--- a/tests-behat/bootstrap/Context.php
+++ b/tests-behat/bootstrap/Context.php
@@ -59,10 +59,10 @@ class Context extends RawMinkContext implements BehatContext
         };
 
         $css = $toCssFx('*', [
-            'animation-delay' => '0.02s',
-            'animation-duration' => '0.02s',
-            'transition-delay' => '0.02s',
-            'transition-duration' => '0.02s',
+            'animation-delay' => '0.005s',
+            'animation-duration' => '0.005s',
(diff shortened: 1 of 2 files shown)


tests-behat/bootstrap/Context.php
```diff
@@ -59,10 +59,10 @@ protected function disableAnimations(): void
         };
 
         $css = $toCssFx('*', [
-            'animation-delay' => '0.02s',
-            'animation-duration' => '0.02s',
-            'transition-delay' => '0.02s',
-            'transition-duration' => '0.02s',
+            'animation-delay' => '0.005s',
```
mvorisek, line 62: improve stability, needed for modal.feature in about 1 of 20 tests

with this fix, it seems hundreds of tests can be run in a row without an issue

I belive, we do not want zero/none here, as it will prevent JS animatins events to trigger (but tests still pass)

ibelar (approved):
LGTM

PR #1503, opened 2020-10-23
Js Build using Webpack Copy plugin

ibelar:
Will copy all file in resources folder into plublic folder.


mvorisek:
SUI requires specific google font with absolute link

please come with a CLEAN solution to copy every resource needed by browser/client or let's close this for now

mvorisek:
closing as this does not solve the major issue with links in CSS files like fonts on Google or other CDNs

PR #1505, opened 2020-10-24
Revert "Trigger Bundler CI using better filter"

mvorisek:
see https://github.community/t/trigger-job-on-branch-created/16878/5 , there is no branch trigger filter for create

must be merged asap to prevent bad commit and branches to be created


georgehristov (approved):
LGTM

PR #1507, opened 2020-10-24
Rename Template to HtmlTemplate

mvorisek:
and remove code scheduled to be removed in the 2.4 version


georgehristov (approved):
👍

DarkSide666:
Actually `Template` class can be used not only for HTML templates ...
So this looks again just renaming something and creating BC breaks without any advantage for developer.

mvorisek:
> 
> 
> Actually `Template` class can be used not only for HTML templates ...
> So this looks again just renaming something and creating BC breaks without any advantage for developer.

I am working on much safer and OOP Template. The Template can currently work for "non HTML" - but you have to set everything using the `setHtml` method :D. But I am ok to revert the name with the #1509 and #1508.

DarkSide666:
No I guess it's fine. No big deal.

PR #1508, opened 2020-10-25
Rework HtmlTemplate from nested array to tree/OOP structure

mvorisek:
values are newly linked HtmlTemplate\Value objects and instanced only once per tag name

IMPORTANT:
- tree is dependent structure of parent template, ie. it can be modified with set/append/del operation on the parent template
- if another full template (instead of tree) is added, it is never affected with the owning template (except it can be deleted by it)
- new `Template::toLoadableString()` method is added
- data are checked if UTF-8 valid

BC:
- for user, all works like before
- except silently allowed bad templates with tag defined multiple times with different content - see changes (probably mistakes)
- `Template::get()` method is removed, the most closest equivalent is `Template::getTagTree()`, but the result type/structure is different
- `Template::eachTag()` method is removed as no more than one value per tag can exists - use `Template::getTagTree()` and walk/update tag tree as needed

PERFORMANCE:
- about the same in real world (parse/cache clone is about 4x slower, setters/getters are about 2x faster, render is about 2x slower)

`loadFromString`:
image

`renderToHtml`
image


DarkSide666 (commented):
If it performs the same or better than current implementation, then I like it :)

demos/collection/jssortable.php
```diff
@@ -13,7 +13,7 @@
     '<div class="ui header">Click and drag country to reorder</div>
     <div id="{$_id}" style="cursor: pointer">
         <ul>
-            {List}<li class="ui icon label" data-name="{$name}"><i class="{iso}ae{/} flag"></i> {name}andorra{/}</li>{/}
+            {List}<li class="ui icon label" data-name="{$name}"><i class="{iso}ae{/} flag"></i> {$name}</li>{/}
```
mvorisek, line 16: old bug: tag defined twice with different content

demos/layout/templates/layout1.html
```diff
@@ -123,12 +123,12 @@ <h4>Ungrouped transaction list</h4>{Report}
...
         <tfoot>
           <tr>
-            <th class="right aligned" colspan="6">Closing balance after {count}29{/} records:</th>
+            <th class="right aligned" colspan="6">Closing balance after {foot_count}29{/} records:</th>
             <th>€1,056.91</th>
             <th>€0.00</th>
           </tr>
           <tr>
-            <th class="right aligned" colspan="6">Total for {nominal}Sales:Limerick{/nominal}:</th>
+            <th class="right aligned" colspan="6">Total for {foot_nominal}Sales:Limerick{/}:</th>
```
mvorisek, line 131: old bug: tag defined twice with different content

docs/template.rst
```diff
@@ -717,8 +717,8 @@ under ``$template->template`::
 Property tags would contain::
 
     array (
-      'subject'=> array( &array ),
-      'subject#0'=> array( &array )
+      'subject#0'=> array( &array ),
```
mvorisek, line 720: old typo

tests/HtmlTemplateTest.php
```diff
@@ -7,253 +7,168 @@
...
-    /**
-     * Test set() exception.
-     */
-    public function testSetException1()
+    public function testSetBadTypeException()
     {
         $t = new HtmlTemplate('{foo}hello{/} guys');
         $this->expectException(Exception::class);
-        $t->set('qwe', 'Hello'); // not existent tag
+        $t->set('foo', new \StdClass());
```
mvorisek:
```suggestion
        $t->set('foo', new \stdClass());
```

demos/layout/templates/layout1.html
PhilippGrashoff: If we are reworking this: Usually Regions are written with a capital letter `{Count}`, whereas Tags are written lowercase `{$something}`. Not mandatory, but good style.

mvorisek: both are exactly the same :D

PhilippGrashoff: yes, technically there is no difference, but in most ATK templates regions are capital, while pure tags (with a $) are written lowercase. As I said, not mandatory, but imo good style.

mvorisek:
@georgehristov can you please review this PR?

georgehristov (approved):
LGTM

PR #1512, opened 2020-10-28
Fix Template::cloneRegion

mvorisek:
this is a bug, `_top` key should not be added (and is not set before cloned)


georgehristov (approved):
LGTM

PR #1513, opened 2020-10-28
Deprecate Template array like access

mvorisek:
array access does not make a lot of sense with https://github.com/atk4/ui/pull/1508 as child objects can not be array accessed

allow, append/try methods were never (and can not be) array accessible, thus removing support completely


docs/template.rst
```diff
@@ -75,10 +71,10 @@ There are some operations you can do with a region, such as::
 
     $main_template->del('Content');
 
-    $content->set(['user' => 'Joe', 'amount' => 100]);
+    $content->set('user', 'Joe')->set('amount', 100);
     $main_template->dangerouslyAppendHtml('Content', $content->renderToHtml());
 
-    $content->set(['user' => 'Billy', 'amount' => 50]);
+    $content->set('user', 'Billy')->set('amount', 50);
```
georgehristov, line 77: Would be good to have `setMulti` method for this

georgehristov (approved):

docs/template.rst
mvorisek: we still support `set([])`, I only updated it to chained methods to make the example better

PR #1515, opened 2020-10-28
Fix bug 1514 - ipp variable must be int

### LINKED ISSUE #1514: [Critical Error][BUG] IPP Bug in https://ui.agiletoolkit.org/demos/collection/grid.php
Hi!

Try changing the IPP dropdown on the page...

you get the following error:

Critical Error
TypeError: Argument 1 passed to atk4\data\Model::setLimit() must be of the type int or null, string given, called in src/Grid.php on line 659

mvorisek:
fixes https://github.com/atk4/ui/issues/1514

### DIFF
--- a/src/ItemsPerPageSelector.php
+++ b/src/ItemsPerPageSelector.php
@@ -76,7 +76,7 @@ class ItemsPerPageSelector extends View
     public function onPageLengthSelect(\Closure $fx)
     {
         $this->cb->set(function () use ($fx) {
-            $ipp = $_GET['ipp'] ?? null;
+            $ipp = isset($_GET['ipp']) ? (int) $_GET['ipp'] : null;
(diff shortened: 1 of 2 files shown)


ibelar (approved):
LGTM

src/ItemsPerPageSelector.php
```diff
@@ -76,7 +76,7 @@ public function jsSetLabel($ipp)
     public function onPageLengthSelect(\Closure $fx)
     {
         $this->cb->set(function () use ($fx) {
-            $ipp = $_GET['ipp'] ?? null;
+            $ipp = isset($_GET['ipp']) ? (int) $_GET['ipp'] : null;
```
mvorisek, line 79: this is need to fix the bug, other fixes are theoretical

PR #1518, opened 2020-10-28
[fix] Multitable demo

### LINKED ISSUE #1516: Demo bug - entity ID reassign
step to reproduce:

1. open `/demos/collection/multitable.php`
2. click any item in the table
3. you will see:
image

ibelar:
fix #1516

### DIFF
--- a/demos/collection/multitable.php
+++ b/demos/collection/multitable.php
@@ -38,7 +38,7 @@ $finderClass = get_class(new class() extends \atk4\ui\Columns {
 
         while ($selections && $id = array_shift($selections)) {
             $path[] = $id;
-            $pushModel = $model->newInstance();
+            $pushModel = clone $model;
             $pushModel->tryLoad($id);
             if (!$pushModel->loaded()) {
                 break;
--- a/demos/collection/multitable.php
+++ b/demos/collection/multitable.php
@@ -38,8 +38,9 @@ $finderClass = get_class(new class() extends \atk4\ui\Columns {
 
         while ($selections && $id = array_shift($selections)) {
             $path[] = $id;
-            $model->tryLoad($id);
-            if (!$model->loaded()) {
+            $pushModel = $model->newInstance();
+            $pushModel->tryLoad($id);
+            if (!$pushModel->loaded()) {
                 break;
             }
             $ref = array_shift($route);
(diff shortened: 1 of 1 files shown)


demos/collection/multitable.php
```diff
@@ -38,23 +38,24 @@ public function setModel(\atk4\data\Model $model, $route = [])
 
         while ($selections && $id = array_shift($selections)) {
             $path[] = $id;
-            $model->tryLoad($id);
-            if (!$model->loaded()) {
+            $pushModel = $model->newInstance();
```
mvorisek:
```suggestion
            $pushModel = clone $model;
```

otherwise we will loose scope if set (or other settings not set in `Model::init()`)

@georgehristov should we remove `newInstance()` from Model or change it, so clone is returned instead?

mvorisek (approved):

demos/collection/multitable.php
georgehristov: clone $model != $model->newInstance

Using newInstance is not supposed to take over the scope but only use the same connection.

mvorisek: ok - it is not clone, as it does not clone ID etc.

so `$obj->newInstance()` is strictly equivalent to `new get_class($obj)()` (only theoretical, new expr is not supported)?

if model is dynamic like now, ie. defined with scopes, what is the advantage of this function?

georgehristov: As I mentioned - also the same connection is used. Actual code speaks for itself:

```
public function newInstance(string $class = null, array $options = [])
    {
        $model = (self::class)::fromSeed([$class ?? static::class], $options);

        if ($this->persistence) {
            return $this->persistence->add($model);
        }

        return $model;
    }
```

ibelar: @mvorisek - clone $model is not fully working here too. It prevents from going deeper if a folder is click in right column. You should be able to click on the rightmost columns until you see no records. Was working using $model->newInstance()

ibelar: this is how it should look
Screen Shot 2020-10-29 at 12 11 32 PM

mvorisek: @ibelar I tested yesterday, I almost wanted to submit an issue... But it works! Ok - see https://github.com/atk4/ui/pull/1523 , we simply revert my change.

PR #1524, opened 2020-10-29
Fix debouce for Behat testing (fake jQuery.active)

mvorisek:


tests-behat/bootstrap/Context.php
```diff
@@ -426,19 +427,6 @@ public function iClickIconUsingCss($arg1)
...
 
-    /**
-     * Wait for an element, usually an auto trigger element, to show that loading has start"
-     * Example, when entering value in JsSearch for grid. We need to auto trigger to fire before
-     * doing waiting for callback.
-     * $arg1 should represent the element selector for jQuery.
-     *
-     * @Then I wait for loading to start in :arg1
-     */
-    public function iWaitForLoadingToStartIn($arg1)
```
mvorisek, line 437: very badly designed check - be must wat until loading is complete, not started (which we may miss completely)

tests-behat/grid.feature
```diff
@@ -5,7 +5,6 @@ Feature: Grid
     Given I am on "collection/grid.php"
     Then I search grid for "kingdom"
 #    make sure auto query trigger
-    And I wait for loading to start in "button.atk-search-button"
```
mvorisek, line 8: @ibelar this is failing, is debounce really disabled?

ibelar: Yes, debounce is disable. You should probably call $this-jqueryWait() after $search->value($arg1) because the search will query the server with a callback and you are not before or after step.

mvorisek: I already do
https://github.com/atk4/ui/blob/6e749d0a02d2e1b928a4dd48212e6994c8bbbe0d/tests-behat/bootstrap/Context.php#L36-L41
 that is why I am asking :)

ibelar: or probably because 'keyup' listener function is set on page load thus when setting debounce value afterward, it has no effect. Let me check...

ibelar: That is the reason, if we set debounce value prior to initializing the plugin then it is fine.

Suggestion:
- drop grid.feature since we now have crud.feature that does the same and more;
- in crud.feature we control the file so we can add this line of code: `$app->html->js(true, new JsExpression('atk.options.set("debounceTimeout", 20)'));` in _unit-test/crud.php file

Adding this line will make the test pass.

mvorisek: `debounceTimeout` should be resolved INSIDE callback - https://github.com/atk4/ui/blob/develop/js/src/agile-toolkit-ui.js#L25 - so value should be resolved once called

The debounce code is here: https://github.com/component/debounce/blob/master/index.js , maybe we can tweak debound in atk more if needed or completely copy and modify that impl.

mvorisek: `jqueryWait` waits at least 20 ms, why do we need `iWaitForLoadingToStartIn`, what causes such large delay before `jQuery.active` is non-zero, eg. detectable?

.github/workflows/test-unit.yml
```diff
@@ -220,6 +220,7 @@ jobs:
       - name: Install JS dependencies (only for Slow)
         if: matrix.type == 'Chrome Slow'
         run: |
+          git config --global url."https://github.com".insteadOf "ssh://git@github.com"
```
mvorisek: not needed once https://github.com/atk4/image/pull/43/files#diff-8aad5bc8efe55f15feea83ad1cb1febdd84a072ef653e323afa72d9810890429R32 is merged

src/Behat/Context.php
```diff
@@ -89,17 +120,12 @@ protected function assertNoException(): void
...
-    protected function disableDebounce(): void
-    {
-        $this->getSession()->executeScript('atk.options.set("debounceTimeout", 20)');
-    }
-
     /**
      * Sleep for a certain time in ms.
      *
-     * @Then I wait :arg1 ms
+     * @Then I sleep :arg1 ms
```
ibelar: keep Behat nomenclature using "Then I wait" - you can name the function as you want.

ibelar (changes requested):

js/src/atk-utils.js
```diff
@@ -69,4 +68,37 @@ const atkUtils = (function () {
     };
 }());
 
-export { atkOptions, atkEventBus, atkUtils };
+function atkDebounce(func, wait, options) {
+    let timerId;
```
ibelar: let timerId = 0 - setInterval return a Number.

js/src/atk-utils.js
```diff
@@ -69,4 +68,37 @@ const atkUtils = (function () {
...
+    let debouncedInner;
+
+    function createTimer() {
+        timerId = setInterval(() => {
+            if (!debouncedInner.pending()) {
+                clearInterval(timerId);
+                timerId = undefined;
```
ibelar: Set it back to 0

js/src/atk-utils.js
```diff
@@ -69,4 +68,37 @@ const atkUtils = (function () {
...
+                jQuery.active--;
+            }
+        }, 25);
+        jQuery.active++;
+    }
+
+    debouncedInner = lodashDebounce(func, wait, options);
+
+    function debounced(...args) {
+        if (timerId === undefined) {
```
ibelar: you can check for 0 value or if(!timerId)

js/src/atk-utils.js
mvorisek: `undefined` is used also by the `lodash` lib :)

src/Behat/Context.php
mvorisek: I sleep and do nothing, sleep is explicit term for that.

ibelar: most of the Behat implementation I have seen use wait for Behat. I like to keep it that way.

js/src/atk-utils.js
ibelar: Still - better use a Number type instead of 'undefined', always.

mvorisek: `null` maybe, but never a number as `0` may ba valid in theory 8-)

src/Behat/Context.php
mvorisek: We can drop it completely :D

js/src/atk-utils.js
ibelar: My understanding is that browser will always return a value higher than 0.

src/Behat/Context.php
ibelar: No - because it can be useful when debugging new test.

mvorisek: That is why I did not delete it :)

js/src/atk-utils.js
mvorisek: I changed it ot `null`.

ibelar (approved):
LGTM

PR #1525, opened 2020-10-29
Fix jQueryWait typo - do not wait 1 000 000 times longer

mvorisek:
this was the issue, why Behat stuck on debug sometimes :)

### DIFF
--- a/tests-behat/bootstrap/Context.php
+++ b/tests-behat/bootstrap/Context.php
@@ -67,11 +67,12 @@ class Context extends RawMinkContext implements BehatContext
             'animation-duration' => '5s',
             'transition-duration' => '5s',
         ]);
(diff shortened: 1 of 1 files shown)


tests-behat/bootstrap/Context.php
```diff
@@ -480,7 +478,7 @@ protected function jqueryWait($duration = 5000)
             }
         }
 
-        throw new \Exception('JQuery did not finished within a given time limit');
+        throw new \Exception('jQuery did not finished within a time limit');
```
georgehristov: Should be "jQuery did not finish within the time limit"

georgehristov (approved):

tests-behat/bootstrap/Context.php
mvorisek: yes, thanks!

PR #1533, opened 2020-11-09
[fix] Texarea issue with Form can leave

### LINKED ISSUE #1527: form->canLeave=false not working for text fields
In a form with a control that renders a type=text field (it's displayed as a textarea field), the canLeave=false function does not work. Changes in that textarea field are not noted, and thus the form can be left without notification. Hope that this will be fixed, otherwise canLeave can have the users leave with unsaved changes. @ibelar

ibelar:
fix: #1527


mvorisek (approved):

mvorisek:
fix js CS before merge

PR #1535, opened 2020-11-11
Reorganize Action Executor demo

ibelar:
related with: https://github.com/atk4/data/pull/783

- Create New demo section for Data Action executor;
- Reorganize demo files for action executor.
- Add basic documentation;

Screen Shot 2020-11-11 at 2 38 08 PM


romaninsh (approved):

mvorisek:
here are breaking undocumented changes - no label, not described in the PR description

`getCaption()` at least

mvorisek:
@ibelar this has also broken the UI - add button in CRUD is now labeled as "Add" instead of "Add Model name/class" as before

ibelar:
> @ibelar this has also broken the UI - add button in CRUD is now labeled as "Add" instead of "Add Model name/class" as before

To keep as before you need to update the UserAction caption.
This was to keep consistency between executors. It now uses action caption for Header content or Button label; and description for Subheader content or Modal title. 

But you are right, I should have put a mention for it.

mvorisek:
@ibelar see `/demos/collection/crud3.php` demo, the "Add" button is affected there as well - should be rechecked everywhere where description is still used

ibelar:
Not sure I see what you mean. Crud now use `$action-getCaption() ` for labelling button. Thus, showing 'Add' is normal.

mvorisek:
But previusoly is was `Add <class name>`

PR #1537, opened 2020-11-16
Make outputResponseJson() and outputResponseHtml() protected

PhilippGrashoff:
These methods were private before, making it unneccessarily hard for example to overwrite caughtException(). In my case, I want to return a jsToast if it was a JS request, not a full page load:
```
 //this code is in the if clause that checks if it was a JS request
$jsToast = new JsToast(['message' => $message, 'class' => 'error']);
           $this->outputResponseJson(
                [
                    'success' => true,
                    'message' => '',
                    'atkjs' => $jsToast->jsRender()
                ]
            );
```

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -1067,7 +1067,7 @@ class App
      *
      * @param string[] $headers
      */
-    private function outputResponseHtml(string $data, array $headers = []): void
+    protected function outputResponseHtml(string $data, array $headers = []): void
     {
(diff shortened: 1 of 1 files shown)


mvorisek (changes requested):
these methods are intended to be private as they may be removed once output is reworked

in atk - these are too low level methods - you should break execution using some callback and return the data from it with specific mime

PR #1538, opened 2020-11-16
[fix] Behat window size issue

ibelar:
rebased once https://github.com/atk4/ui/pull/1524 is merged

related with https://github.com/minkphp/MinkSelenium2Driver/pull/327 , finish once merged & released

original discussion:
- Use Context instead of DumpContext for CI
- Remove 'show_cmd' option since this is mainly used for debugging. 
see: https://github.com/Behat/MinkExtension/blob/master/doc/index.rst

Fix issue with selenium/standalone-chrome image when test are run using arguments: --headless and --window-size together and window-size value is set using low value. 
Most of the time, an element is not found or the element text value returns null, although the element contains the text.
see #1469 

Note: --window-size option value must make sense since error like 'element click intercept'  or 'element not interactable' will occur on certains tests.


behat.yml.dist
```diff
@@ -19,5 +18,4 @@ default:
               extra_capabilities:
                 chrome:
                   args:
-                    - '--headless'
```
mvorisek, line 21: Headless must be supported.

tests-behat/bootstrap/Context.php
```diff
@@ -93,13 +94,11 @@ protected function disableDebounce(): void
...
-     * Sleep for a certain time in ms.
+     * Wait for a certain time in ms.
      *
-     * @Then I sleep :arg1 ms
-     *
-     * @param $arg1
+     * @Then I wait :arg1 ms
      */
-    public function iSleep($arg1)
+    public function iWait($arg1)
```
mvorisek: We sleep with `sleep` command / fied time, not wait for something.

tests-behat/bootstrap/Context.php
```diff
@@ -275,8 +274,7 @@ public function iPressModalButton($arg)
      */
     public function modalIsOpenWithText($arg1)
     {
-        // get modal
-        $modal = $this->getSession()->getPage()->find('css', '.modal.transition.visible.active.front');
+        $modal = $this->waitForNodeElement('.modal.transition.visible.active.front');
```
mvorisek: for what - with jqueryWait, we wait for AJAX + some enough time in which all animations should finish.

tests-behat/bootstrap/Context.php
```diff
@@ -318,6 +305,40 @@ public function modalIsShowingText($arg1, $arg2)
         }
     }
 
+    /**
+     * Get a node element by it's selector.
+     * Will try to get element for 20ms.
+     * Exemple: Use with a modal window where reloaded content
+     * will resize it's window thus making it not accessible at first.
+     */
+    private function waitForNodeElement(string $selector): ?NodeElement
```
mvorisek: I do not see any need for this method, when we need to wait, we wait with jquery wait which wait some extra time.

mvorisek (changes requested):

behat.yml.dist
ibelar: NO - The image in use (selenium/standalone-chrome:latest) is already set for --headless mode. 
Beside when --headless mode is set as arguments, the the scopebuilder.feature is failing:

Screen Shot 2020-11-17 at 9 33 13 AM

Test pass otherwise.

tests-behat/bootstrap/Context.php
ibelar: Has a user I wait, not sleep and the BDD you should describe what the user does.
Plus look at the method name for mink `$this->getSession()->wait($arg1);` It say wait not sleep.

tests-behat/bootstrap/Context.php
ibelar: Because when content has finished loading inside the modal, the modal will resize according to new content.
If you try to get the modal window while it is reloading then the test fails because the modal return null.
You can look at the commit history to see.

behat.yml.dist
mvorisek: this is what we distribute, not only for GitHub

do you have any reference that `selenium/standalone-chrome:latest` is already set for --headless mode?

tests-behat/bootstrap/Context.php
ibelar: same as above.

tests-behat/bootstrap/Context.php
mvorisek: User sleeps. It does nothing. When user waits, he can do something - like browse Facebook.

Why is code `sleep` is not called `wait`?

tests-behat/bootstrap/Context.php
mvorisek: I agree.

Now how to handle it - jquery wait should do that FOR EVERYTHING :)

do you interact with Modal as the "1st operation in Behat step"

tests-behat/bootstrap/Context.php
mvorisek: same as above.

tests-behat/bootstrap/Context.php
ibelar: I do not agree with you on this one. User wait for something to finish. Like an animation or page loading.

mvorisek: Again:

- SLEEP - you freeze in time and wake up after some time
- WAIT - you do something in meanwhile, like checking if finish, browse Facebook ;-), etc.

And here - nothing is checked, we simply dummy wait, thus SLEEP.

Makes sense?

tests-behat/bootstrap/Context.php
ibelar: I do not see anything that will let us know that the modal window is resizing and has finish resizing... Do you?
That is why I put a loop there. The other solution will be to intentionally wait for a few ms but I think the loop is better.

behat.yml.dist
ibelar: No other reference other than the test is working without it.
For the test to pass using headless mode, window-size needs to be bigger.

tests-behat/bootstrap/Context.php
ibelar: Not from a user standpoint, it does not. BDD is about user behaviour...

mvorisek: Do you agree with SLEEP/WAIT definition, ie. if conditional or dummy? Then what the user does here?

tests-behat/bootstrap/Context.php
mvorisek: jquery wait waits some extra time for exactly these situations :)

ibelar: Well I guess it does not wait long enough... I can try to increase time there but it will increase for every step, no?

tests-behat/bootstrap/Context.php
ibelar: he is waiting...

mvorisek: for buzzer/delay?

tests-behat/bootstrap/Context.php
mvorisek: > do you interact with Modal as the "1st operation in Behat step"

if yes, then there is probably some other issue, as the additional delay is already quite long. And I tested it very extensively - there is probably other issue.

ibelar: There is an issue and this is how to fix it...

mvorisek:
@ibelar the issue is that `getText()` seems to not work for elements outside the current viewport. If https://github.com/atk4/ui/blob/bf95e1a/tests-behat/bootstrap/Context.php#L532 is replaced with `getHtml()`, it works even with the smaller window size (when whitespaces are normalized).

Investigating more to identify the issue.

getText getter:
https://github.com/minkphp/Mink/blob/v1.8.1/src/Element/Element.php#L174
https://github.com/minkphp/MinkSelenium2Driver/blob/v1.4.0/src/Selenium2Driver.php#L540

getHtml getter:
https://github.com/minkphp/Mink/blob/v1.8.1/src/Element/Element.php#L182
https://github.com/minkphp/MinkSelenium2Driver/blob/v1.4.0/src/Selenium2Driver.php#L554

mvorisek:
closing because https://github.com/minkphp/MinkSelenium2Driver/pull/327

might be interesing to replace the limited Mink impl. from our Context class

PR #1543, opened 2020-11-25
[fix] check if buttonSave set

georgehristov:


mvorisek (approved):
write here usecase/steps to reproduce, then LGTM

DarkSide666 (approved):
It could be that we have removed buttonSave button at some step earlier, so this checking is only good thing to have.

PR #1544, opened 2020-11-25
Feature/Add lookup component to ScopeBuilder

ibelar:
Reference field can use Lookup component within scope builder.

Screen Shot 2020-11-25 at 4 26 09 PM


demos/_unit-test/scope-builder.php
```diff
@@ -15,14 +19,15 @@
 $model = new Stat($app->db, ['caption' => 'Demo Stat']);
 
 $project = new Condition('project_name', Condition::OPERATOR_REGEXP, '[a-zA-Z]');
-$brazil = new Condition('client_country_iso', '=', 'Brazil');
+$brazil = new Condition('client_country_iso', '=', '30');
```
mvorisek: use ISO code (BR) lookup preferably

src/Form/Control/ScopeBuilder.php
```diff
@@ -316,11 +343,40 @@ public function setModel(Model $model)
...
+     */
+    public function outputApiResponse()
+    {
+        $fieldName = $_GET['atk_vlookup_field'] ?? null;
+        $query = $_GET['atk_vlookup_q'] ?? null;
+        $data = [];
+        if ($fieldName) {
+            $model = $this->model->getField($fieldName)->reference->refModel();
+            if (!empty($query)) {
+                $model->addCondition($model->title_field, 'like', '%' . $query . '%');
```
mvorisek, line 368: this is quite fragile query and it needs much more code to be really universal, like to work correctly across collations - can we instance a standard Lookup here are reuse code for data query from there?

tests-behat/bootstrap/Context.php
```diff
@@ -542,6 +569,23 @@ public function compareInputValueToElementText($inputName, $selector)
...
 
+    /**
+     * Find a dropdown component within an html element
+     * and check if value is set in dropdown.
+     */
+    private function assertDropdownValue(NodeElement $element, string $value, string $selector)
+    {
+        $dropdown = $element->find('css', $selector);
+        if (!$dropdown) {
+            throw new \Exception('Dropdown input not found using selector: ' . $selector);
```
mvorisek: Even I normally do not like to import non namespaced classes, I would recommend to import/add `use Exception` to stress that we want to throw this type of exceptions in Behat Context.

demos/_unit-test/scope-builder.php
DarkSide666: @mvorisek is right - iso field should contain iso code not id.

demos/_unit-test/scope-builder.php
```diff
@@ -72,23 +77,30 @@
...
               },
               {
                 "type": "query-builder-rule",
                 "query": {
                   "rule": "client_country_iso",
                   "operator": "equals",
-                  "value": "Brazil"
+                  "value": "30",
+                  "option": {
+                    "key": "30",
```
DarkSide666: What's the difference between key and value? And i think it would be better to use name `label` instead of `text` if that's what it means. Why this `option` block is needed at all?

src/Form/Control/ScopeBuilder.php
```diff
@@ -146,6 +157,13 @@ class ScopeBuilder extends Control
         self::OPERATOR_NOT_EMPTY,
     ];
 
+    protected const ENUM_OPERATORS = [
+        self::OPERATOR_EQUALS,
+        self::OPERATOR_DOESNOT_EQUAL,
```
DarkSide666, line 162: I think `OPERATOR_DOESNOT_EQUAL` and `OPERATOR_NOT_EMPTY` should be in same "formatting".
I mean both `OPERATOR_NOT_something` like `OPERATOR_NOT_EQUAL` even if that's maybe against spelling rules (not sure)

src/Form/Control/ScopeBuilder.php
```diff
@@ -200,6 +218,10 @@ class ScopeBuilder extends Control
             self::OPERATOR_EQUALS => Condition::OPERATOR_EQUALS,
             self::OPERATOR_DOESNOT_EQUAL => Condition::OPERATOR_DOESNOT_EQUAL,
         ],
+        'lookup' => [
+            self::OPERATOR_EQUALS => Condition::OPERATOR_EQUALS,
+            self::OPERATOR_DOESNOT_EQUAL => Condition::OPERATOR_DOESNOT_EQUAL,
```
DarkSide666, line 223: ah I see - it's like that in Condition class too. Then probably it's correct to have it like this here too.

src/Form/Control/ScopeBuilder.php
```diff
@@ -674,9 +762,34 @@ public static function conditionToQuery(Scope\Condition $condition, $inputsMap =
...
+    /**
+     * return extra value option associate with certain inputType or null otherwise.
+     */
+    protected static function getOption(string $type, string $value, Condition $condition): ?array
+    {
+        $option = null;
+        switch ($type) {
+            case 'lookup':
+                $model = $condition->getModel()->getField($condition->key)->reference->refModel();
+                $rec = $model->tryLoad($value);
```
DarkSide666: probably here `tryLoadBy($condition->getModel()->getField($condition->key)->reference->their_field, $value)`

DarkSide666 (changes requested):

demos/_unit-test/scope-builder.php
ibelar: This was per the original implementation from @georgehristov - I can check how/why it does not store ISO instead.
@georgehristov - any input on this one. 
Besides there was also a bug before as it was using the name instead of id value when using $field->reference

src/Form/Control/ScopeBuilder.php
ibelar: not sure what you mean here. This is basically how all other components are doing. Can you explain a bit more?

demos/_unit-test/scope-builder.php
ibelar: Option block is use to format atkLookup component correctly when a value is set. In other word, the atk-lookup component required the option key, text and value to be properly set and display. 

As of now, only atk-lookup required this, but in future, it could be a range slider component that would require min or max value for example. These would be pass via the option property.

demos/_unit-test/scope-builder.php
DarkSide666: Yes using name there also looks wrong. It should be value of that field so in this case iso value. If you lookup by client_country_id, then country_id value etc.

src/Form/Control/ScopeBuilder.php
DarkSide666: I think he means that it should reuse something we already have implemented in Lookup field. But I guess it was not in lookup field, but in QuickSearch.

demos/_unit-test/scope-builder.php
DarkSide666: Can't these be pulled from respective data model not saved in json? For example, what if we rename Brazil to Argentina in some other form? Will this json will still be valid then?

ibelar: It will still be working because the option is generated on load. Also, option is irrelevant when doing query to scope.

demos/_unit-test/scope-builder.php
ibelar: done. Now use their_field to properly create dropdowns list key-value pair. Therefore will use 'BR' in query value.

src/Form/Control/ScopeBuilder.php
ibelar: done! now use $reference->getTheirField() to properly set option value for dropdown.

tests-behat/bootstrap/Context.php
ibelar: done!

demos/_unit-test/scope-builder.php
DarkSide666: Option was needed for vue builder. It has no meaning (to store it) in our case, but removing of it from vue requires a lot of changes in vue class, so we can just ignore it on our side. It's reloaded on widget load with proper values anyway.

demos/_unit-test/scope-builder.php
```diff
@@ -43,11 +48,10 @@
...
 
 $expectedWord = <<<'EOF'
-     Project Budget is greater or equal to '1000' 
-     and (Project Name is regular expression '[a-zA-Z]' 
-            and Client Country Iso is equal to 'Brazil' 
-            and Start Date is equal to '2020-10-22') 
-    and (Finish Time is not equal to '22:22' or Is Commercial is equal to '0')
+    Project Budget is greater or equal to '1000' 
+    and (Project Name is regular expression '[a-zA-Z]' 
+    and Client Country Iso is equal to 'Brazil' and Start Date is equal to '2020-10-22')
```
DarkSide666, line 53: Should be "Client Country Iso is equal to 'BR'" here isn't it?

DarkSide666 (approved):
LGTM

src/Form/Control/ScopeBuilder.php
ibelar: both, Lookup and QuickSearch use` $model->addCondition($model->title_field, 'like', '%' . $query . '%');`

mvorisek: ok with me for now

demos/_unit-test/scope-builder.php
ibelar: Change should be done in Model/Scope::toWords() method in atk4/Data

PR #1545, opened 2020-11-26
Remove manual waits from Behat testings

mvorisek:
fix unresolved issues from #1469


ibelar:
@mvorisek - can you explain what are these major issues?

mvorisek:
replaced with https://github.com/atk4/ui/pull/1524

PR #1546, opened 2020-11-26
Partial revert "Reorganize Action Executor demo"

mvorisek:
to be fixed by @ibelar , there are major unresolved issues in #1535


mvorisek:
breaking a lot, related with data PR #783 which defines caption as "text for button", but description still then uses caption as a fallback

as UA reorganize has been in the same PR, so reverted without reorganize

ibelar:
This was identified as BC Break. 

As state before:
> This was to keep consistency between executors. It now uses action caption for Header content or Button label; and description for Subheader content or Modal title.

Is there any other issues? please advice.

mvorisek:
I isolated this and the another PR as you merged it too soon as discussed.

I literaly do not know what ALL is broken :(

I noticed the captions which are not only BC breaking in sense of code, but also in sense of result UI (like buttons without names).

ibelar:
> I noticed the captions which are not only BC breaking in sense of code, but also in sense of result UI (like buttons without names).

Can you share some examples? how to reproduce?

mvorisek:
before

image



after

image

ibelar:
@mvorisek - Yes, exactly what this BC Break is all about. The crud menu button now uses the UserAction caption for label instead of action description like before. 
You can change button label by overiding default caption:
`$model-getUserAction('add')->caption = 'Add Invoice';`

PR #1547, opened 2020-11-28
[fix] - Crud menu icon

ibelar:
In the Crud menu, passing an UserAction action would automatically add a plus icon to the menu button.
This fix will only add a plus icon when action modifier === CREATE.


mvorisek:
makes sense

DarkSide666 (approved):
LGTM

PR #1550, opened 2020-11-30
Feature/Executor refactor

### LINKED ISSUE #1548: Remove Data\UserAction::ui property dependency in UI
UI class like Crud, Grid and any UserAction Executor should not be dependant on Data\Model\UserAction->ui property.

Take this data model user action as an example:

```
 $model->addUserAction('edit', [
    'fields' => true,
    'modifier' => Model\UserAction::MODIFIER_UPDATE,
    'appliesTo' => Model\UserAction::APPLIES_TO_SINGLE_RECORD,
    'callback' => 'save',
    'ui' => ['icon' => 'edit', 'button' => [null, 'icon' => [\atk4\ui\Icon::class, 'edit']], 'execButton' => [\atk4\ui\Button::class, 'Save', 'blue']], 
]);
```

Creating Crud, Grid or using Ui UserAction executor should not rely upon the 'ui' property of Data\Model\UserAction. In fact, I believe this property should be removed.

- Ui element should rely on other model user action properties in order to provide basic information to UI View class on what to display to the user and how to react after the user execute the model action.
- Ui element that uses model user action should offer a way to override Views, like buttons, that fire them.
- Executor should also offer a way of overriding View (ModelExecutor button) that executes the action.

ibelar:
Remove all dependency to `atk4\Data\UserAction::ui` property

Fixes #1548 

Use of an ExecutorFactory class to properly wired UI element from its model action.

- Factory method to create action executor based on action definition;
- Enable registration of specific Ui executor for certain model UserAction
`  $app->getExecutorFactory()->registerExecutor($country->getUserAction('confirm'), [ConfirmationExecutor::class]);
`

- Allow overriding UI element responsible to trigger the action for a specific component like Crud or Grid
```
$app->getExecutorFactory()->registerTrigger(
    $app->getExecutorFactory()::TABLE_BUTTON,
    [Button::class, 'ui' => 'atk-test button', 'icon' => 'pencil'],
    $m->getUserAction('edit')
);
```
- Allow overriding action caption per action or per model/action
`$app->getExecutorFactory()->registerCaption($model->getUserAction('add'), 'Add new record');
`

- Add new method for Grid to setup button using column action.
```
// Creating a button for executing model test user action.
$grid->addExecutorButton($grid->getExecutorFactory()->create($model->getUserAction('test'), $grid));
```

### Change look and buttons default label by overriding ExcutorFactory

````
// Overriding basic ExecutorFactory in order to change Table and Modal button.
// and also changing default add action label.
class MyFactory extends ExecutorFactory {
    public const BUTTON_PRIMARY_COLOR = 'green';

    protected $triggerSeed = [
        self::TABLE_BUTTON => [
            'edit' => [Button::class, null, 'icon' => 'pencil'],
            'delete' => [Button::class, null, 'icon' => 'times red'],
        ],
        self::CARD_BUTTON => [
            'edit' => [Button::class, 'Edit', 'icon' => 'pencil', 'ui' => 'tiny button'],
            'delete' => [Button::class, 'Remove', 'icon' => 'times', 'ui' => 'tiny button'],
        ],
    ];

    protected $triggerCaption = [
        'add' => 'Add New Record',
    ];
}
````
Screen Shot 2021-01-26 at 10 56 29 AM

### ExecutorFactory can be applied to a specific View object instance:

````
// Overriding basic ExecutorFactory in order to change Card button.
class SpecialCardFactory extends ExecutorFactory {
    public const BUTTON_PRIMARY_COLOR = 'green';

    protected $actionIcon = [
        'callback' => 'sync',
        'preview' => 'eye',
        'edit_argument' => 'user edit',
        'edit_argument_prev' => 'pen square',
        'edit_iso' => 'pencil',
        'confirm' => 'check circle',
        'multi_step' => 'window maximize outline',
    ];

    public function __construct()
    {
        // registering card button default with our own method handler.
        $this->triggerSeed = array_merge(
            $this->triggerSeed,
            [self::CARD_BUTTON => ['default' => [$this, 'getCardButton']]]
        );
    }

    protected function getCardButton($action, $type)
    {
        return [Button::class, 'icon' => $this->actionIcon[$action->short_name]];
    }
}

$card = Card::addTo($app, ['useLabel' => true, 'executorFactory' => new SpecialCardFactory()]);
$card->setModel($country);
````

Screen Shot 2021-01-26 at 11 10 24 AM

## BC Break

Setting Executor via the $action-ui property might break your code since because the executor uses for the action will be revert to the default provide by executor factory. If you need to use a specific executor class for your action, you will need to register one: 

`$app->getExecutorFactory()->registerExecutor($file->getUserAction('edit'), [$myExecutorClass]);
`

If you were relying heavily on Model\UserAction::$ui property to set up your action button then this PR may break the look of these buttons.
You will need to register your user action ui property into an action trigger in ExecutorFactory to restore them.

Previous action definition using ui property:
```
$country->addUserAction('mail', [
    'appliesTo' => \atk4\data\Model\UserAction::APPLIES_TO_SINGLE_RECORD,
    'preview' => function ($model) { return 'here is email preview for ' . $model->get('name'); },
    'callback' => function ($model) { return 'email sent to ' . $model->get('name'); },
    'description' => 'Email testing',
    'ui' => ['icon' => 'mail', 'button' => [null, 'icon' => 'green mail']],
 ]);
```

From this pr - you will need to remove ui property in your action definition and register the button in ExecutorFactory:
New action definition

```
$country->addUserAction('mail', [
    'appliesTo' => \atk4\data\Model\UserAction::APPLIES_TO_SINGLE_RECORD,
    'preview' => function ($model) { return 'here is email preview for ' . $model->get('name'); },
    'callback' => function ($model) { return 'email sent to ' . $model->get('name'); },
    'description' => 'Email testing',
]);

// Register a trigger for mail action in Crud
$app->getExecutorFactory()->registerTrigger(
    $app->getExecutorFactory()::TABLE_BUTTON,
    [Button::class, null, 'icon' => 'green mail'],
    $country->getUserAction('mail')
);
```
### Important

Please note that your code should not break if you where using ui property for buttons - only the look of the button since they will be revert to ExecutorFactory default.

### CODE REVIEW

**Comment by mvorisek** [comment]
> Remove all dependency to `atk4\Data\UserAction::ui` property

more formally remove ui property from atk4/data, right? (as atk4/data does not depend on atk4/ui)

**Comment by ibelar** [comment]
> 
> more formally remove ui property from atk4/data, right? (as atk4/data does not depend on atk4/ui)

Yes but this pr is for ui property in Model\UserAction only, not Field.

File: src/UserAction/ExecutorFactory.php [hunk 1]
```diff
@@ -67,15 +67,15 @@ class ExecutorFactory
      */
     public static function registerActionExecutor(UserAction $action, array $seed)
     {
-        self::$actionExecutorSeed[self::getModelId($action)][$action->short_name] = $seed;
+        static::$actionExecutorSeed[static::getModelId($action)][$action->short_name] = $seed;
```
mvorisek: I see a lot of static calls, can we make it singleton at least?

src/View.php
```diff
@@ -265,6 +265,10 @@ protected function init(): void
         $this->_add_later = [];
         parent::init();
 
+        if (!$this->executorFactory) {
+            $this->executorFactory = $this->getApp()->defaultExecutorFactory;
```
mvorisek: I would highly prefer an instance here, so we do not need static at all

And - probably drop `executorFactory` in view, we do not want that for almost all render tree item.

If user should have option to customize this factory on per View instance basis - why? - then add getter with default to app here which can be overriden.

ibelar: > If user should have option to customize this factory on per View instance basis - why? - then add getter with default to app here which can be overriden.

Why not? Let says you have one specific $card using a lot of model action, you can then easily set this card only with buttons using icon only, while you would still need full button label on other $card view.

mvorisek: if needed in multiple extended View classes, add with trait, View is base building block and I do not belive `View` class itself benefits by having executor factory in any way

mvorisek: then annotate the property with `ExecutorFactory` type only (no string support)

ibelar: > if needed in multiple extended View classes, add with trait, View is base building block and I do not belive `View` class itself benefits by having executor factory in any way

Trait will not do any good, since executorFactory is needed within View::on() method.

mvorisek: Let's take two simple UI elements:

- bare `View`
- a `Button`

how does having a property with executor factory helps on these elements?

ibelar: $btn-on('click', $model-getUserAction('test));

ibelar: You can still pass an action to on method. But since the action needs an executor, then we need ExecutorFactory

mvorisek: Well, ok, let's have that property for now but do not assing by default. Always fallback to app - add a protected getter method for it like `return prop ?? app->...`.

ibelar: already done! :)

src/View.php
```diff
@@ -120,6 +121,9 @@ class View extends AbstractView implements JsExpressionable
      */
     public $element;
 
+    /** @var ExecutorFactory Executor factory. */
```
mvorisek:
```suggestion
    /** @var ExecutorFactory */
```

src/View.php
mvorisek: nice! :)

src/View.php
```diff
@@ -121,7 +121,7 @@ class View extends AbstractView implements JsExpressionable
      */
     public $element;
 
-    /** @var ExecutorFactory */
+    /** @var string Seed class name */
```
mvorisek: No! We do not want static factory, we want factory instance.

ibelar: I will try to get there. hold on ;-)
This is the plan,  Let me know if this is ok:
- View::getExecutorFactory() will return a Singleton instance;
   - Either from View::$excutorProperty class name or using app default

I still need to add a few things in the factory the I will try to get this working.

mvorisek: I think this way:

- app instance executor factory
- View `excutorProperty` property is either null - factory from app is used, or an instance, which is used directly

phpstan.neon.dist
```diff
@@ -304,7 +304,7 @@ parameters:
         - '~^Call to an undefined method atk4\\ui\\JsChain\:\:emit\(\)\.$~'
         - '~^Call to an undefined method atk4\\ui\\JsChain\:\:clearData\(\)\.$~'
         - '~^Call to an undefined method atk4\\ui\\JsChain\:\:addJsonData\(\)\.$~'
-        - '~^Access to an undefined property atk4\\ui\\UserAction\\JsExecutorInterface&atk4\\ui\\View\:\:\$viewForUrl\.$~'
+        # - '~^Access to an undefined property atk4\\ui\\UserAction\\JsExecutorInterface&atk4\\ui\\View\:\:\$viewForUrl\.$~'
```
mvorisek: completely delete it (exclude rules in "generated" section), it is actually the state we want to achive

mvorisek:
```suggestion
```

ibelar: I was testing to see if this was the case.

mvorisek: 👍

ibelar:
@mvorisek - I will start working on refactoring ExecutorFactory to use Instance instead of static.
What do you think about method naming? Should we drop the word 'Action' in each method name?
Ex: createActionTrigger to become createTrigger etc...

mvorisek:
> @mvorisek - I will start working on refactoring ExecutorFactory to use Instance instead of static.
What do you think about method naming? Should we drop the word 'Action' in each method name?
Ex: createActionTrigger to become createTrigger etc...

In factory for "action" only I am ok with names /wo "action"..

tests/ExecutorFactoryTest.php
```diff
@@ -0,0 +1,115 @@
...
+
+        // register an executor for a specific type.
+        $factory->registerTypeExecutor('MY_TYPE', [BasicExecutor::class]);
+        $myRequiredExecutor = $factory->create($this->m->getUserAction('confirm'), $view, 'MY_TYPE');
+
+        // register executor for a specific action
+        $factory->registerExecutor($this->m->getUserAction('basic'), [BasicExecutor::class]);
+        $myBasicExecutor = $factory->create($this->m->getUserAction('basic'), $view);
+
+        $this->assertInstanceOf(ModalExecutor::class, $modalExecutor, 'Not ModalExecutor Type');
```
mvorisek, line 77: absolutely unneeded comments...

tests/ExecutorFactoryTest.php
```diff
@@ -0,0 +1,115 @@
...
+        ]);
+
+        $this->addUserAction('basic', []);
+    }
+}
+
+class ExecutorFactoryTest extends AtkPhpunit\TestCase
+{
+    /** @var Model */
+    public $m;
```
mvorisek: better to name `$model`

mvorisek:
@ibelar can we merge it?

ibelar:
@mvorisek - This is ready. Waiting for @romaninsh or @DarkSide666 to approve the changes.

DarkSide666 (approved):
I can't go in details of this now and not so good in frontend things.
I will approve just in case you need that, but does this mean we can't define UI executor in atk4/data action ui property anymore?

mvorisek:
> 
> 
> I can't go in details of this now and not so good in frontend things.
> I will approve just in case you need that, but does this mean we can't define UI executor in atk4/data action ui property anymore?

You can define executor factory in App, see https://github.com/atk4/ui/pull/1550/files#diff-ad46c7350003ccf7f086dccbecc6fcb7706a1558b14eb5e437a856b8c619e71fR227, thus I would say you can, as long as you will provided factory that does that :)

Issue #1462, opened 2020-09-29
DropDown and Lookup in connection with hasOne is not working in CRUD anymore


karakal:
In the newest code in develop (which I need to use...) DropDown and Lookup in a model will not be filled anymore.

I have multiple models with hasOne and in the table everything is fine, but when I try to edit an element, the dropdown or lookup will always return "No results found".

This error is even in the demo visible at https://ui.agiletoolkit.org/demos/collection/actions.php when you try to edit an element, that has a parent-folder set:
![grafik](https://user-images.githubusercontent.com/2448293/94605527-8d5f0a80-0299-11eb-926a-3abbda1bbede.png)

When you set the model explicit to the dropdown everything is working fine.

ibelar:
@karakal - Can you share code on how you defined your model? All seems to be working fine. I think the demo is fine too because hasOne only returns one parent only. @DarkSide666 - can you confirm about file demo data?

karakal:
@ibelar 
All of my hasOne are defined as follows:
`$this->hasOne('Unit', [
            'model' => '<path>\Unit',
            'our_field' => 'unit_id',
            'their_field' => 'id',
            'caption' => 'Unit',
            'required' => true
        ])->withTitle();`

This works out as correctly displayed in the overview:
![grafik](https://user-images.githubusercontent.com/2448293/95204484-73b34b00-07e4-11eb-84f4-90297101cfea.png)

and it will display correctly the selected entry:
![grafik](https://user-images.githubusercontent.com/2448293/95204535-87f74800-07e4-11eb-973c-ab274204161f.png)

But it displays "No result found":
![grafik](https://user-images.githubusercontent.com/2448293/95204596-99d8eb00-07e4-11eb-86a1-f4457e2f4079.png)

although there should be a selection of them:
![grafik](https://user-images.githubusercontent.com/2448293/95204651-ad845180-07e4-11eb-8f5e-139aa0c57aaa.png)

And I don't think the demo is right, because there should be a selection of a lot of root folders...

mkrecek234:
Hi @karakal I had similar strange effects with dropdowns not working - just because the installed PHP version was too old. Upgrade to 7.3 or 7.4 and some mysterious effects might be gone immediately.

DarkSide666:
There surely is some issue (or issues) here with hasOne and dropdowns.
Also model addTitle() added expression field don't load values. Maybe that's because of that.

Will work more on that tomorrow https://github.com/atk4/ui/pull/1477

[...]

PR #1556, opened 2020-12-03
Add lookup addCondition issue to demo

mhuser:
NOT TO BE MERGED

In atk4/ui 2.2.x it is possible to `addCondition()` to a form control `lookup->model`.
In atk4/develop it is still possible to `addCondition()` at initialization of the lookup, but not after. This PR add a segment to `lookup-dep` demo to show both at init or after addCondition behaviour on lookup.


mvorisek:
@georgehristov do we duplicate somewhere the model without cloning it, so the conditions are unset? That is what I think is the issue.

mhuser:
@mvorisek , @georgehristov seems adding this to `renderView()` fix the behaviour:
>         $this->callback = \atk4\ui\Callback::addTo($this);
>         $this->callback->set([$this, 'outputApiResponse']);
Not sure what it is doing.
These lines have been moved to `init()`

mvorisek:
@ibelar I think you want the one who moved this part, please check

ibelar:
> @ibelar I think you want the one who moved this part, please check

@mvorisek - Yes - It was moved to fix #1462

mvorisek:
> 
> 
> > @ibelar I think you want the one who moved this part, please check
> 
> @mvorisek - Yes - It was moved to fix #1462

but for some reasons no PR is linked there...

ibelar:
Close because replace by #1558

PR #1557, opened 2020-12-04
Requiring atk4/data is enough

mvorisek:
json ext is always available since PHP 8.0


DarkSide666 (approved):
Change as such is OK, but tests fail because new PHPUnit 9.5 doesn't have `getAnnotations()` method anymore.

DarkSide666 (approved):

PR #1558, opened 2020-12-04
[fix] Lookup add condition after model is set

ibelar:
replacement for PR #1556


mvorisek:
I think our code should not lead to these fixes, but ok for now

DarkSide666 (approved):
LGTM

mhuser:
Great, thank you so much !

PR #1559, opened 2020-12-04
Convert hardcoded strings in demos to hintable fields

mvorisek:
## Preface
The hintable support is another major game changer in sense of cleaness and maintainability. For now, adding it to demos only and it is up too every user to use it or not.

## Description
Annotate model fields and used them magically as a real properties.

## Usage
1. add Model class phpdoc
2. use the properties either as standard value property or supermagically for field names itself (see below)

#### standard ("value") property
`$model->is_folder` === `$model->get('is_folder')`
`$model->is_folder = false` === `$model->set('is_folder', false)`
- shorter to type
- refactorable
- typed for static analysis

#### "name" property
`$model->fieldName()->is_folder` === `'is_folder'`
- refactorable name


demos/collection/multitable.php
```diff
@@ -74,7 +74,7 @@ public function setModel(\Atk4\Data\Model $model, $route = [])
 
 $model = new File($app->db);
 $model->addCondition('parent_folder_id', null);
-$model->setOrder(['is_folder' => 'desc', 'name']);
+$model->setOrder([$model->fieldName()->is_folder => 'desc', $model->fieldName()->name]);
```
mvorisek, line 77: `$model->fieldName()->is_folder` returns `is_folder` string, but still refactorable like a value property

demos/collection/card-deck.php
```diff
@@ -14,13 +14,13 @@
...
-$countries->addCalculatedField('Cost', function ($model) {
+$countries->addCalculatedField('Cost', function (Country $country) {
     return '$ ' . number_format(random_int(500, 1500));
 });
 
 $action = $countries->addUserAction('book', [
-    'callback' => function ($model, $email, $city) {
-        return 'Your request to visit ' . ucwords($city) . ' in ' . $model->get('name') . ' was sent to: ' . $email;
+    'callback' => function (Country $country, $email, $city) {
+        return 'Your request to visit ' . ucwords($city) . ' in ' . $country->name . ' was sent to: ' . $email;
```
PhilippGrashoff, line 23: Shouldn't that be `$country->fieldName()->name`?

mvorisek: no ;-) - we want the value here, not name

mvorisek:
@DarkSide666 @georgehristov @ibelar please review, very powerfull lib I released for atk4. For now, I am adding it as a dev dependency ie. for demos only

I will then adjust everything in demos to make it analyseable/typehintable/refactorable.

mvorisek:
@DarkSide666 @georgehristov @ibelar please review, very powerfull lib I released for atk4. For now, I am adding it as a dev dependency ie. for demos only

I will then adjust everything in demos to make it analyseable/typehintable/refactorable.

romaninsh:
As pointed by others: 
Screenshot 2020-12-09 at 10 06 42

Checking if we'll need to bundle it into our ZIP distribution....

**Edit:** looks like it is unspecified.

romaninsh:
@mvorisek - I think this is fine but would be nice to have a bit more info in the readme file there just to understand what is that library doing and why is it needed:

https://github.com/mvorisek/atk4-hintable-mirror

**Edit:** found notes here, reading: https://github.com/mvorisek/atk4-hintable-mirror/blob/develop/src/Data/HintableModel.php#L14

romaninsh:
@mvorisek - why not bring magic properties natively into ATK Model ?

georgehristov:
@mvorisek I would also suggest adding native trait `\Atk4\Data\Model\HintableTrait`

mvorisek:
> I think this is fine but would be nice to have a bit more info in the readme file there just to understand what is that library doing and why is it needed:

> **Edit:** found notes here, reading: https://github.com/mvorisek/atk4-hintable-mirror/blob/develop/src/Data/HintableModel.php#L14

updated PR title
 
> @mvorisek - why not bring magic properties natively into ATK Model ?

I need more control over this than waiting days until someone merge something. This is why I also removed the PRs for it a few months ago.

> @mvorisek I would also suggest adding native trait `\Atk4\Data\Model\HintableTrait`

There are two issues with trait:
- `HintableModel` can not be then specified as a type
  - I checked now, very minor issue as super well written and specifying this type does not bring any benefict for end user ;-)
- `Model::init` method must be overrided
  - trait can NOT enforce it, see https://3v4l.org/MX9pm
  - user can alias a init method from trait or we can offer aliased method from trait directly
  - we can check if init() method from trait was called or not later, thus this should not be issue as well

So I can change the functionality to trait and offer it also under `\Atk4\Data\Model\HintableTrait` name.

If welcomed, I can add it also as atk4/data dep (so it does not have to be required in any other project) and possibly annotate `Model` id field in `Model` class phpdoc directly.

DarkSide666 (approved):
LGTM

georgehristov:
@mvorisek as far as I can see the hintable fields functionality covers only 'get'. It would be great to expand this to cover also 'set' for consistency and being intuitive.
```
$model->is_folder = false; // equivalent to
$model->set('is_folder', false);
```

mvorisek:
@georgehristov this of course supported already, easy one ;-)

https://github.com/mvorisek/atk4-hintable-mirror/blob/1.1.4/src/Data/HintableModelTrait.php#L179

demos/form-control/dropdown-plus.php
```diff
@@ -21,7 +21,7 @@
 
 $form->addControl('category_id', [Form\Control\Dropdown::class, 'model' => new Category($app->db)]);
 $form->addControl('sub_category_id', [Form\Control\DropdownCascade::class, 'cascadeFrom' => 'category_id', 'reference' => Category::hinting()->fieldName()->SubCategories]);
-// DEBUG only, uncomment before merge!!! $form->addControl('product_id', [Form\Control\DropdownCascade::class, 'cascadeFrom' => 'sub_category_id', 'reference' => SubCategory::hinting()->fieldName()->Products]);
+$form->addControl('product_id', [Form\Control\DropdownCascade::class, 'cascadeFrom' => 'sub_category_id', 'reference' => SubCategory::hinting()->fieldName()->Products]);
```
mvorisek, line 24: when this line is commented, tests pass

@ibelar, please help

mvorisek: when this line is commented, tests pass

@georgehristov can you please check why?

mvorisek, line 24: when this line is commented, tests pass

@georgehristov can you please help with fixing this?

ibelar:
Is there a specific reason to prefix table field name using 'xxx' or 'yyy'? Why not using the table name instead like 'country_name', 'stats_name' etd?

mvorisek:
> 
> 
> Is there a specific reason to prefix table field name using 'xxx' or 'yyy'? Why not using the table name instead like 'country_name', 'stats_name' etd?

The reason I added is solely to identify direct usage of original/actual DB field names.

~I have choosen `xxx_` to be able to easily find it. I would merge and revert later once everything is replaced (I replaced only for classes in `init-db.php`).~

Beatifulness addressed in https://github.com/atk4/data/pull/831

demos/form-control/dropdown-plus.php
georgehristov: The issue is caused by automatic creation of `their_field` value:

https://github.com/atk4/data/blob/07803e48828a2b6c12b9c05cb369344e5fd929ca/src/Reference/HasMany.php#L58

In this case it defaults to `product_category_id` and used in scope condition but field is not found due to your renaming.

mvorisek: Thanks, I created a PR for it https://github.com/atk4/data/pull/830, wdyt?

mvorisek: Thanks, I created a PR for it https://github.com/atk4/data/pull/830 . We should not guess any references at runtime, if absolutely needed, then lately on __construct

tests/DemosTest.php
```diff
@@ -234,7 +234,7 @@ protected function getPathWithAppVars(string $path): string
 
     public function demoFilesProvider(): array
     {
-        $excludeDirs = ['_demo-data', '_includes', '_unit-test', 'special'];
+        $excludeDirs = ['_demo-data', '_includes'];
```
mvorisek, line 237: @ibelar FYI do not add anything else here, this is why https://github.com/atk4/ui/pull/1586 was not caught by tests before

src/Component/ItemSearch.php
```diff
@@ -84,7 +84,7 @@ public function getQuery()
     public function setModelCondition(Model $model): Model
     {
         if ($q = $this->getQuery()) {
-            $model->addCondition('name', 'like', '%' . $q . '%');
+            $model->addCondition($model->title_field, 'like', '%' . $q . '%');
```
mvorisek, line 87: was a bug

PR #1565, opened 2020-12-07
[fix] Property access for Modal Executor

ibelar:
Allow private property access.
Useful when a special form is needed in Modal Executor.
 - When model action uses Arguments you can safely return using $this->getStep() when the customs form is for field step only;
- Get previous step data access using getActionData($step)


PhilippGrashoff:
Talking about private, wdyt about https://github.com/atk4/ui/pull/1537? I think putting a lot of private in our code makes it less extendable and customizable. IMO private should only be used over protected when there is a very good reason for it.

src/UserAction/ModalExecutor.php
```diff
@@ -397,15 +397,25 @@ protected function doPreview(View $modal)
...
 
         $this->_jsSequencer($modal, $this->jsGetExecute($return, $this->action->getOwner()->getId()));
     }
 
+    protected function getActionData(string $step): array
+    {
+        return $this->actionData[$step] ?? [];
+    }
+
+    protected function getStep(): string
```
mvorisek, line 414: not sure if needed now...

ibelar: When overriding addFormTo - you need to know in which step you are in order to see if you are dealing with 'args' or 'fields'

mvorisek: prop can be made protected... But I am ok with both... No more questions for this once.

DarkSide666 (approved):
LGTM

PR #1566, opened 2020-12-07
[fix] ReadMe link

### LINKED ISSUE #1564: README links are broken
links to component demos are broken, see https://github.com/atk4/ui#bundled-and-planned-components

ibelar:
Fix link in ReadMe

fix: #1564


mvorisek (approved):
I would still prefer the table with valid links, but I have nothing against merging as is.

PR #1567, opened 2020-12-08
Add more examples to grid.php demo + issue changing id:1 record

mkrecek234:
I am adding further action buttons as a good example how to use a virtual page or addActionButton to attach customised forms to a grid.
There is a strange error that occurs: If you click the "envelope", the selected record's name will be lower cased - BUT the first record with id:1 will not. This seems to be a bug.

By the way, any more elegant way to close the JsModal form onSubmit other than the workaround in my code?

### DIFF
--- a/demos/collection/grid.php
+++ b/demos/collection/grid.php
@@ -34,6 +34,51 @@ $grid->menu->addItem(['Delete All', 'icon' => 'trash', 'red active']);
 
 $grid->addColumn(null, [\Atk4\Ui\Table\Column\Template::class, 'hello<b>world</b>']);
 
+$modalviewvp = \Atk4\Ui\VirtualPage::addTo($grid->getApp()); // Virtual Page with customized content for Grid button
+$modalviewvp->set(function ($frame) use ($grid){
+    \Atk4\Ui\Message::addTo($frame, ['Attention - you are doing something dangerous!']);
+    $modalmodel = clone $grid->model;
+    $modalmodel->load($frame->getApp()->stickyGet('mid'));
+
+    $modalform = \Atk4\Ui\Form::addTo($frame);
+    $modalform->setModel($modalmodel);
+
+    $modalform->onSubmit(function($form) use ($grid) {
+        $form->model->save();
+        
+        $modal_chain = new \atk4\ui\jQuery('.atk-modal'); // Close the modal form? Is there a more elegant way?
+        $modal_chain->modal('hide');
+        
+        return [$modal_chain,
+            new \Atk4\Ui\JsToast([
+                'title'   => 'Save form data.',
+                'message' => 'Message is saved.',
+                'class'   => 'success',
+            ]),
+            new \Atk4\Ui\JsReload($grid),
+            
+        ];
+    });
+});
+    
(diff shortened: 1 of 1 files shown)


demos/collection/grid.php
```diff
@@ -34,6 +34,51 @@
...
+    
+$modal = new \Atk4\Ui\JsModal('Edit', $modalviewvp->getUrl('cut'), ['mid' => $grid->table->jsRow()->data('id')]);
+$grid->addActionButton('Modal form in VP', $modal);
+
+$grid->addActionButton(['icon'=>'envelope'], function ($v, $id) use ($grid) {
+    
+    $model = (clone $grid->model)->load($id);
+    $countryname =  $model->get('name');
+    $model->set('name', lcfirst($countryname));
+    $model->save();
```
ibelar, line 72: The demo file should not save to DB directly.

demos/collection/grid.php
```diff
@@ -34,6 +34,51 @@
...
+$modalviewvp->set(function ($frame) use ($grid){
+    \Atk4\Ui\Message::addTo($frame, ['Attention - you are doing something dangerous!']);
+    $modalmodel = clone $grid->model;
+    $modalmodel->load($frame->getApp()->stickyGet('mid'));
+
+    $modalform = \Atk4\Ui\Form::addTo($frame);
+    $modalform->setModel($modalmodel);
+
+    $modalform->onSubmit(function($form) use ($grid) {
+        $form->model->save();
```
ibelar, line 47: The demo file should not save to DB directly.

demos/collection/grid.php
```diff
@@ -34,6 +34,51 @@
...
+        $modal_chain = new \atk4\ui\jQuery('.atk-modal'); // Close the modal form? Is there a more elegant way?
+        $modal_chain->modal('hide');
+        
+        return [$modal_chain,
+            new \Atk4\Ui\JsToast([
+                'title'   => 'Save form data.',
+                'message' => 'Message is saved.',
+                'class'   => 'success',
+            ]),
+            new \Atk4\Ui\JsReload($grid),
```
ibelar, line 58: Perhaps reload $grid->container instead of just $grid in order to see spinner while reloading.

ibelar (commented):
I understand you would like to add some examples.

Please use cameCase for your variable name: `$modalmodel` to `$modalModel` etc...
Also, note that we do not save data directly in DB otherwise, DB could quickly contain unrelated data.

ibelar:
@mkrecek234 - @romaninsh just open Discussion for this repo. What do you think about moving your example into 'Show and Tell' category instead of a PR? I think it would be a good example to show to the community.

mkrecek234:
Up to you. I just thought that two actions showing both a modal view and doing database operations can help.

We can also delete the PR if it does not fulfill your standards.

Still, Please see the issue: The database operation on id=1 is not performed for whatever reason. Has to be a bug.

ibelar:
@mkrecek234 - We definitely cannot save directly in db using `$model->save() `. Which might break the purpose of demonstrating what you want into a demo file. So probably best within Discussion 'Show and Tell' category. Are you ok with this? 

Besides, I did run your demo and I do not see any issue with id:1 record when using lowercase action in develop.

ibelar:
@mkrecek234 - Closing, for now. Feel free to reopen if needed or add the example to the discussion.

PR #1569, opened 2020-12-10
[feature] add Crud::getUserActionMenuItem method to provide access to menu items

georgehristov:


ibelar:
@georgehristov - I have to put a hold on this one. This is in process of refactoring in #1550 
if it goes through then this pr will not be necessary because menuItem will hold executor instance instead of model action.

PR #1573, opened 2020-12-17
Fix App::encodeJson regex backtrack limit issue

### LINKED ISSUE #1540: App::encodeJson(): catch case that preg_replace_callback fails
Hi,

I recently managed (don't know how, but not too important) to get this error message: 

Bildschirmfoto von 2020-11-17 21-16-16

Logic behind this seems that preg_replace_callback() returns null on failure. We should take care of this.
I'd go for this, WDYT?

```

    public function encodeJson($data, bool $forceObject = false): string
    {
        $options = JSON_UNESCAPED_SLASHES | JSON_PRESERVE_ZERO_FRACTION | JSON_UNESCAPED_UNICODE | JSON_PRETTY_PRINT;
        if ($forceObject) {
            $options |= JSON_FORCE_OBJECT;
        }

        $json = json_encode($data, $options | JSON_THROW_ON_ERROR, 512);

        // IMPORTANT: always convert large integers to string, otherwise numbers can be rounded by JS
        // replace large JSON integers only, do not replace anything in JSON/JS strings
        $jsonLargeNumbers = preg_replace_callback('~(?:"(?:[^"\\\\]+|\\\\.)*")?+\K|(?:\'(?:[^\'\\\\]+|\\\\.)*\')?+\K|(?:^|[{\[,:])'
            . '[ \n\r\t]*\K-?[1-9]\d{15,}(?=[ \n\r\t]*(?:$|[}\],:]))~s', function ($matches) {
                if ($matches[0] === '' || abs((int) $matches[0]) < (2 ** 53)) {
                    return $matches[0];
                }

                return '"' . $matches[0] . '"';
            }, $json);
        
        if($jsonLargeNumbers == null) {
            return $json;
        }
       
        return $jsonLargeNumbers;
    }
```

mvorisek:
fixes https://github.com/atk4/ui/issues/1540


tests/jsTest.php
```diff
@@ -27,7 +27,9 @@ public function testNumbers()
             $this->markTestIncomplete('Test is not supported on 32bit php');
         }
 
+        $longStr = str_repeat('"a":10,"b":9007199254740992,x="\"c\":10,\"d\":9007199254740992,"', 50 * 1000);
```
mvorisek: FYI the issue is reproducible even with 1000 repetitions

georgehristov (approved):
LGTM

PR #1575, opened 2020-12-18
[fix] Id generation when using Executor as Modal View

ibelar:
Remove special characters, like dot, from table name when generating View::id.


PhilippGrashoff:
am curious: why is the table name needed at all?

ibelar:
> am curious: why is the table name needed at all?

If I remember correctly, it was because of callback issues. 
Modal view are generated outside their parent and since Executor are using modal and related on their name or id to carry the model id being used by the action, it was causing issue when executor was create from a reload.
That why we needed a way to make sure Modal name (id) was the same for each Model user action, with no regards to their location in the render tree.

mvorisek:
seems not abstract enough, how name is build normally, is there a method for it?

src/UserAction/ConfirmationExecutor.php
```diff
@@ -68,7 +68,7 @@ public function afterActionInit(Model\UserAction $action)
 
         $table_name = is_array($action->getModel()->table) ? $getTableName($action->getModel()->table) : $action->getModel()->table;
 
-        $this->id = mb_strtolower($this->name . '_' . $table_name . '_' . $action->short_name);
+        $this->id = preg_replace('~[^A-Za-z0-9\-]~', '_', mb_strtolower($this->name . '_' . $table_name . '_' . $action->short_name));
```
mvorisek, line 71: `extends Modal implements JsExecutorInterface` ... the same base

can we dedup these classes a little bit?

mvorisek:
So executor with a non unique name is put to some non-unique render tree?

This should have a Behat test.

src/UserAction/ConfirmationExecutor.php
```diff
@@ -61,17 +61,6 @@ protected function init(): void
      */
     public function afterActionInit(Model\UserAction $action)
     {
-        $getTableName = function ($arr) {
```
mvorisek, line 64: why it was needed/is not needed now?

ibelar: Was related to previous callbacks issue with reloading and the fact that the executor was added to $app->html.
Should not be needed now.

mvorisek: perfect! 👍

mvorisek (approved):

PR #1577, opened 2020-12-19
Feature/ Add Lookup to Multiline

### LINKED ISSUE #1560: WIP - Add lookup to Multiline
Add Lookup to Multiline control.

ibelar:
- Multiline control and components refactor.
- Now support the Lookup component. Fix: #1560 
 
Screen Shot 2020-12-19 at 1 45 11 PM

## BC Break

1) `$options` property has been removed and passing Vue Props to components is now done using the new `$componentProps` property. Passing Props to each Vue component also required to properly identify the component.

Ex: passing Dropdown Props to all Dropdown component used in Multiline:
```
$componentProps = [Multiline::SELECT => ['floating' => true]]
```

2) Passing component Props via `$field->ui `property also required specifying the component name. Array key 0 is no longer support.

Ex of setting Input component Props via` $field->ui`:

```
$this->addField('password', [
    'required' => true,
    'ui' => ['multiline' => [Multiline::INPUT => ['icon' => 'key', 'type' => 'password']]],
]);
```

3) Since `$options` have been removed, passing Table Props is now done via the` $tableProps` property.

```
$multiline = $form->addControl('ml', [Multiline::class, 'tableProps' => ['color' => 'blue']]);
```

4) Controlling each cell component Props, like width, using `$field->ui ` also required to name that component.

```
$this->addExpression('total', [
    'expr' => function (Model $row) {
        return $row->get('qty') * $row->get('box');
    },
    'type' => 'integer',
    'ui' => ['multiline' => [Multiline::TABLE_CELL => ['width' => 1, 'class' => 'blue']]],
]);

```


DarkSide666 (approved):
LGTM

src/Callback.php
```diff
@@ -53,7 +53,7 @@ public function getUrlTrigger(): string
     /**
      * Executes user-specified action when call-back is triggered.
      *
-     * @param \Closure $fx
+     * @param callable $fx
```
mvorisek, line 56: we should keep `\Closure` to be stricter and prevent contextual issues with generic callable

PR #1585, opened 2021-01-05
Fix autoload for demos with classmap authoritative

mvorisek:


demos/init-autoloader.php
```diff
@@ -10,5 +10,6 @@
 if (!$isRootProject && !class_exists(\Atk4\Ui\Tests\ViewTest::class)) {
     throw new \Error('Demos can be run only if atk4/ui is a root composer project or if dev files are autoloaded');
 }
+$loader->setClassMapAuthoritative(false);
```
mvorisek, line 13: otherwise on runtime `setPsr4` is not guaranteed to work

georgehristov (approved):
LGTM

PR #1593, opened 2021-01-20
Use also different actual name

mvorisek:


georgehristov (approved):

mvorisek:
@georgehristov seems atk4/data does not resolve their_field etc. correctly (using model's `actual`). Would you like to help with this?

georgehristov:
@mvorisek yes, I will have a look

mvorisek:
@georgehristov or drop `actual` completely? (with hintable, why do we need it at all?)

DarkSide666 (approved):
ok

mvorisek:
@georgehristov @DarkSide666 this data bug should be fixed asap, can you help with the fix?

PR #1595, opened 2021-01-20
Prefix/test also all ID fields

mvorisek:


georgehristov (approved):

demos/collection/lister-ipp.php
```diff
@@ -50,7 +50,7 @@
     $row->iso = mb_strtolower($row->iso);
 });
 $lister->setModel(new Country($app->db))
-    ->addCondition('id', -1); // no such records so model will be empty
+    ->addCondition(Country::hinting()->fieldName()->id, -1); // no such records so model will be empty
```
DarkSide666, line 53: Why not `(new Country())->id_field)` ?

mvorisek: also possible, but this seems better to me when we expect hintable `id` is id field

mvorisek: best would be `addCondition(false)`, right? :)

/cc @georgehristov

demos/init-db.php
```diff
@@ -188,6 +196,7 @@ protected function init(): void
 
         $this->hasOne($this->fieldName()->client_country_iso, [
             'model' => [Country::class],
+            'our_field' => $this->fieldName()->client_country_iso,
```
mvorisek: When prefixing is used, guessing `our_field` for hasOne doesn't work

code: https://github.com/atk4/data/blob/68aa5efd536820d1d7b92061f4c18786425c77e9/src/Reference/HasOne.php#L157

is it possible to fix that?

demos/collection/lister-ipp.php
DarkSide666: Or `addCondition(1, '<>', 1)` :)

DarkSide666 (approved):

demos/init-db.php
DarkSide666: Looks like it's working now

PR #1600, opened 2021-01-26
[fix] - Demo Tutorial

ibelar:
fix - Intro Exception in Business model step;
fix - Make delete action enable property with predictable callback in order to prevent exception.

### DIFF
--- a/demos/_includes/DemoInvoice.php
+++ b/demos/_includes/DemoInvoice.php
@@ -25,6 +25,9 @@ class DemoInvoice extends \Atk4\Data\Model
                 function ($v) {
                     return ($v instanceof \DateTime) ? date_format($v, $this->dateFormat) : $v;
                 },
+                function ($v) {
+                    return $v;
+                },
             ],
         ]);
     }
--- a/demos/tutorial/actions.php
+++ b/demos/tutorial/actions.php
@@ -189,7 +189,7 @@ $wizard->addStep('Crud integration', function ($page) {
     $page->add(new Demo())->setCodeAndCall(function (View $owner) {
         $country = new \Atk4\Ui\Demos\CountryLock($owner->getApp()->db);
         $country->getUserAction('add')->enabled = false;
-        $country->getUserAction('delete')->enabled = function ($m) { return $m->id % 2 === 0; };
+        $country->getUserAction('delete')->enabled = function (\Atk4\Ui\Demos\Country $m) { return $m->id % 2 === 0; };
         $country->addUserAction('mail', [
             'appliesTo' => \Atk4\Data\Model\UserAction::APPLIES_TO_SINGLE_RECORD,
             'preview' => function (CountryLock $country) { return 'here is email preview for ' . $country->name; },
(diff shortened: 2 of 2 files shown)


demos/tutorial/actions.php
```diff
@@ -189,7 +189,7 @@ functionality and more. Next example shows how you can disable user action (add)
     $page->add(new Demo())->setCodeAndCall(function (View $owner) {
         $country = new \Atk4\Ui\Demos\CountryLock($owner->getApp()->db);
         $country->getUserAction('add')->enabled = false;
-        $country->getUserAction('delete')->enabled = function () { return random_int(1, 2) > 1; };
+        $country->getUserAction('delete')->enabled = function ($m) { return $m->getId() % 2 === 0; };
```
mvorisek:
```suggestion
        $country->getUserAction('delete')->enabled = function ($m) { return $m->id % 2 === 0; };
```

demos/_includes/DemoInvoice.php
```diff
@@ -25,6 +25,9 @@ protected function init(): void
                 function ($v) {
                     return ($v instanceof \DateTime) ? date_format($v, $this->dateFormat) : $v;
                 },
+                function ($v) {
```
mvorisek, line 28: I wonder, what is default, when the 2nd typecast closure is not specified.

ibelar: Not sure what you mean by default but when the closure is not specify here is the Exceptions thrown.

Screen Shot 2021-01-27 at 8 28 12 AM
Screen Shot 2021-01-27 at 8 28 25 AM

mvorisek: I mean what typecast is used when it is not specified.

ibelar: This is used with Persistence\Array_, not sure about this but to my understanding, Array_ persistence does not provide default typecast..

mvorisek: the exceptions seems to provide "default typecast"... Otherwise it will not warn and dummy closure will be not needed

mvorisek: so I think the issue is somewhere else

demos/tutorial/actions.php
```diff
@@ -189,7 +189,7 @@ functionality and more. Next example shows how you can disable user action (add)
     $page->add(new Demo())->setCodeAndCall(function (View $owner) {
         $country = new \Atk4\Ui\Demos\CountryLock($owner->getApp()->db);
         $country->getUserAction('add')->enabled = false;
-        $country->getUserAction('delete')->enabled = function () { return random_int(1, 2) > 1; };
+        $country->getUserAction('delete')->enabled = function ($m) { return $m->id % 2 === 0; };
```
mvorisek:
```suggestion
        $country->getUserAction('delete')->enabled = function (\Atk4\Ui\Demos\Country $m) { return $m->id % 2 === 0; };
```

DarkSide666 (approved):
LGTM

demos/_includes/DemoInvoice.php
ibelar: Looks like a data  Persistence\Array issue. Will merge has is for now so the demo does not crash.

Issue #1587, opened 2021-01-08
Lookup values are not loaded when callbacks are in a particular order


acicovic:
See https://github.com/acicovic/atk4-lookup-test

ibelar:
@acicovic - Thanks for supplying code for this bug.

ibelar:
fix in #1602

PR #1602, opened 2021-02-01
[fix] VirtualPage

ibelar:
Fix for #1587 

The callback `Callback::set()` method needs to be called prior to VirtualPage::getHtml() method in order for other's callback inside VirtualPage to be execute properly.

## BC Break

- Property `$fx`  was remove and no longer necessary because `Callback::set()` method is called during VirtualPage::set() and not during `VirtualPage::getHtml()` as before. This is to ensure that  `Callback::set()` is always run. 

    Therefore, using DI to set $fx property via` VirtualPage::addTo($app, ['fx' => function(){}]) `is no longer supported.

- Virtual::Page::set() only accept \Closure instead of callable as before. You must convert your callable into Closure via \Closure::fromCallable()

### DIFF
--- a/demos/_unit-test/lookup-virtual-page.php
+++ b/demos/_unit-test/lookup-virtual-page.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * Behat testing.
+ * Test for Lookup inside VirtualPage.
+ */
+
+namespace Atk4\Ui\Demos;
+
+use Atk4\Ui\Form;
+use Atk4\Ui\Grid;
+use Atk4\Ui\JsModal;
+use Atk4\Ui\JsToast;
+use Atk4\Ui\VirtualPage;
+
+/** @var \Atk4\Ui\App $app */
+require_once __DIR__ . '/../init-app.php';
+
+$product = new ProductLock($app->db);
+
+$vp = VirtualPage::addTo($app);
+
+$vp->set(function ($page) {
+    $form = Form::addTo($page);
+    $form->addControl('category', [Form\Control\Lookup::class, 'model' => new Category($page->getApp()->db)]);
+    $form->onSubmit(function ($f) {
+        $category = $f->getControl('category')->model->load($f->model->get('category'));
+
+        return new JsToast($category->gettitle());
+    });
+});
+
+$g = Grid::addTo($app, ['menu' => ['class' => ['atk-grid-menu']]]);
+$g->setModel($product);
+
+$g->menu->addItem(
+    ['Add Category'],
+    new JsModal('New Category', $vp)
+);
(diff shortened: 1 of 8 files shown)


src/VirtualPage.php
```diff
@@ -40,26 +37,18 @@ protected function init(): void
...
      *
      * @return $this
      */
-    public function set($fx = [], $junk = null)
+    public function set($fx = null, $junk = null)
     {
-        if (!$fx) {
-            return $this;
+        if (!$fx || !is_callable($fx)) {
+            throw new Exception('Virtual page required a callable.');
```
PhilippGrashoff: Typo here, "requires", not past tense "required" ;)

ibelar: update - thanks

demos/interactive/virtual.php
```diff
@@ -68,7 +68,7 @@
 
 $frame = \Atk4\Ui\VirtualPage::addTo($app);
 $frame->set(function ($frame) {
-    \Atk4\Ui\Header::addTo($frame, ['Clicked row with ID = ' . $_GET['id']]);
+    \Atk4\Ui\Header::addTo($frame, ['Clicked row with ID = ' . ($_GET['id'] ?? null)]);
```
mvorisek:
```suggestion
    \Atk4\Ui\Header::addTo($frame, ['Clicked row with ID = ' . ($_GET['id'] ?? '')]);
```

src/VirtualPage.php
```diff
@@ -40,26 +37,18 @@ protected function init(): void
...
+     * @param mixed    $junk
      *
      * @return $this
      */
-    public function set($fx = [], $junk = null)
+    public function set($fx = null, $junk = null)
     {
-        if (!$fx) {
-            return $this;
+        if (!$fx || !is_callable($fx)) {
```
mvorisek:
```suggestion
        if (!$fx instanceof \Closure) {
```

src/VirtualPage.php
```diff
@@ -40,26 +37,18 @@ protected function init(): void
     /**
      * Set callback function of virtual page.
      *
-     * Note that only one callback function can be defined.
-     *
-     * @param array $fx   Need this to be defined as array otherwise we get warning in PHP7
-     * @param mixed $junk
+     * @param callable $fx   Need this to be defined as array otherwise we get warning in PHP7
```
mvorisek:
```suggestion
     * @param \Closure $fx
```

demos/_unit-test/lookup-virtual-page.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * Behat testing.
```
mvorisek: delete this comment, implied by filename

demos/interactive/virtual.php
ibelar: update - thanks

src/VirtualPage.php
ibelar: Done

demos/_unit-test/lookup-virtual-page.php
ibelar: Done

src/VirtualPage.php
ibelar: Done

DarkSide666 (approved):
LGTM

PR #1603, opened 2021-02-04
[fix] VirtualPage inside VirtualPage

ibelar:
Allow using VirtualPage inside others VirtualPage.

- add Behat test for VirtualPage.


DarkSide666 (approved):
LGTM

Issue #1607, opened 2021-02-20
Wrong/missing charset on JSON response


FabulousGee:
I had some issues with Umlauts and so I double-checked everything on my side. All fine.
But when I checked the header-handling in atk I found a possible bug. When I set the charset on termination myself, it gets omitted.

I use the following line of code to output my JSON:
```php
$this->_app->terminate(json_encode($data), ['Cache-Control' => 'no-transform,public,max-age=300,s-maxage=900', 'content-type' => 'application/json; charset=utf-8', 'Status' => ''.$code]);
```

Note the "charset=utf-8" to set the encoding (which should be standard on JSON but isn't necessary to set by HTTP guidelines)

When I now look at \atk4\ui\src\App.php I saw the content-type being hardcoded:
```php
private function outputResponseJson($data, array $headers = []): void
    {
        if (!is_string($data)) {
            $data = $this->encodeJson($data);
        }
        $this->outputResponse(
            $data,
            array_merge($this->normalizeHeaders($headers), ['content-type' => 'application/json'])
        );
    }
```

I think it should be considered to either hardcode it to "['content-type' => 'application/json; charset=utf-8']" or check the original content-type more properly and reapply any attached strings again

FabulousGee:
@DarkSide666 @mvorisek regarding the discussion on my pull request, I want to continue the discussion about it here. The PR was just my first idea, I am sure there could be a better way. For example it would suffice to check, if the content-type has already been set and just skip it then (to let the user override it if they want to)

I am not into all of these things really, but I read a lot about it the last days and e.g. in [this StackOverflow thread](https://stackoverflow.com/questions/43148464/how-do-browsers-determine-the-encoding-used) they mention that the charset (encoding) isn't implied as a default but furthermore should always be explicitely stated.
Also, there is a deeper conversation with some serious sources (like some HTTP specs) about the topic on the libraries page on GitHub as well: https://github.com/dart-lang/http/issues/175#issuecomment-619593544

So I still think a change on atk side should really be considered, no matter which way is preferred

mvorisek:
Which browser/client does detect the encoding wrongly?

PS: your PR is for JSON mime, which should be defaulted to UTF-8. HTML5 has UTF-8 encoding as default too.

FabulousGee:
It's the HTTP library which I use in dart (Flutter) - I linked the issue above. Did you read through the post on their tracker? They describe very well which defaults apply and which do not.

[...]

PR #1608, opened 2021-02-20
Change default set charset on JSON response

FabulousGee:
Check corresponding issue #1607 for further discussion.

Maybe this should be made more generic?

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -1094,7 +1094,7 @@ class App
 
         $this->outputResponse(
             $data,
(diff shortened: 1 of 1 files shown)


DarkSide666:
I believe charset should be  set at App level. At least as specific App property and not hardcoded like this.
And it should then be reused when rendering HTML page and/or json response or any other server response.

mvorisek:
UTF-8 is already implied as a default by modern browsers

PR #1613, opened 2021-04-08
[fix] registerExecutor parameter type

ibelar:
Allow using Object instance as seed.


mvorisek (approved):

mvorisek:
merged but the seed object is not instanced neewly, thus allowing object seed in general may lead to very strange results, be careful

ibelar:
Is it better to check for proper instance then and throw error if not array and not of type ExecutorInterface?

mvorisek:
Phpstan will do this check if enabled in project where used.

The issue is when you pass some object seed, that is expected to be instanced for every use.

ibelar:
yes right, I misunderstood your comment. Perhaps check should be at the object creation in order to make sure the created instance if of type ExecutorInterface.

PR #1623, opened 2021-04-26
Upgrade atk4/data with strict entity model

mvorisek:
related with https://github.com/atk4/data/pull/862


demos/basic/breadcrumb.php
```diff
@@ -23,7 +23,7 @@
 
 if ($id = $app->stickyGet('country_id')) {
     // perhaps we edit individual country?
-    $model->load($id);
+    $model = $model->load($id);
```
PhilippGrashoff, line 26: how about $entity = $model->load() or $record = $model->load()? Would be much clearer naming the variables accordingly.

mvorisek: For simplicity, this PR will be merged with the same variable names. It is usualy more than evident if something represent a single entity or a general model. If not obvious, you can always check easily with `Model::isEntity` method or an assertion methods like `Model::assertIsEntity` :)

src/Card.php
```diff
@@ -252,7 +252,7 @@ public function addModelActions(Model $model)
      */
     public function addSection(string $title = null, Model $model = null, array $fields = null, bool $useTable = false, bool $useLabel = false)
     {
-        $section = $this->add([$this->cardSection, 'card' => $this], 'Section');
+        $section = CardSection::addTo($this, ['card' => $this], ['Section']);
```
mvorisek, line 255:
```suggestion
        $section = CardSection::addToWithCl($this, [$this->cardSection, 'card' => $this], ['Section']);
```

@ibelar why this change?

src/UserAction/JsCallbackExecutor.php
```diff
@@ -62,9 +62,9 @@ public function executeModelAction(array $args = [])
     {
         $this->set(function ($j) {
             // may be id is passed as 'id' or model->id_field within $post args.
-            $id = $_POST['c0'] ?? $_POST['id'] ?? $_POST[$this->action->getOwner()->id_field] ?? null;
-            if ($id && $this->action->appliesTo === Model\UserAction::APPLIES_TO_SINGLE_RECORD) {
-                $this->action->getOwner()->tryLoad($id);
+            $id = $_POST['c0'] ?? $_POST['id'] ?? $_POST[$this->action->getModel()->id_field] ?? null;
+            if ($id && $this->action->appliesTo === Model\UserAction::APPLIES_TO_SINGLE_RECORD && !$this->action->getModel()->isEntity()) {
```
mvorisek: `!$this->action->getModel()->isEntity()` will be of course always true ;-)

ibelar: True, I simply forgot to remove it. :-)

demos/_includes/DemoActionsUtil.php
```diff
@@ -82,8 +84,8 @@ public static function setupDemoActions(CountryLock $country): void
             'edit_iso',
             [
                 'caption' => 'Edit ISO3',
-                'description' => function () {
-                    return 'Edit ISO3 for country:';
+                'description' => function (UserAction $action) {
+                    return 'Edit ISO3 for country: ' . $action->getEntity()->getTitle();
```
ibelar, line 88: No need to put it back since it will always be empty. This evaluate prior to entity being set.

mvorisek: I noticed that. It is easy fixable?

ibelar: Dialog title is set when Modal is added to app->html which is always done prior to executing the action I believe. 
Perhaps in Modal Executor we could run javascript for changing Modal title when action is execute after entity is set.
As of now we can only use it to display Action or Model information only; not entity.

mvorisek: > Dialog title is set when Modal is added to app->html which is always done prior to executing the action I believe.

Then this must be addressed, no JS please. I will keep the code with "empty entity title" so it can be fixed later...

I also think we can make `getTitle` throw easily if entity is passed, but unloaded :) (a lot of other issue will pop up)

demos/layout/layouts_admin.php
```diff
@@ -37,7 +37,7 @@
 \Atk4\Ui\Header::addTo($layout, ['Basic Form Example']);
 
 $form = \Atk4\Ui\Form::addTo($layout, ['segment']);
-$form->setModel(new \Atk4\Data\Model());
+$form->setModel((new \Atk4\Data\Model())->createEntity());
```
ibelar, line 40: Would it be better to have Form::setModel() method to check for an entity and create an empty one if model is passed? This would allow passing a model instance as before.

mvorisek: I am for as much less guessing as possible. In user action class, the guessing is already more like for BC.

My question here is - why would anyone pass model to a form, which can always represent only one entity?

ibelar: I agree with you but AFAIK there is no entity class and the method name is still setModel() not setEntity()...
I think it would be confusing the way it is. Until we can pass an Entity without being a Model, I think we should check this in setModel()

mvorisek: I understand it can be confusing, but when entity should be allowed only, it is better to not allow anything else. When model is passed, very clear exception is throw and the developer should grasp very intuitively that only entity can be passed for components that operate with one record only. The same for model, when model is excepted, entity is rejected.

ibelar: Probably best changing method name then eventually, but I am ok with this now.

ibelar (approved):
LGTM

PR #1625, opened 2021-04-28
Replace call_user_func() calls as much as possible

mvorisek:
`call_user_func()` calls cannot be fully removed as long as property defaults should be supported - they are php-src constant expressions and they currently do not support Closures, not even first class callables.

### DIFF
--- a/src/Form/Control/Multiline.php
+++ b/src/Form/Control/Multiline.php
@@ -108,7 +108,7 @@ class Multiline extends Form\Control
     /** @var array SuiTable component props */
     public $tableProps = [];
 
-    /** @var array<string, array{component: string, componentProps: mixed}> Set Vue component to use per field type. */
+    /** @var array<string, array<string, mixed>> Set Vue component to use per field type. */
     protected $fieldMapToComponent = [
         'default' => [
             'component' => self::INPUT,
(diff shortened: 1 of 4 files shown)


src/Form/Control/Multiline.php
```diff
@@ -690,7 +690,7 @@ private function outputJson(): void
 
                 break;
             case 'on-change':
-                $response = call_user_func($this->onChangeFunction, $this->getApp()->decodeJson($_POST['rows']), $this->form);
```
mvorisek, line 693: `call_user_func` is slow and adds unneeded backtrace frame

src/UserAction/ExecutorFactory.php
```diff
@@ -202,7 +202,7 @@ protected function createActionTrigger(UserAction $action, string $type = null):
             $seed = $viewType['default'];
         }
 
-        $seed = is_array($seed) && is_callable($seed) ? call_user_func($seed, $action, $type) : $seed;
+        $seed = is_array($seed) && is_callable($seed) ? $seed($action, $type) : $seed;
```
mvorisek, line 205: @ibelar can you please fix the seed creation above? For some reasons, array seed can not be invoked as expected not mentioning, in theory, array seed can callable (`is_callable($seed) === true`), but intended to be standard seed - https://3v4l.org/mMgpT

PR #1629, opened 2021-05-01
Use hintable property names in seeds

mvorisek:
related with https://github.com/atk4/core/pull/315

make seed keys refactorable


georgehristov (approved):

mvorisek:
factory validates these names, although not hintable...

PR #1631, opened 2021-05-07
[fix] Multiline

ibelar:
Align with latest data changes.

### BC-BREAK

Multiline->setReferenceModel() method parameters has changed.

You now have to passed the reference name and the model entity to where reference apply:

````
$multiline->setReferenceModel('Products', $categoryModel->load($id));
````

### DIFF
--- a/src/Form/Control/Multiline.php
+++ b/src/Form/Control/Multiline.php
@@ -869,7 +869,7 @@ class Multiline extends Form\Control
             case 'integer':
             case 'float':
                 // Value is 0 or the field value.
-                $value = $model->get($fieldName) ?: 0;
+                $value = (string) $model->get($fieldName) ?: 0;
 
                 break;
             default:
--- a/src/Form/Control/Multiline.php
+++ b/src/Form/Control/Multiline.php
@@ -210,7 +210,7 @@ class Multiline extends Form\Control
 
         // load the data associated with this input and validate it.
         $this->form->onHook(Form::HOOK_LOAD_POST, function ($form, &$post) {
-            $this->rowData = $this->loadRowData();
+            $this->rowData = $this->typeCastLoadValues($this->getApp()->decodeJson($_POST[$this->short_name]));
             if ($this->rowData) {
                 $this->rowErrors = $this->validate($this->rowData);
                 if ($this->rowErrors) {
@@ -247,11 +247,11 @@ class Multiline extends Form\Control
     /**
      * Typecast each loaded value.
      */
-    protected function loadRowData()
+    protected function typeCastLoadValues(array $values): array
     {
         $dataRows = [];
 
-        foreach ($this->getApp()->decodeJson($_POST[$this->short_name]) as $k => $row) {
+        foreach ($values as $k => $row) {
             foreach ($row as $fieldName => $value) {
                 if ($fieldName === '__atkml') {
                     $dataRows[$k][$fieldName] = $value;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/Form/Control/Multiline.php [hunk 1]
```diff
@@ -324,24 +344,22 @@ public function saveRows(): self
...
 
-                $field = $row_model->getField($fieldName);
-                if ($field->isEditable()) {
-                    $field->set($value);
+                if ($model->getField($fieldName)->isEditable()) {
+                    $entity->set($fieldName, $value);
                 }
             }
-            $id = $row_model->save()->getId();
+            $id = $entity->save()->getId();
```
mvorisek, line 357: save is outside foreach, is that intended?

otherwise feel free to merge it

src/Form/Control/Multiline.php
```diff
@@ -324,24 +344,22 @@ public function saveRows(): self
...
 
             foreach ($row as $fieldName => $value) {
                 if ($fieldName === '__atkml') {
                     continue;
                 }
 
-                if ($fieldName === $row_model->id_field && $value) {
-                    $row_model = $row_model->load($value);
+                if ($fieldName === $model->id_field && $value) {
+                    $entity = $model->load($value);
```
PhilippGrashoff: Theoretically the id_field could be anywhere in the array, couldn't it? So it could be that some values are already set to the entity by the rows below ($entity->set($fieldName, $value), and then $entity->load($value) comes after. That could overwrite values already set by $entity->set($fieldName, $value).
I think load() should be the first thing happening, any set() should happen after.

src/Form/Control/Multiline.php
ibelar: Yes - the intention is to collect all record values for one row into the for loop and save after.

src/Form/Control/Multiline.php
ibelar: you are right - I guess it never breaks before probably because id is generally the first field in a row. I will make changes. Good catch!

src/Form/Control/Multiline.php
mvorisek: but a few lines above there is a `load`... related with https://github.com/atk4/ui/pull/1631/files#r628487155

ibelar: bad explanation from me. 

The data collect from load contains all rows and each row has one set of record. 
The second for loop is too loop for each field and check if the current field is editable prior to set is value. 
Once all fields are set within the row, then the record is save.

Hope it is clearer now... 

Maybe we could use $entity->save($row) directly but I am not sure how Data will react if trying to pass a field => value set to a non editable field... This is were my knowledge of Data fails 😄 Multiline will display the non editable field as readonly input.

src/Form/Control/Multiline.php
ibelar: update

src/Form/Control/Multiline.php
mvorisek: AFAK this relies on :N typecasting from single cell value which... Which at least is not good to be saved n times when looping.

feel free to merge without my review once you are happy with the changes ;-)

src/Form/Control/Multiline.php
```diff
@@ -420,13 +433,15 @@ public function setModel(Model $model, array $fieldNames = []): Model
      * Otherwise, form will try to save 'multiline' field value as an array when form is save.
      * $multiline = $form->addControl('multiline', [Multiline::class], ['never_persist' => true])
      */
-    public function setReferenceModel(Model $refModel, string $linkByFieldName, array $fieldNames = []): Model
+    public function setReferenceModel(string $refModelName, Model $modelEntity = null, array $fieldNames = []): Model
```
mvorisek, line 436: search `setReferenceModel`, docs needs to be adjusted.

In general, do we need to pass `Model` at all?

ibelar: Doc updates. In general, the Model entity will be set using the model set in Form. But there might be cases where you need to pass it with the method.

mvorisek (approved):

PR #1636, opened 2021-05-11
[fix] Js Dependencies

ibelar:


mvorisek:
@ibelar did you enabled dependabot or why were the #1634 and #1635 PR created? As the JS files neeeds to be recompiled anyway, I think dependabot is useless for this repo and should be disabled.

mvorisek (approved):

mvorisek (approved):

PR #1646, opened 2021-07-06
Improve callbacks url generation

ibelar:
- move back URL method to View class;
- remove Modal $viewUrl property hack;
- simplify URL generation for callbacks;


src/AbstractView.php
```diff
@@ -119,91 +119,4 @@ public function add($object, $args = null): self
 
         return $object;
     }
-
-    // }}}
-
-    // {{{ Sticky URLs
-
-    /** @var string[] stickyGet arguments */
-    public $stickyArgs = [];
```
mvorisek, line 128: So Callback can not longer have sticky args?

src/UserAction/ExecutorFactory.php
```diff
@@ -172,19 +171,7 @@ protected function createExecutor(UserAction $action, View $owner, string $requi
...
-                // very dirty hack, @TODO, attach modals in the standard render tree
-                // but only render the result to a different place/html DOM
-                $executor->viewForUrl = $owner;
-                $executor = $owner->getApp()->html->add($executor, 'Modals'); //->setAction($action);
-            }
-        } else {
-            $executor = $owner->add($executor);
-        }
-
+        $executor = $owner->add(Factory::factory($seed));
```
mvorisek, line 174: We make Modal standard part of render tree, this is the trick?

If yes, well done!

src/AbstractView.php
ibelar: Yes, only View have stickyArgs

src/UserAction/ExecutorFactory.php
ibelar: Yes, but since Fomantic will move the modal HTML elsewhere in the document, Modal has to be registered within App in order for it to re-render them for JSON output.

src/AbstractView.php
mvorisek: Callback can be API that should preserve args. I am neutral on this change, but I would personally keep it in AbstractView. Adding BC break label at least.

src/UserAction/ExecutorFactory.php
mvorisek: 👍

src/Wizard.php
```diff
@@ -181,7 +155,7 @@ public function urlNext()
      */
     public function jsNext()
     {
-        return new JsExpression('document.location = []', [$this->urlNext()]);
+        return new JsExpression('document.location = []', [$this->getUrl($this->currentStep + 1)]);
```
mvorisek: `$this->urlNext()` seems fine to me

src/AbstractView.php
ibelar: Since sticky get is applied to the owner of the callback (View) I think it is ok to have stickyArgs property to View only.

I believe that Callback should not manage URL at all but only the arguments needed to run the callback function. It should be to the View using the callback to properly managed its URL. 

For example in Loader::jsLoad() method it should be:

````
$this->url($this->cb->getUrlArguments()),
````
instead of:
```
$this->cb->getUrl(),
````

File: src/Wizard.php (hunk 3 above)
Comment (ibelar) [reply src/Wizard.php]: Done.

File: src/Callback.php [hunk 4]
```diff
@@ -43,6 +43,8 @@ protected function init(): void
     public function setUrlTrigger(string $trigger = null)
     {
         $this->urlTrigger = $trigger ?: $this->name;
+
+        $this->getOwner()->stickyGet($this->urlTrigger);
```
mvorisek: Now I understand how this PR is supposed to work - what is a `Callback` definition?

Because if more than one is added, tree/independent design is gone because of shared sticky.

ibelar: A callback is a GET request, that when triggered execute a function. The function should terminate the code execution and render a View (normally the one consuming the callback) via JSON. 
Many Views may consume callback at the same time but only one of them can terminate execution.

mvorisek: I am in favor of keeping `mergeStickyArgsFromChildView` as it allows View to control which child View can modify it's sticky args. Without this method (as this PR is coded now), the Callback modifies a parent View, which introduces shared/global state which is not wanted in my opinion.

ibelar: I think `mergeStickyArgsFromChildView` is actually not necessary, or I do not understand the use case there. The callback can be triggered, i.e. execute the set function or terminate, i.e. render the proper View as JSON.
The view is actually controlling their stickyArgs, so nested View with callback can still be triggered.

mvorisek: If Callback has some args that should be sticky, with `mergeStickyArgsFromChildView` the args are stored inside the Callback object. If that object is mutated to some different state, the parent/independent view CAN (but does not have to) reflect that.

If some URL in callback needs to be generated without the Callback sticky args, with this PR, it is not possible as the args are merged in the parent View together. Imagine like generating a reload URL (of parent component) from Callback. Impossible.

With this reasoning, it makes sense to me to keep the sticky args not hardcoded with the parent component, but instead of keep `mergeStickyArgsFromChildView` method and allow the parent component to indicate that the child view args should propagate (from 1st level child or even from deeped level).

I belive when `mergeStickyArgsFromChildView` is kept, it can help solving https://github.com/atk4/ui/pull/1649#discussion_r672293568

ibelar: > If Callback has some args that should be sticky, with `mergeStickyArgsFromChildView` the args are stored inside the Callback object. If that object is mutated to some different state, the parent/independent view CAN (but does not have to) reflect that.

Only the View consuming the callback should have the stickyArgs, not the Callback itself. The callback should fire or not depending on the argument carry by the view. If a View needs to generate an URL without Callback arguments, then it can do so as well.

mvorisek: https://github.com/atk4/ui/pull/1646/files#diff-489a31f422a2238af13855d2db35e0df62c99606b21cb5c678330f7f857ea90eR43-R47

really? This method sets an URL trigger and makes it sticky. Sticky to the parent, thus when any URL from parent is generated, it is always sticky, even if Callback trigger is not wanted.

Example render tree:

```
App - View (alias V1) - Callback 1 (alias C1)
                      - Callback 2 (alias C2)
```

Let's imagine C1 is triggered.

C1 is made sticky. Can we agree here?

Now, if C1 wants to render a V1 or C2 URL, it can not, as V1 contains sticky args from C1. This is a problem. Can we agree here too?

ibelar: View generates proper Url for the Callbacks - It should know which one to generate. Callback is not a View and it is not rendered. Only the arguments that need to fire the callback should be carry by the View and this should be done by the View itself when generating the URL.

mvorisek: > Callback is not a View and it is not rendered.

It is rendered, eg. triggered = executed = rendered but terminated with a custom output, right?

What is the diffrent vs. VirtualPage?

The situation I describe is even more evident with this render tree:

```
App - View 1 (alias V1) - Callback 1 (alias C1)
                        - View 2 (alias V2)
```

When C1 is triggered and C1 is supposed (because coder wants) to return V2 URL, V2 will contains sticky args from C1, which is wrong.

ibelar: > It is rendered, eg. triggered = executed = rendered but terminated with a custom output, right?

Callback itself is not rendered. A View is, depending on the implementation.  There are two arguments need to run the callback function. One argument will trigger the callback method (canTriggered) and the other will output content via the terminate method (canTerminate).

 These arguments are:
 - The triggering argument:
    - The callback name (default) = the callback mode (mode is used in VirtualPage and determine the type of output)
       When the View consuming the callback reaches the triggering argument, then the callback method is executed.
- The current callback or callback that can terminate output.
  - __atk_callback = the callback name (default)
    This is the current callback that can terminate app execution.

Let's take VirtualPage demo https://github.com/atk4/ui/blob/develop/demos/_unit-test/virtual-page.php as an example.

In order to reach the callback that populates the Dropdown control, all previous VirtualPage callback method has to be triggered thus carrying all triggered callback argument with the URL firing the dropdown.

See below the URL firing the Dropdown request for content. As you can see from the picture, only one callback can terminate output, the Dropdown control in this case.

Screen Shot 2021-07-19 at 5 46 33 PM

If you missed one triggering argument, then the Dropdown callback can never be reached.

mvorisek: > In order to reach the callback that populates the Dropdown control, all previous VirtualPage callback method has to be triggered thus carrying all triggered callback argument with the URL firing the dropdown.

This is clear to me. VP is a component that can be nested, Callback is the possible execution target (eg. can not be nested). Is there any reason to use Callback over VP for the end-user?

> When C1 is triggered and C1 is supposed (because coder wants) to return V2 URL, V2 will contains sticky args from C1, which is wrong.

`__atk_callback` will not be set, thus it will behave correctly, but is there any need for sticky V1 or C1 args if V2 is rendered from triggered C1?

ibelar: > Is there any reason to use Callback over VP for the end-user?

Loader, Form, Console; they all use a callback. Loader can be nested. They can even be nested inside a VirtualPage.

I do not see any problem carrying over argument. A callback is used to terminate output at some point. Before the regular app->output happen. Thus you want to make sure that View added within View that consume the callback can triggered this one.

mvorisek: > Loader, Form, Console; they all use a callback. Loader can be nested. They can even be nested inside a VirtualPage.

but nested using View class...

> I do not see any problem carrying over argument. A callback is used to terminate output at some point. Before the regular app->output happen. Thus you want to make sure that View added within View that consume the callback can triggered this one.

URL of a View should never be affected by a component that is NOT within a sub render tree nor parents. This is important for debugging or for things like immutable state/diffing. So V2 URL must be rendered always the same no matter if rendered from App/V1 or (triggered/requested) C1.

To achive this, C1 sticky args must never be saved within V1. This usecase should be also covered by a test.

ibelar: > To achive this, C1 sticky args must never be saved within V1. This usecase should be also covered by a test.

Done! Also had use case for it.

src/View.php
```diff
@@ -584,6 +583,83 @@ public function removeAttr($property)
...
+    /**
+     * Build an URL which this view can use for call-backs. It should
+     * be guaranteed that requesting returned URL would at some point call
+     * $this->invokeInit().
+     *
+     * @param string|array $page URL as string or array with page name as first element and other GET arguments
+     *
+     * @return string
+     */
+    public function url($page = [], bool $includeRunningCallback = false)
```
mvorisek, line 614: Can the 2nd arg be an instance of the running Callback?

If yes, it will be more robust as we will not have to filter `$_GET` which can be spoofed.

src/Callback.php
mvorisek: Can the 2nd arg be an instance of the running Callback?

If yes, it will be more robust as we will not have to filter `$_GET` which can be spoofed. We can easily traverse the passed Callback to it's root/parents/App and collect all triggering names.

To be fair, I like `$this->getOwner()->stickyGet($this->urlTrigger);` more than this. What about reverting this impl. and commenting the new test for now out? I can look at a better impl. in the future then.

src/AbstractView.php
mvorisek: `$this->url($this->cb->getUrlArguments()),` I am fine with this, collecting trigger/sticky names is special case...

mvorisek (approved):

ibelar:
Why did you put back global stickyGet to the callback? V2 Url as discuss above will contains the callback arguments which we want to prevent right? Seem like the test is passing no matter what....not sure why but it need to be fixed.

My manual test shows that when Callback is run; V2 URL will contain the callback argument.

````
$app->catch_runaway_callbacks = false;

$b1 = Button::addTo($app)->set('trigger callback');

$v1 = \Atk4\Ui\View::addTo($app);
$cb = \Atk4\Ui\Callback::addTo($v1);

$cb->set(function () use ($v1, $cb) {
    $v3 = \Atk4\Ui\View::addTo($v1);
    $v3->set($cb->getUrl());
    \Atk4\Ui\View::addTo($v3)->set($v3->url());
});

$v2 = \Atk4\Ui\View::addTo($v1);
$v2->set('v2 url: ' . $v2->url());

$b1->link($cb->getUrl());
`````

When using include running callback V2 URL was correct.

mvorisek:
I think we should handle collecting the trigger names differently, eg. collect them when building the URL, eg. traversing up the render tree/parents.

The test passes, please fix and open a new draft PR. I can look at the better impl. then.

ibelar:
Can you revert to the state it was prior to your changes? I will fix the test then.

Envoyé de mon iPhone

mvorisek:
> 
> Can you revert to the state it was prior to your changes? I will fix the test then. Envoyé de mon iPhone
> […](#)
> Le 21 juill. 2021 à 19:07, Michael Voříšek ***@***.***> a écrit : ﻿ I think we should handle collecting the trigger names differently, eg. collect them when building the URL, eg. traversing up the render tree/parents. The test passes, please fix and open a new draft PR. I can look at the better impl. then. — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or unsubscribe.

here you go https://github.com/atk4/ui/pull/1651

PR #1649, opened 2021-07-14
Feature/Virtual Page Action Executor

ibelar:
Modal executors can be problematic sometimes on Mobile devices. This PR will added options to use a VirtualPage Executor instead which is more mobile-friendly.

In order to change executor type, simply register the new class for Model\UserAction
````
$factory = $app->getExecutorFactory();
$factory->registerTypeExecutor($factory::STEP_EXECUTOR, [VpExecutor::class]);
````

- Added a return button for returning to the base page;
- Added a step list information with current step highlight in the list;
- When action is executed (final step) it will return to base page;

Executor preview:

Executor in arguments step:
Screen Shot 2021-07-14 at 2 34 46 PM

Executor in fields step:
Screen Shot 2021-07-14 at 2 34 54 PM

Executor in preview step:
Screen Shot 2021-07-14 at 2 35 02 PM

### CODE REVIEW

File: src/CardDeck.php [hunk 1]
```diff
@@ -103,6 +103,7 @@ protected function init(): void
 
         if ($this->paginator !== false) {
             $this->addPaginator();
+            $this->issetApp() ? $this->getApp()->stickyGet($this->paginator->name) : $this->stickyGet($this->paginator->name);
```
mvorisek: this seems wrong to me, we do not want any app/global sticky

ibelar: Same principle as for Grid paginator. Since Paginator is added outside the grid or card deck, we need a way to remember which page a user has click prior to run the action and get back to that same page after the action finishes its execution.

mvorisek: That is correct reasoning, but incorrect implementation - Paginator must be somehow bound with Grid/Table.

ibelar: Agree with you on this one. But implementation changes are for another PR.

mvorisek: With `mergeStickyArgsFromChildView` (removed in https://github.com/atk4/ui/pull/1646) it was very easily possible. Now I think you need to code someting simillar, let's decise before https://github.com/atk4/ui/pull/1651

ibelar: This is fixed now. Did not realize it at first but the error was somewhere else.

src/CardDeck.php
```diff
@@ -103,6 +103,7 @@ protected function init(): void
 
         if ($this->paginator !== false) {
             $this->addPaginator();
+            $this->stickyGet($this->paginator->name);
```
mvorisek, line 106: if removed, does Behat testing fails?

mvorisek (approved):
I checked only the URL building, which seems ok now

src/CardDeck.php
ibelar: there is no behat test for Card deck with either VpExecutor or new PanelExecutor. This is because the default executor register in ExecutorFactory  is ModalExecutor. I can added tests but we need a way to be able to switch executor types dynamically. Any idea? Should we switch base on GET argument in init-app?

mvorisek: I meant if sticky args in Card Deck are tested.

> with either VpExecutor or new PanelExecutor...

I belive this does not need extra cross testing

PR #1650, opened 2021-07-15
Feature/Panel Action Executor

ibelar:
Another option instead of using Modal for executing Data\UserAction.

- Will play step within the Right panel.

Merge after #1646 

````
$factory = $app->getExecutorFactory();
$factory->registerTypeExecutor($factory::STEP_EXECUTOR, [PanelExecutor::class]);
````

Screen Shot 2021-07-15 at 5 03 17 PM

Screen Shot 2021-07-15 at 4 56 07 PM

### CODE REVIEW

**Comment by mvorisek** [comment]
conflicts needs to be resolved and should be tested with changes in https://github.com/atk4/ui/pull/1651 (tests should pass if all implemented correctly)

File: src/Layout.php [hunk 1]
```diff
@@ -6,11 +6,4 @@
 
 class Layout extends View
 {
-    /**
-     * Add a loadable View.
-     */
-    public function addRightPanel(Panel\Loadable $panel): Panel\Loadable
-    {
-        return $this->getOwner()->add($panel, 'RightPanel');
```
mvorisek, line 14: This was setting `RightPanel` `region` for template. 👍 not needing it anymore, can we remove `RightPanel` region from the templates then?

src/CardDeck.php
```diff
@@ -103,7 +103,7 @@ protected function init(): void
 
         if ($this->paginator !== false) {
             $this->addPaginator();
-            $this->stickyGet($this->paginator->name);
+            $this->issetApp() ? $this->getApp()->stickyGet($this->paginator->name) : $this->stickyGet($this->paginator->name);
```
mvorisek: Why do we need global sticky?

src/App.php
```diff
@@ -578,9 +578,9 @@ public function run()
             $this->html->template->dangerouslyAppendHtml('HEAD', $this->html->getJs());
             $this->is_rendering = false;
 
-            if (isset($_GET[Callback::URL_QUERY_TARGET]) && $this->catch_runaway_callbacks) {
+            if (isset($_GET['__atk_callback']) && $this->catch_runaway_callbacks) {
```
mvorisek: Is this wanted change?

src/App.php
```diff
@@ -238,14 +238,14 @@ static function (int $severity, string $msg, string $file, int $line): bool {
...
-     * Register a modal view.
-     * Fomantic-ui Modal are teleported in HTML template
+     * Register a portal view.
+     * Fomantic-ui Modal or atk Panel are teleported in HTML template
      * within specific location. This will keep track
-     * of modals when terminating app using json.
+     * of them when terminating app using json.
      */
-    public function registerModal(Modal $modal): void
+    public function registerPortals(View $portal): void
```
mvorisek: I would prefer `Modal|Panel\Right` phpdoc here to prevent others to use this wrongly.

js/src/services/panel.service.js
```diff
@@ -295,19 +308,24 @@ class PanelService {
...
-            that.closePanel(id);
+            this.closePanel(id);
         }, 250));
+    }
 
+    /**
+     * Add esc away closing event handler.
+     */
+    addEscAwayEvent(id) {
+        // const that = this;
```
mvorisek: remove this line :)

demos/data-action/action-setup.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
```
mvorisek: In https://github.com/atk4/ui/commit/ac4b2512bed644bd92f47cedb0886ba8bd127ccd I have moved this file to `jsactions2.php` to not need the phpunit exclude. I prefer it also when manually testing some demos from the folders, I do not have to think if this file is functional on it's own or not.

ibelar: my bad. I screw up during resolve conflict. Thanks.

src/App.php
ibelar: No, again resolve conflict screw up. I put it back. Thanks

src/App.php
ibelar: done!

js/src/services/panel.service.js
ibelar: done

src/CardDeck.php
ibelar: remove!

src/Layout.php
ibelar: Right! Same with Modals too. I remove both.

mvorisek (approved):

PR #1651, opened 2021-07-22
Improve callbacks url generation 2

mvorisek:
as per https://github.com/atk4/ui/pull/1646#issuecomment-884577372

replace https://github.com/atk4/ui/pull/1646/files#diff-489a31f422a2238af13855d2db35e0df62c99606b21cb5c678330f7f857ea90eR53
because of removed
https://github.com/atk4/ui/pull/1646/files#diff-1314d7908e6167b9fbe3d52c6f0c3cadf76637c75eca9c9c36d66106c579edd3L120
and
https://github.com/atk4/ui/pull/1646/files#diff-489a31f422a2238af13855d2db35e0df62c99606b21cb5c678330f7f857ea90eL148

The motivation is to make URL generation independed of the actual request.

Currently, for ex. CRUD generates some link to open a Modal. But when it is opened & saved, a different link is generated to open is again, it can be obseved on the `demos/_unit-test/crud.php` demo:

- 1st time http://localhost/dequ_ui_vendor/atk4/ui/demos/_unit-test/crud.php?__atk_cbtarget=_035c5ce8278dcf7e6c79c50288172d00__lexecutor_loader_callback&__atk_cb__035c5ce8278dcf7e6c79c50288172d00__lexecutor_loader_callback=callback&atk_layout_maestro_crud_useraction_modalexecutor=1&step=fields
- 2nd time http://localhost/dequ_ui_vendor/atk4/ui/demos/_unit-test/crud.php?atk_layout_maestro_crud_view_view_paginator=1&__atk_cbtarget=_035c5ce8278dcf7e6c79c50288172d00__lexecutor_loader_callback&__atk_cb__035c5ce8278dcf7e6c79c50288172d00__lexecutor_loader_callback=callback&atk_layout_maestro_crud_useraction_modalexecutor=1&step=fields

related with https://github.com/atk4/ui/issues/2070


ibelar:
Tks! I fixed test and also added a behat test for url

tests/CallbackTest.php
```diff
@@ -81,6 +81,9 @@ public function testViewUrlCallback(): void
         // simulate triggering
         $_GET[Callback::URL_QUERY_TRIGGER_PREFIX . $cb->name] = '1';
 
+        // Test for non sticky
+        $cb->setUrlTrigger($cb->name);
```
mvorisek: I belive this is the reason why the original test was passing - why do we need the Behat test at all?

We can even extend the unit test to set $_GET with target name which should add the sticky trigger to the URL.

ibelar: > We can even extend the unit test to set $_GET with target name which should add the sticky trigger to the URL.

Not sure about this. If you think it will do it, then go ahead.

I thought that if stickyGet is set by means of another method, then the URL will be wrong and the test will still pass.
That is why I added the Behat test for it.

ibelar:
I can update the PR using a static property for runningCallback. Then we can grab argument from it when need.

````

class Callback 
{
   
    protected static $runningCallbacks = [];

    public static function getRunningCallbackArgs(): array
    {
        $args = [];
        foreach (self::$runningCallbacks as $cb) {
            $args[$cb->urlTrigger] = $cb->getTriggeredValue();
        }

        return $args;
    }

    public function set($fx = null, $args = null)
    {
        if ($this->isTriggered() && $this->canTrigger()) {
            self::$runningCallbacks[] = $this;
            return $fx(...($args ?? []));
        }
    }
}
````


This way, we do not need to check for Get value and also do not need `public const URL_QUERY_TRIGGER_PREFIX`

@mvorisek What do you think?

**Comment by ibelar** [comment]
I have made the changes. Feel free to revert if you do not like it.

**Comment by mvorisek** [comment]
> I can update the PR using a static property for runningCallback. Then we can grab argument from it when need.

I belive running callbacks can be obtained quite easily by traversing the parents /wo the need to persist them in a static array, I will impl. that later.

File: src/View.php [hunk 2]
```diff
@@ -613,16 +613,38 @@ public function jsUrl($page = [])
...
-        return $this->getApp()->url($page, false, $this->_getStickyArgs());
+        return $this->getApp()->url($page, false, array_merge($this->getRunningCallbackArgs(false, is_array($page) ? $page : []), $this->getStickyArgs()));
+    }
+
+    protected function getRunningCallbackArgs(bool $isTerminated, array $page): array
+    {
+        $args = [];
+        foreach ($this->elements as $v) {
+            if ($v instanceof Callback) { // @phpstan-ignore-line
+                if (($page[Callback::URL_QUERY_TARGET] ?? null) === $v->getUrlTrigger()) {
```
mvorisek, line 453: @ibelar can you review and look at the Behat testing?

1. callback triggers are now based purely on render tree
2. the `CallbackTest::testViewUrlCallback` should cover the original Behat testing
3. Behat testing is failing, not sure exactly why, maybe some extra condition like in https://github.com/atk4/ui/commit/e3c299b36cc26bc30eef3360d3f0aa188c1a53dc#diff-c5bb4a733c5ddd3bb0744ef4c29a8fd1e0bd5999fcf06b7f3cfef4c3d1926ce3R624 is needed - but please make sure, point 1 stays true, eg. $_GET is never evaluated

src/VirtualPage.php
```diff
@@ -32,6 +32,12 @@ protected function init(): void
         parent::init();
 
         $this->cb = $this->add([Callback::class, 'urlTrigger' => $this->urlTrigger ?: $this->name]);
+        unset($this->{'urlTrigger'});
```
mvorisek: this is a very bad hack to dedup `urlTrigger` property stored at 2 places, but let's keep is for now, related with https://github.com/atk4/core/pull/245 , otherwise this property can not be set from seed

src/View.php
ibelar: It seems that the delete action callback (JsCallback trigger argument) is added to the add action callback URL.
That is why you get the exception. It tries to run (`JsCallbakc::set()`) the delete action first.

Screen Shot 2021-07-26 at 12 04 16 PM

mvorisek: any idea what is wrong with the currect algorithm?

Consider this render tree:
```
App -> V1 -> C1
          -> C2
          -> V2 -> C3
```

currently, when URL for C3 is rendered, both C1 and C2 URL triggers are added. Is that correct?

ibelar:
In theory, C3 should contains trigger for the other callbacks only if they where activated (Callback::set()). Maybe you could check for this condition before adding the trigger. 

Envoyé de mon iPhone

src/View.php
```diff
@@ -613,16 +613,38 @@ public function jsUrl($page = [])
...
+                    $isTerminated = true;
+                }
+
+                if ($isTerminated && $v->isTriggered() && $v->canTrigger()) {
```
mvorisek, line 457: @ibelar does `$v->canTrigger()` make sense here and any idea how to fix the remaining failing Behat tests?

ibelar: > @ibelar does `$v->canTrigger()` make sense here and any idea how to fix the remaining failing Behat tests?

Yes, canTrigger method will check for a reload. Normally, JsCallback should not run during a reload.

ibelar: I see why the test is failing. ViewTester reload is missing Callback arguments. Thus when reload is call, the callback method is not run. 

I think the reason is that ViewTester is added directly to Modal, while the callback is added to Button

`$button->on('click', $vp1Modal->show());`

Resulting in 
```
App -> Button -> Cb
    -> Modal -> ViewTester (reload)
```

Thus the callback is not part of the Modal render tree... missing its arguments when reload URL is rendered.

mvorisek: > Thus the callback is not part of the Modal render tree...

any idea how to solve it?

ibelar:
As a quick fix, one way is to collect Callback trigger argument in a static property. 

Other than that, I do not know.

Envoyé de mon iPhone

ibelar:
One other way might be to have a get argument method in JsReload that will scan all App elements for callbacks and collect their trigger argument if triggered.

mvorisek:
Why modal can not have normal owner from render tree? I know, it is then rendered differently, but with callbacks concept, there should be no difference on the programmer side.

ibelar:
> Why modal can not have normal owner from render tree?

I do not think it has to do with the render tree. The problem happens during the reload of a View.

Let's take this failing code:

```
$vp1Modal = \Atk4\Ui\Modal::addTo($app, ['title' => 'Lorem Ipsum load dynamically']);
$vp1Modal->set(function ($modal) use ($vp2Modal) {

    // Reload
    ViewTester::addTo($modal);

    \Atk4\Ui\View::addTo($modal, ['Showing lorem ipsum']); // need in behat test.
    \Atk4\Ui\LoremIpsum::addTo($modal, ['size' => 2]);
    $form = \Atk4\Ui\Form::addTo($modal);
    $form->addControl('color', null, ['enum' => ['red', 'green', 'blue'], 'default' => 'green']);
    $form->onSubmit(function (\Atk4\Ui\Form $form) use ($vp2Modal) {
        return $vp2Modal->show(['color' => $form->model->get('color')]);
    });
});

$button = \Atk4\Ui\Button::addTo($app)->set('Open Lorem Ipsum');
$button->on('click', $vp1Modal->show());
```

This result in this render tree:

```
App -> Button -> Callback
    -> Modal -> ViewTester (reload)
```

When reload URL is output; It is missing the callback URL arguments, because the callback is attached to the Button, not Modal.

Thus, when reload is execute (ViewTester) then the callback is never triggered again, which is wrong. 
We need a way to include all callback arguments for reloading URLs, when reload are added within Callback::set() method.

mvorisek:
Why the render tree can not look like?

```
App -> Button -> Callback -> Modal -> ViewTester (reload)
```

👍 for the code example. WDYT about throwing an exception if a render (sub)tree not part of current subtree of the callback one is modified? I think that should be not allowed to prevent such issues.

Or should we collect callbacks trace also from stack trace? (if a reload would be rendered from callback, it will work)

ibelar:
> WDYT about throwing an exception if a render (sub)tree not part of current subtree

I think that will be too limiting.

src/View.php
mvorisek: @ibelar the issue is here: https://github.com/atk4/ui/blob/8014b6c1cb5beb103f337af8ace5ac350f73ce19/src/JsReload.php#L58

1. in https://github.com/atk4/ui/pull/1646 you have removed `mergeStickyArgsFromChildView` which allowed to delegate sticky args from it's children.
2. in this PR, sticky args are collected from the render tree, but the URL must be obtained thru terminating callback
3. here, hovewer the URL is obtained thru `$this->view` which is placed correctly within the Modal render tree, but the Modal is not delegating it's callback sticky args as `mergeStickyArgsFromChildView` was removed

To better understand the issue, I have pushed a debug code in https://github.com/atk4/ui/pull/1651/commits/53a03522f528673618482463f1cf71c87a66fc81#diff-0741d95fd53e7a9a50ab8b371b0d621c3eaf3cb460c79dfe5a886deaec0814a9R68 , when you open the test URL, the expected and actual value must be the same.

I belive, `mergeStickyArgsFromChildView` is needed. Or how else should the `$this->view->jsUrl(...` in JsReload work?

With https://github.com/atk4/ui/pull/1651/commits/29ecd83ff8a956b001defb7b7988341da6cf95d1#diff-c5bb4a733c5ddd3bb0744ef4c29a8fd1e0bd5999fcf06b7f3cfef4c3d1926ce3R640-R668 Behat passes.

Here is a summary of problems to solve:

1. traverse all callbacks that CAN trigger (can be solved easily but by brute force by analysis whole reachable render tree) - temporary implemented by analysis stack trace
2. identify THE LAST TERMINATING callback - temporary implemented by pretending that all Modals can terminate - I need a help with this, as terminating callback must be identified from the render tree (not by the user requested terminating callback from URL, as it can differ)

mvorisek:
Probably impossible to satisfy all goals with the current design - new URL handling and multipass rendering needed.

PR #1652, opened 2021-07-23
Refactor Behat\Context methods

ibelar:
- re-organize code.

### DIFF
--- a/src/Behat/Context.php
+++ b/src/Behat/Context.php
@@ -535,7 +535,6 @@ class Context extends RawMinkContext implements BehatContext
      */
     public function toastDisplayShouldContainText(string $text): void
     {
-        // get toast
         $toast = $this->getElementInPage('.ui.toast-container');
         if (mb_strpos($this->getElementInElement($toast, '.content')->getText(), $text) === false) {
             throw new Exception('Cannot find text: "' . $text . '" in toast');
--- a/src/Behat/Context.php
+++ b/src/Behat/Context.php
@@ -120,6 +120,26 @@ class Context extends RawMinkContext implements BehatContext
         }
     }
 
+    protected function getElementInPage(string $selector, string $method = 'css'): NodeElement
+    {
+        $element = $this->getSession()->getPage()->find($method, $selector);
+        if ($element === null) {
+            throw new Exception('Could not get element in page using this selector: ' . $selector);
+        }
+
+        return $element;
+    }
+
+    protected function getElementInElement(NodeElement $element, string $selector, string $method = 'css'): NodeElement
+    {
+        $find = $element->find($method, $selector);
+        if ($find === null) {
+            throw new Exception('Could not get element in element using this selector: ' . $selector);
+        }
+
+        return $find;
+    }
+
     /**
      * Sleep for a certain time in ms.
      *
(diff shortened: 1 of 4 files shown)


src/Behat/Context.php
```diff
@@ -48,8 +48,8 @@ public function waitUntilLoadingAndAnimationFinished(AfterStepScope $event): voi
     protected function getFinishedScript(): string
     {
         return 'document.readyState === \'complete\''
-            . ' && typeof jQuery !== \'undefined\' && jQuery.active === 0'
-            . ' && typeof atk !== \'undefined\' && atk.vueService.areComponentsLoaded()';
+               . ' && typeof jQuery !== \'undefined\' && jQuery.active === 0'
```
mvorisek: 4 extra spaces should be enough

src/Behat/Context.php
```diff
@@ -629,35 +583,33 @@ private function assertSelectedValue(NodeElement $element, string $value, string
      */
     private function assertInputValue(NodeElement $element, string $value, string $selector = 'input'): void
     {
-        $input = $element->find('css', $selector);
-        if (!$input) {
-            throw new Exception('Input not found in selector: ' . $selector);
-        }
-        $inputValue = $input->getValue();
-        if ($inputValue !== $value) {
+        if ($this->assertElementInElement($element, $selector)->getValue() !== $value) {
```
mvorisek: `assertElementInElement` is getter, it should be named so then and made `protected` so it can be used when context is extended

src/Behat/Context.php
```diff
@@ -629,35 +583,33 @@ private function assertSelectedValue(NodeElement $element, string $value, string
...
-        if (!$rule) {
-            throw new Exception('Rule not found: ' . $ruleName);
+        return $this->assertElementInPage('.vqb-rule[data-name=' . $ruleName . ']');
+    }
+
+    private function assertElementInPage(string $selector, string $method = 'css'): NodeElement
+    {
+        $element = $this->getSession()->getPage()->find($method, $selector);
+        if ($element === null) {
+            throw new Exception('No element found with class name: ' . $selector);
```
mvorisek: here you call it "class name", in exception below "selector", if it can be more than class name, use "selector" keyword instead

src/Behat/Context.php
ibelar: Make sense. Done!

src/Behat/Context.php
ibelar: Done!

src/Behat/Context.php
ibelar: Done!

src/Behat/Context.php
```diff
@@ -48,8 +48,8 @@ public function waitUntilLoadingAndAnimationFinished(AfterStepScope $event): voi
     protected function getFinishedScript(): string
     {
         return 'document.readyState === \'complete\''
-            . ' && typeof jQuery !== \'undefined\' && jQuery.active === 0'
-            . ' && typeof atk !== \'undefined\' && atk.vueService.areComponentsLoaded()';
+           . ' && typeof jQuery !== \'undefined\' && jQuery.active === 0'
+           . ' && typeof atk !== \'undefined\' && atk.vueService.areComponentsLoaded()';
```
mvorisek:
```suggestion
            . ' && typeof jQuery !== \'undefined\' && jQuery.active === 0'
            . ' && typeof atk !== \'undefined\' && atk.vueService.areComponentsLoaded()';
```

src/Behat/Context.php
```diff
@@ -536,73 +471,117 @@ public function scopeBuilderBoolRule(string $name, string $value): void
...
+    /**
+     * @Then Toast display should contains text :arg1
+     */
+    public function toastDisplayShouldContainText(string $text): void
     {
-        $expected = $this->getSession()->getPage()->find('css', $selector)->getText();
-        $input = $this->getSession()->getPage()->find('css', 'input[name="' . $inputName . '"]');
-        if (!$input) {
-            throw new Exception('Unable to find input name: ' . $inputName);
+        // get toast
```
mvorisek:
```suggestion
```

mvorisek (approved):
👍 less LOC, very good! :)

PR #1655, opened 2021-08-06
Fix click event bubbling in Table::onRowClick()

### LINKED ISSUE #1601: Table->onRowClick() - Ignore click on checkboxes
`Table->onRowClick()` should not execute its action when a checkbox gets checked or unchecked.

ibelar:
fix #1601


src/Table.php
```diff
@@ -594,12 +594,12 @@ public function renderRow()
...
      * @return Jquery
      */
-    public function onRowClick($action)
+    public function onRowClick($action, array $excludeSelector = [])
     {
         $this->addClass('selectable');
         $this->js(true)->find('tbody')->css('cursor', 'pointer');
 
-        return $this->on('click', 'tbody>tr', $action);
+        return $this->on('click', 'tbody tr td' . ($excludeSelector ? ':not(' . implode(',', $excludeSelector) . ')' : ''), $action);
```
mvorisek: why not simply exclude a checkbox control?

ibelar: More flexible this way. You may decide to exclude any table cell you need.

mvorisek (changes requested):
If I understand the issue, we need to prevent event bublibling from controls in table cells. This does not solve it, at least not automatically. Then I belive we do not even need the option to pass any selector to exlude (and for rare cases, the user can do it on it's own as the method is simple).

src/Table.php
mvorisek: adding `Checkbox::CHECKBOX_COLUMN_CSS` constant seems not needed for me, but if we absolutely need to use exclusing selector, I am ok with it

mkrecek234:
A thought to get this finalised: Shouldn't we add a class to all links, checkboxes or other clickable items that are added to a table row, just like you did it for `Checkbox::CHECKBOX_COLUMN_CSS` like `atk4-table-column-clickable`. This class we set automatically for classes in Atk4\Ui\Table\Column like you did for Checkboxes, but also for Link. We can then define  `function onRowClick($action, array $excludeSelector = ['atk4-table-column-clickable])`. So developers have all flexibility. 

Whatever you guys consider best to get this released ;-) Please comment and I can adjust the code.

mkrecek234:
We should do a generalisation by a defined class, here atk4-norowclick.
The selector td:not(atk4-norowclick) does not work, it does not prevent the action be triggered as you can see in the grid.php demo.

mkrecek234:
Fixed, working now as specified - please check demo and revert demo changes in grid.php before merge

demos/collection/grid.php
```diff
@@ -69,5 +69,12 @@
     [$sel->jsChecked()]
 ));
 
+$grid->addDecorator($model->titleField, [\Atk4\Ui\Table\Column\Link::class, 'url' => 'https://hmpg.net']);
```
mvorisek:
```suggestion
$grid->addDecorator($model->titleField, [Table\Column\Link::class, 'url' => 'https://hmpg.net']);
```

demos/collection/grid.php
```diff
@@ -69,5 +69,12 @@
...
+
+
+$grid->table->onRowClick(function() {
```
mvorisek: needs test

also test the checkbox, as with link, it is hard to test if the event was propagated or not when navigated (URL changed)

src/Table.php
```diff
@@ -502,13 +502,20 @@ public function renderRow(): void
...
      */
     public function onRowClick($action): void
     {
         $this->addClass('selectable');
         $this->js(true)->find('tbody')->css('cursor', 'pointer');
 
+        /* Do not trigger row click event if click stems from row content
+         like checkboxes, or links, marked as atk4-norowclick
+         */
+        $this->js(true )->find('.atk4-norowclick')->on('click',
```
mvorisek: well, why do we need a selector then :) (in chat, you have written `:not(...)` class is not working)

mkrecek234: I do not understand - we need to stopPropagation for Link, or Checkbox, so we flag them with the `atk4-norowclick` class.

mvorisek: component level (checkbox, link, ...) is responsible to prevent such events, why `event.stopPropagation` is then setup from a different component?

I see the exclusions are setup `src/Table/Column/*` and this kinda makes sense then...

but this is again a Table/Grid/Crud specific fix and CardCard sail out of sync again :)

let's start with checkbox click test

then I am fine to merge the current fix as for Table components it LGTM

src/Table.php
```diff
@@ -509,6 +509,14 @@ public function onRowClick($action): void
         $this->addClass('selectable');
         $this->js(true)->find('tbody')->css('cursor', 'pointer');
 
+        /* Do not trigger row click event if click stems from row content
+         like checkboxes, or links, marked as atk4-norowclick
+         */
+        $this->js(true)->find('.atk4-norowclick')->on(
```
mvorisek: after test is added, before merge, consider to just exclude `a` and `checkbox` tags instead

mkrecek234: I do not think it is a good idea: The developer might be willing to add clickable labels, or other items into the columns, that are clickable - by adding atk4-norowclick he can easily ensure the correct behaviour for those as well.

mvorisek: yes, but is there a case when the exclusion is not wanted for `a`? If not, we only generate too complex component code as the tests shown...

mkrecek234: A label can sometimes be configured to have an on('click') event to trigger something.

mvorisek: maybe `a, checkbox, .atk4-norowclick`

src/Table.php
```diff
@@ -509,6 +509,14 @@ public function onRowClick($action): void
         $this->addClass('selectable');
         $this->js(true)->find('tbody')->css('cursor', 'pointer');
 
+        /* Do not trigger row click event if click stems from row content
+         like checkboxes, or links, marked as atk4-norowclick
+         */
+        $this->js(true)->find('a, .checkbox, .atk4-norowclick')->on(
+            'click',
+            new JsExpression('function(event) { event.stopPropagation() }')
```
mvorisek: the comment does just dummily tells now what the code does/says itself - remove

`new JsExpression('function(event)` I think `JsFunction` can simplify it

also check whole PR diff/changes, `src/Table/Column/Checkbox.php` has some unreverted changes

mvorisek:
@mkrecek234 PR looks very well made!

I just need to write the test.

Also, the coverage has decreased hugely. I am already on it. The problem is present since https://github.com/atk4/ui/pull/1840. It seems there is some bug with https://github.com/sebastianbergmann/php-code-coverage itself when caching is enabled.

mvorisek:
Thank you @ibelar and @mkrecek234!

PR #1657, opened 2021-09-08
Fix Stan

mvorisek:
I belive this is an issue in phpstan: https://github.com/phpstan/phpstan/issues/5607, do NOT merge before the issue is cleared.


mvorisek:
@ibelar let's merge this, not sure when this will be fixed in phpstan, we can revert anytime...

ibelar (approved):

PR #1658, opened 2021-09-13
Fix Multiline Model validate

ibelar:
- display error when using model validate().

Waiting for #1657


mvorisek:
LGTM, phpstan errors seems unrealted

mvorisek (approved):
LGTM, phpstan errors seems unrelated

PR #1659, opened 2021-09-14
Allow to initialize CascadeDropdown values

ibelar:
Two values for proper initialization of the DropdownCascade is need: The actual value of the dropdown and the reference value for populating the dropdown items.

Usage:

```
$form = Form::addTo($app);

$form->addControl('cat', [Form\Control\Dropdown::class, 'model' => new Category($app->db)])->set(2);

$form->addControl('sub', [Form\Control\DropdownCascade::class, 'cascadeFrom' => 'cat', 'reference' => Category::hinting()->fieldName()->SubCategories])->set(9);

$form->addControl('prod', [Form\Control\DropdownCascade::class, 'cascadeFrom' => 'sub', 'reference' => SubCategory::hinting()->fieldName()->Products])->set(3)
```

Screen Shot 2021-09-14 at 3 10 11 PM

### DIFF
--- a/src/Form/Control/DropdownCascade.php
+++ b/src/Form/Control/DropdownCascade.php
@@ -18,14 +18,14 @@ class DropdownCascade extends Dropdown
     /** @var string|Form\Control|null the form input to use for setting this dropdown list values from. */
     public $cascadeFrom;
 
-    /** @var string|Model|null the hasMany reference model that will generated value for this dropdown list. */
+    /** @var string|Model|null the hasMany reference model that will generate value for this dropdown list. */
     public $reference;
 
     /** @var Form\Control The form control object created based on cascadeFrom */
-    protected $cascadeControl;
+    protected $cascadeFromControl;
 
-    /** @var string The casacade input value. */
-    protected $cascadeControlValue;
+    /** @var string|int The cascade input value. */
+    protected $cascadeFromControlValue;
 
     protected function init(): void
     {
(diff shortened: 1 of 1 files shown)


src/Form/Control/DropdownCascade.php
```diff
@@ -35,46 +35,61 @@ protected function init(): void
...
+        $this->cascadeFromControl->onChange($expr, ['args' => [$this->cascadeFromControl->name => $this->cascadeFromControl->jsInput()->val()]]);
+    }
+
+    /**
+     * Allow initializing CascadeDropdown with preset value.
+     *
+     * @param string|int $initialValue       The initial ID value to set this dropdown using reference model values
+     * @param string|int $fromReferenceValue The Cascade from reference ID value where reference model values
+     */
+    public function setInitialValues($initialValue, $fromReferenceValue): self
```
mvorisek: why is the `$fromReferenceValue` needed?

mvorisek:
If the `'category_id'` and `'sub_category_id'` is from some demo/doc, replace it with `SubCategory::hinting()->fieldName()->...`

Also check the default values in Behat test.

src/Form/Control/DropdownCascade.php
```diff
@@ -35,46 +35,61 @@ protected function init(): void
...
      */
-    public function getNewValues(string $id): array
+    public function getNewValues($id): array
     {
         if (!$id) {
             return [['value' => '', 'text' => $this->empty, 'name' => $this->empty]];
         }
 
-        $model = $this->cascadeControl->model->load($id)->ref($this->reference);
+        $model = $this->cascadeFromControl->model->tryLoad($id)->ref($this->reference);
```
mvorisek, line 93: why `tryLoad`?

src/Form/Control/DropdownCascade.php
ibelar: $fromReferenceValue is the model Id need for setting up the reference model in order to build the dropdown menu items.

src/Form/Control/DropdownCascade.php
ibelar: No need to throw an exception - showing an empty dropdown is enough if the model is not loaded.

src/Form/Control/DropdownCascade.php
mvorisek: But why this field is not available from `cascadeFromControl`?

ibelar: > But why this field is not available from `cascadeFromControl`?

make sense... I refactor using set()

mvorisek (approved):

PR #1664, opened 2021-09-29
Adjust to use DBAL type

mvorisek:
for https://github.com/atk4/data/pull/727 change

marking as bug & BC-break as DateTime handling was buggy and not only managed values should be set in PHP

Also improve:
- Callback exception are caught in order to suppress ultra-hard-to-debug `Callback requested, but never reached` errors
- Behat testing, exceptions are now parsed from HTML and dumped

callback exceptions before:

image

callback exceptions after:

image

### DIFF
--- a/demos/_includes/DemoActionsUtil.php
+++ b/demos/_includes/DemoActionsUtil.php
@@ -131,7 +131,7 @@ class DemoActionsUtil
                 'args' => [
                     'age' => ['type' => 'integer', 'required' => true],
                     'city' => [],
-                    'gender' => ['type' => 'enum', 'values' => ['m' => 'Male', 'f' => 'Female'], 'required' => true, 'default' => 'm'],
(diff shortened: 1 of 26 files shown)


DarkSide666 (approved):
pre-approved

demos/_includes/DemoActionsUtil.php
```diff
@@ -131,7 +131,7 @@ public static function setupDemoActions(CountryLock $country): void
                 'args' => [
                     'age' => ['type' => 'integer', 'required' => true],
                     'city' => [],
-                    'gender' => ['type' => 'enum', 'values' => ['m' => 'Male', 'f' => 'Female'], 'required' => true, 'default' => 'm'],
+                    'gender' => ['type' => 'string', /*'values' => ['m' => 'Male', 'f' => 'Female'],*/ 'required' => true, 'default' => 'm'],
```
mvorisek, line 134: @ibelar any idea how to fix this?

PR #1667, opened 2021-10-02
Fix UI typecasting

mvorisek:
merge after https://github.com/atk4/core/pull/329 and https://github.com/atk4/data/pull/897


src/Form/Control.php
```diff
@@ -100,7 +100,6 @@ protected function init(): void
     public function set($value = null, $junk = null)
     {
         if ($this->field) {
-            $value = $this->getApp()->ui_persistence->typecastLoadField($this->field, $value);
```
mvorisek, line 103: @ibelar any idea why this was there?

mvorisek, line 103: This was wrong here. `Control::set` should expect already fully loaded data and can not be used to set data from frontend directly.

tests-behat/scopebuilder.feature
```diff
@@ -12,7 +12,8 @@ Feature: ScopeBuilder
     Then bool rule "atk_fp_stat__is_commercial" has value "No"
     Then I check if input value for "qb" match text in "p.atk-expected-input-result"
     And I press button "Save"
-    Then I check if text in "p.atk-expected-word-result" match text in ".atk-scope-builder-response"
+    # TODO uncomment once "Object serialization is not supported" is fixed
+    # Then I check if text in "p.atk-expected-word-result" match text in ".atk-scope-builder-response"
```
mvorisek, line 16: @georgehristov disabling now as we can not serialize for security reasons from the user input. Why do we need to serialize (object type) at all?

ibelar (approved):
Look Good. - There are some issues in demos for calendar input and multiline but it might be because of some refactoring need after #1664 
I think it is ok to merge, we can submit pr for each issues.

mvorisek:
Thanks for the review. The issue is here https://github.com/atk4/ui/issues/1668. If you know about any other issue, please post.

PR #1671, opened 2021-10-05
Fix and test nothing is GET sticky globally

mvorisek:
No global GET sticky is important design goal of component based atk4/ui and we must fix it and assert it across all tests.

The presented test should reject usages like in https://github.com/atk4/ui/pull/1650#discussion_r721775867

Also removing `tools/get-assets.php` as it is dangerous to cache resources as they might become out-of-sync and for some files, such dummy caching is not enought, as the resources might access another resources using relative paths.


src/Grid.php
```diff
@@ -342,7 +342,7 @@ public function addQuickSearch($fields = [], $hasAutoQuery = false)
         $view = View::addTo($this->menu
             ->addMenuRight()->addItem()->setElement('div'));
 
-        $q = trim($this->stickyGet('_q') ?? '');
+        $q = trim($this->stickyGet($this->name . '_q') ?? '');
```
mvorisek: @ibelar please help with adjust of `js/src/plugins/js-search.plugin.js` file to the GET key is prefixed with the component name, the tests should be passing then

mvorisek: @ibelar can you please help adjusting the JS code to the new prefixed name?

mvorisek: ibelar seems very budy, @abbadon1334 do you have some JS knowledge and can you help adjusting the JS?

abbadon1334: I try to help, but I'm not an expert on JS, but I think the only change needed is this :
```js
    /**
   * Send request to server using the search query.
   *
   * @param uri
   * @param options
   */
    doSearch(uri, query, options, cb = function () {}) {
        const elKey = this.$el.id + '_q';
        if (query) {
            const extOpts = {};
            extOpts[elKey] = query;
            options = $.extend(options, extOpts);
        }

        if (this.settings.useAjax) {
            this.$el.api({
                on: 'now',
                url: uri,
                data: options,
                method: 'GET',
                obj: this.$el,
                stateContext: this.searchAction,
                onComplete: cb,
            });
        } else {
            uri = $.atkRemoveParam(uri, elKey);
            if (options.__atk_reload) {
                delete options.__atk_reload;
            }
            uri = $.atkAddParams(uri, options);
            window.location = uri;
        }
    }
```
I never touch a single line of Atk js, I cannot test it, because on `npm run dev` i get this error `ERR_OSSL_EVP_UNSUPPORTED`

mvorisek: Thanks for the help, I added a settings param in php, but your post pushed me into the right direction 👍

mvorisek: There are still 2 issues:

a) `stickyGet` on l114 in `src/Grid.php` is still need, Behat tests pass then (filter is kept after row is edit)
b) when page is reload, but filter was applied previously (eg. search box is filled with some value), the filter is not applied, I belive this was also the case before this PR

PR #1674, opened 2021-10-06
Add Behat coverage

mvorisek:
https://docs.codecov.com/docs/merging-reports multiple coverage reports are supported, let's add Behat coverage too

image


src/Table.php
```diff
@@ -170,7 +170,7 @@ public function initChunks()
     public function addColumn(?string $name, $columnDecorator = null, $field = null)
     {
         if (!$this->_initialized) {
-            throw new Exception\NoRenderTree($this, 'addColumn()');
```
mvorisek, line 173: the exception was used only here, we do not want to make this exception available as the reason to use it is not good

DarkSide666 (approved):

PR #1683, opened 2021-10-19
Fix DB config for CI webserver started with one env

mvorisek:
previously, Behat was tested with Sqlite only

merge after https://github.com/atk4/data/pull/908


DarkSide666 (approved):
Behat test fails

mvorisek:
👍 @DarkSide666 for noticing the CI error, I belive the PgSQL Behat is failing really because of some DB specific issue.

PR #1687, opened 2021-11-04
UserAction StepExecutor improvement

### LINKED ISSUE #1680: Step Executor should better allow overriding each step method.
Current implementation does not allow for easily overriding each step method: 
- StepExecutorTrait::doArgs(),
- StepExecutorTrait::doFields(),
- StepExecutorTrait::doPreview(),

ibelar:
Fix atk4/ui#1680 

- Refactor in order to improve Class extension;
- Allow settings Data\UserAction argument from a Data\Model
  - arguments can be validated using Model validation.


src/UserAction/StepExecutorTrait.php
```diff
@@ -102,18 +105,55 @@ protected function setFormField(Form $form, array $fields, string $step): Form
...
         return $form;
     }
 
+    /**
+     * Set model for userAction arguments.
+     * Override existing argument with model definition.
+     */
+    protected function getModelForArgs(): Model
+    {
+        $this->cloneAction = clone $this->getAction();
```
mvorisek: why do we need to clone, what is the purpose of using cloned&modified action internally?

this can lead also to many bugs when the action is semi-shallowly cloned, like when it has Closures bound still to the original action

ibelar: Cloning was to duplicate Action arguments when set by model only, i.e. each editable model field need to be added as an argument to the UserAction. However, I realized I do not need to clone UserAction but only it's arguments.

A fix is on the way.

src/UserAction/StepExecutorTrait.php
```diff
@@ -188,11 +215,11 @@ protected function doFields(View $page): void
 
         if (!$form->hookHasCallbacks(Form::HOOK_SUBMIT)) {
             $form->onSubmit(function (Form $form) {
-                // collect fields.
-                $form_fields = $form->model->get();
-                foreach ($this->action->fields as $field) {
-                    $this->actionData['fields'][$field] = $form_fields[$field];
-                }
+                // collect fields define in Model\UserAction.
+                $fields = array_filter($form->model->get(), function ($field) {
```
mvorisek: `array_intersect_key($form->model->get(), array_flip($this->action->fields))`

but I think foreach or array_map(function ($k) { ..->get($k) }, $this->action->fields) is better as it fetches only the needed data

src/UserAction/StepExecutorTrait.php
```diff
@@ -102,18 +105,53 @@ protected function setFormField(Form $form, array $fields, string $step): Form
...
+                $argsModel->invokeInit();
+            }
+
+            unset($args['__atk_model']);
+        } else {
+            $argsModel = new Model(new Array_([]));
+        }
+
+        foreach ($args as $key => $val) {
+            $argsModel->addField($key, $val);
```
mvorisek, line 126: why args are not a Model?

ibelar: UserAction->args are not model, they never was. Only now executor will use them within a model.

src/UserAction/StepExecutorTrait.php
ibelar: what is wrong with array_filter?

mvorisek: a) no need to fn if built in function like `array_intersect_key` does the same
b) we should not materialize all fields and filter afterwards, we should materialize only the needed ones (as some might me lazy loaded in the future)

src/UserAction/StepExecutorTrait.php
```diff
@@ -166,8 +189,12 @@ protected function doArgs(View $page): void
         $this->jsSetPrevHandler($page, $this->step);
 
         $form->onSubmit(function (Form $form) {
+            $form->model->save();
             // collect arguments.
-            $this->actionData['args'] = $form->model->get();
+            foreach ($form->model->getFields('editable') as $k => $field) {
+                $argValues[$k] = $form->model->get($k);
+            }
+            $this->setActionData('args', $argValues ?? []);
```
mvorisek:
```suggestion
            $argValues = [];
            foreach ($form->model->getFields('editable') as $k => $field) {
                $argValues[$k] = $form->model->get($k);
            }
            $this->setActionData('args', $argValues);
```

never use `??` for possibly undefined local variables, use it only for possibly undefined keys, it is otherwise hard to read and errors can be easily introduced (for example when there is typo in the variable name)

src/UserAction/StepExecutorTrait.php
```diff
@@ -102,18 +105,53 @@ protected function setFormField(Form $form, array $fields, string $step): Form
...
+        } else {
+            $argsModel = new Model(new Array_([]));
+        }
+
+        foreach ($args as $key => $val) {
+            $argsModel->addField($key, $val);
+        }
+
+        // set userAction args using model field.
+        foreach ($argsModel->getFields('editable') as $k => $field) {
```
mvorisek, line 131:
```suggestion
        $this->cloneArgs = [];
        foreach ($argsModel->getFields('editable') as $k => $field) {
```
and no default value in the property declaration

ibelar: see https://github.com/atk4/ui/pull/1687/files#diff-bbd75fe6645c71d2c9892c71761a2044e9aa673eeb41f5f5f4a2abd26ba677d1R76

property $cloneArgs is already define as an empty array

src/UserAction/StepExecutorTrait.php
```diff
@@ -102,18 +105,53 @@ protected function setFormField(Form $form, array $fields, string $step): Form
...
+     */
+    protected function getModelForArgs(): Model
+    {
+        $args = $this->getAction()->args;
+        if (array_key_exists('__atk_model', $args)) {
+            /** @var Model $argsModel */
+            $argsModel = Factory::factory($args['__atk_model']);
+            // if seed is supplied, we need to initialize.
+            if (!$argsModel->_initialized) {
+                $argsModel->invokeInit();
```
mvorisek, line 117: Model::init is invoked when persistence is set

when not initialized, I belive we should set `new Array_([])` persistence instead

side question, when custom model for args does bring any advantage?

src/UserAction/StepExecutorTrait.php
mvorisek: but is it correct, that if this method is NOT called, `[]` default value is valid?

if not, `null` should be the default property value

src/UserAction/StepExecutorTrait.php
ibelar: update with array_intersect

src/UserAction/StepExecutorTrait.php
ibelar: - argument can be easily validate using a Validator;
- argument can be easily define within a model;

demos/_includes/DemoActionsUtil.php
```diff
@@ -131,7 +155,37 @@ public static function setupDemoActions(CountryLock $country): void
...
+            ]
+        );
+
+        $country->addUserAction(
+            'arg_using_model',
+            [
+                'caption' => 'Arg with Model',
+                'description' => 'Ask for Arguments set via a Data\Model. Allow usage of model validate() for your arguments',
+                'args' => [
+                    '__atk_model' => new ArgModel(new Array_([])),
```
mvorisek: `[ArgModel::class]` should be probably here

demos/_includes/DemoActionsUtil.php
```diff
@@ -131,7 +155,37 @@ public static function setupDemoActions(CountryLock $country): void
...
+                        ],
+                    ],
+                ],
+                'fields' => [$country->fieldName()->iso3],
+                'callback' => function ($model, $age, $city, $gender) {
+                    $n = $gender === 'm' ? 'Mr.' : 'Mrs.';
+
+                    return 'Thank you ' . $n . ' at age ' . $age;
+                },
+                'preview' => function ($model, $age, $city, $gender) {
```
mvorisek: why `$age`, ... params if model is passed as well?

I prefer to not pass the args as a) it is declaration order dependent I think b) it is not refactorable not testable with phpstan

same for `callback`

demos/_includes/DemoActionsUtil.php
```diff
@@ -131,7 +155,37 @@ public static function setupDemoActions(CountryLock $country): void
...
+            'arg_using_model',
+            [
+                'caption' => 'Arg with Model',
+                'description' => 'Ask for Arguments set via a Data\Model. Allow usage of model validate() for your arguments',
+                'args' => [
+                    '__atk_model' => new ArgModel(new Array_([])),
+                    'extra' => ['type' => 'string'],
                 ],
                 'fields' => [$country->fieldName()->iso3],
                 'callback' => function ($model, $age, $city, $gender) {
```
mvorisek: then define hintable phpdoc above `ArgModel` class and get the values here with magic props

with implicitly declared model (UA above, use ->get($country->fieldName()->age))

demos/_includes/DemoActionsUtil.php
```diff
@@ -4,7 +4,31 @@
...
+        parent::init();
+        $this->addField('age', ['type' => 'integer', 'required' => true, 'caption' => 'Age must be 21 or over:']);
+        $this->addField('city');
+        $this->addField('gender', ['default' => 'm', 'ui' => ['form' => [Dropdown::class, 'values' => ['m' => 'Male', 'f' => 'Female']]]]);
+    }
+
+    public function validate(string $intent = null): array
+    {
+        $error = [];
+        if ($this->get('age') < 21) {
```
mvorisek: use hintable for `'age'` here too

src/UserAction/StepExecutorTrait.php
ibelar: [] is needed as the cloneArg are collected to execute the action.
see: https://github.com/atk4/ui/pull/1687/files#diff-bbd75fe6645c71d2c9892c71761a2044e9aa673eeb41f5f5f4a2abd26ba677d1R527

mvorisek: that is exactly my point :)

is `[]` valid there? it seems UNINITIALIZED to me, thus not valid

src/UserAction/StepExecutorTrait.php
ibelar: > when not initialized, I belive we should set `new Array_([])` persistence instead

If using seed then developper should supply proper property within the seed array, including Persistence.

demos/_includes/DemoActionsUtil.php
ibelar: It support both.

mvorisek: yes, more like recommended usage ;-)

demos/_includes/DemoActionsUtil.php
ibelar: $model here is the UserAction model and not the argument model. The arguments are send separately to preview callback.

demos/_includes/DemoActionsUtil.php
ibelar: Same here, $model is UserAction model not the argument model. Arguments are passed separately to the execute callback function.

src/UserAction/StepExecutorTrait.php
ibelar: done

src/UserAction/StepExecutorTrait.php
ibelar: Why do you say it is not initialized? It is when property is declare.

demos/_includes/DemoActionsUtil.php
ibelar: done

src/UserAction/StepExecutorTrait.php
mvorisek: I mean from app perspective, not language perspective. From brief look, it seems to be args are either set using `getModelForArgs` method or UNINITIALIZED - and the property value must be `null` not `[]`.

demos/_includes/DemoActionsUtil.php
mvorisek: but with the same field names as args, right?

ibelar: No - in this case here $model is the Country model. 
I just push an update with proper type to each callback function. It will help clear the confusion. :)

src/UserAction/StepExecutorTrait.php
ibelar: not necessary - if UserAction does not contains arguments - like an Edit action for example - the $cloneArgs is not initialized.
But call to get $cloneArgs will always be call to execute the action. This is what happen if I set it to null

Screen Shot 2021-11-05 at 10 22 03 AM

Therefore using an empty array as default value is needed.

ibelar: refactor method name and cloneArg  initialisation

mvorisek: > not necessary - if UserAction does not contains arguments

thanks, with `$this->cloneArgs = [];` locally is right now :)

ibelar:
Question:
Will you be ok to add a property to Model\UserAction say `$argModel ` instead of using a special key name '__atk_model' in args array?

Then executor can check for both UserAction property ($arg and $argModel) accordingly.

mvorisek:
> Question: Will you be ok to add a property to Model\UserAction say `$argModel ` instead of using a special key name '__atk_model' in args array?
> 
> Then executor can check for both UserAction property ($arg and $argModel) accordingly.

~If `Model\UserAction::$args` can be replaced with a Model, I would prefer that as otherwise args are passed like `...$args` which has no typing check neither correct arg names/order/count.~

I think lets use/stay with `__atk_model` for now as the args are not refactorable currently anyway. Maybe add it as a const to `Model\UserAction` class prefixed with the cl. name as other constants.

ibelar:
I am currently checking  passing the entire $argModel to callback function as well instead of passing $args separately.
This will allow full usage of Model capability for action argument i.e usage of hook within the model.

mvorisek:
we should probably add a special data persistence that prevent models to be saveable

composer.json
```diff
@@ -54,7 +54,7 @@
     ],
     "require": {
         "php": ">=7.4 <8.2",
-        "atk4/data": "dev-develop"
+        "atk4/data": "dev-norm_id_array_pers as dev-develop"
```
mvorisek, line 57: TODO remove before merge

@ibelar ui is failing with this data PR, can you please fix it here? It seems that a field name is set as a field value somewhere and not passing numeric validation.

mvorisek:
@ibelar how does this PR fix https://github.com/atk4/ui/issues/1680 and what do actaully mean by

> Current implementation does not allow for easily overriding each step method

as described in the issue desc?

src/UserAction/StepExecutorTrait.php
```diff
@@ -167,8 +190,9 @@ protected function doArgs(View $page): void
         $this->jsSetPrevHandler($page, $this->step);
 
         $form->onSubmit(function (Form $form) {
+            $form->model->save();
```
mvorisek: why is now save needed?

src/UserAction/StepExecutorTrait.php
```diff
@@ -72,6 +72,9 @@ trait StepExecutorTrait
     /** @var string */
     public $finalMsg = 'Complete!';
 
+    /** @var array An extended copy of UserAction arguments. It contains original action arguments and arguments set by '__atk_model'. */
+    private $cloneArgs;
```
mvorisek: I think we should add native Model args support to atk4/data model. It is also not completely clear to me, why there are args & fields. In atk4/data, the UserAction::$fields property is described to be used for dirty detection only.

demos/_includes/DemoActionsUtil.php
```diff
@@ -6,9 +6,40 @@
...
+use Atk4\Data\Persistence\Array_;
 use Atk4\Ui\Exception;
 use Atk4\Ui\Form;
 
+/**
+ *  @property string $age    @Atk4\Field()
+ *  @property string $city   @Atk4\Field()
+ *  @property string $gender @Atk4\Field()
+ */
+class ArgModel extends Model
```
mvorisek, line 18: our testing/demos Model should be used here to test if field aliasing is used properly

src/UserAction/StepExecutorTrait.php
```diff
@@ -124,19 +163,13 @@ protected function runSteps(): void
...
-        foreach ($this->action->args as $key => $val) {
-            if (isset($val['model'])) {
-                $val['model'] = Factory::factory($val['model']);
-                $form->addControl($key, [Form\Control\Lookup::class])->setModel($val['model']);
-            } else {
-                $form->addControl($key, [], $val);
-            }
-        }
+
+        $form->setModel($model->createEntity());
```
mvorisek, line 172: 👍

mvorisek:
I like the idea, however this should be addressed mainly in atk4/data. Then adjusting atk4/ui is an easy job.

As it seems you do not plan to work on this topic in the near future, I am closing this PR. Thank you for understanding.

PR #1692, opened 2021-11-06
Adjust to non-entity Field only

mvorisek:
related with https://github.com/atk4/data/pull/912 changes

## BC break

a) `Form\Control::$field` renamed to `$entityField` with new `Atk4\Data\Model\EntityFieldPair` type
b) some atk4/ui methods were simplified (for example to not accept vague string seed), use phpstan with level 6+ on your project to detect the needed changes


src/Form.php
```diff
@@ -525,20 +514,20 @@ public function controlFactory(\Atk4\Data\Field $field, $seed = [])
...
+            $ControlSeed,
             $field->ui['form'] ?? null,
             $this->typeToControl[$field->type] ?? null,
             $fallbackSeed
         );
 
         $defaults = [
             'form' => $this,
-            'field' => $field,
+            'field' => new EntityFieldProxy($entity, $field->short_name),
```
mvorisek, line 526: FYI MAJOR change

composer.json
```diff
@@ -54,7 +54,7 @@
     ],
     "require": {
         "php": ">=7.4 <8.2",
-        "atk4/data": "dev-develop"
+        "atk4/data": "dev-stricter_model_vs_entity as dev-develop"
```
mvorisek, line 57: REMOVE this before merge

DarkSide666 (approved):
OK for 3.1

PR #1695, opened 2021-11-09
Allow Overriding table on grid

ibelar:
Add Grid::initTable() method.

### DIFF
--- a/src/Grid.php
+++ b/src/Grid.php
@@ -144,7 +144,7 @@ class Grid extends View
 
     protected function initTable(): Table
     {
-        /** @var Table $table */
+        /** @var Table */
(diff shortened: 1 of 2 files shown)


src/Grid.php
```diff
@@ -142,6 +142,14 @@ protected function init(): void
         $this->issetApp() ? $this->getApp()->stickyGet('_q') : $this->stickyGet('_q');
     }
 
+    protected function initTable(): Table
+    {
+        /** @var Table $table */
```
```suggestion
        /** @var Table */
```

ibelar: ok

mvorisek (approved):

PR #1703, opened 2021-11-24
Add filtering possibility for `float` and `atk4_money` DBAL types

mkrecek234:
Please add these missing filters for table columns


mvorisek:
I prefer to register `TypeNumber` class to handle the float/atk4_money types instead.

mkrecek234:
It auto-references the Filter today towards FilterAtk4_money, so in case this is missing, no filtered grid will work. I would prefer to include by now, so grid supports any field type and is not throwing errors.

mvorisek:
@mkrecek234 please test

mkrecek234:
@mvorisek Tested and working👍

PR #1706, opened 2021-11-25
Impl. response using PSR-7 interface

abbadon1334:
`App::outputXxx/terminateXxx` methods not longer accepts `$headers` as 2nd param, call `$app->setResponseHeader()` before it if needed


src/App.php
```diff
@@ -1071,25 +1084,24 @@ protected function outputResponse(string $data, array $headers): void
             }
         }
 
-        if ($lateError === null && count($headersNew) > 0 && headers_sent() && !$isCli) {
+        if ($lateError === null && count($headersNew) > 0 && !empty($this->response->getHeaders()) && !$isCli) {
```
mvorisek: `headers_sent()` is correct here, it does not check if headers has been built, it checks if some unexpected output has been flushed and thus additional headers can not be flushed, therefore names also "lateError" (which can not be even rendered as we can not tell the browser the content-type will be html.

src/App.php
```diff
@@ -1099,11 +1111,11 @@ protected function outputResponse(string $data, array $headers): void
 
         if ($lateError !== null) {
             echo "\n" . '!! FATAL UI ERROR: ' . $lateError . ' !!' . "\n";
-
+            $this->callBeforeExit(); // output before forced exit TODO exit here is really needed?
```
mvorisek: see above what is "lateError", basically a fatal error we cannot recover from

the added line is wrong, we must simply echo the fatal error and exit with a non-zero code (as already coded)

src/App.php
```diff
@@ -1153,4 +1165,22 @@ private function getRenderedPortals(): array
 
         return $portals;
     }
+
+    /**
+     * @return void
```
mvorisek: never add phpdoc for harcoded types, provides no value...

src/App.php
```diff
@@ -1153,4 +1165,22 @@ private function getRenderedPortals(): array
...
+    protected function emitResponse(): void
+    {
+        (new \Laminas\HttpHandlerRunner\Emitter\SapiEmitter())->emit($this->response);
+    }
+
+    /**
+     * Used for PSR-11 implementation
+     *
+     * @return ResponseInterface
+     */
```
mvorisek:
```suggestion
```

unuseful, write code as much as great and that is enough

src/App.php
```diff
@@ -157,6 +159,12 @@ class App
 
     public $templateClass = HtmlTemplate::class;
 
+    /** @var ResponseInterface */
+    private $response;
+
+    /** @var string Seed Response class  */
```
mvorisek: seed !== class!
<==> array !== string-class

comment is thus even wrong, remove

mvorisek (changes requested):
request must be handled as PSR request as well

laminas dep should be not needed if this translates to a few echo/header/http_response_code lines

src/App.php
abbadon1334: correct, checking via `header_sent` bypass everything and raise an error for unexcepted output

src/App.php
abbadon1334: OK

abbadon1334:
> request must be handled as PSR request as well
> 
> laminas dep should be not needed if this translates to a few echo/header/http_response_code lines

I replace with code, but because of the small difference in performance, I always prefer the ready-made solutions, which are already tested in almost every situation, even those we don't know at the moment.

abbadon1334:
https://github.com/atk4/ui/blob/beb149f3e15ffdb681df9964cd5f48ed85971851/src/App.php#L1119
Are you sure that this cannot be changed to an Exception or other ways to not force exit the process?
My fear is that if you use Atk4 in a routing/handler contest, that exit in place of an Error will stop the flow without an output, without running post route middleware or logging of the error, without considering that to avoid that line you must override via copy/paste the whole method just to remove it.

abbadon1334:
> request must be handled as PSR request as well
> 
> laminas dep should be not needed if this translates to a few echo/header/http_response_code lines

I see the PR https://github.com/atk4/ui/pull/1671 which remove/assert that there is no global sticky.
i think in App __construct, if not injected via $defaults['request'], we create a Request using `nyholm/psr7-server` and all sticky + $_GET/$_POST/$_FILES that are already present somewhere in the app must point to App->request, what you think?

But before this i need to understand why demos, except for SSE, works perfect, but not in tests.

mvorisek:
> > request must be handled as PSR request as well
> > laminas dep should be not needed if this translates to a few echo/header/http_response_code lines
> 
> I replace with code, but because of the small difference in performance, I always prefer the ready-made solutions, which are already tested in almost every situation, even those we don't know at the moment.

replacing literaly no more than several lines and doubling the non-dev dependencies, is IMHO really not helpful

mvorisek:
> https://github.com/atk4/ui/blob/beb149f3e15ffdb681df9964cd5f48ed85971851/src/App.php#L1119
> 
> Are you sure that this cannot be changed to an Exception or other ways to not force exit the process?
> My fear is that if you use Atk4 in a routing/handler contest, that exit in place of an Error will stop the flow without an output, without running post route middleware or logging of the error, without considering that to avoid that line you must override via copy/paste the whole method just to remove it.

can be put into a protected method that can be overriden with a custom exit handler, but as described, it is too late to have a full control of the output as partly sent already

src/App.php
```diff
@@ -1158,4 +1169,31 @@ private function getRenderedPortals(): array
...
     }
+
+    protected function emitResponse(): void
+    {
+        $http_line = sprintf('HTTP/%s %s %s',
+            $this->response->getProtocolVersion(),
+            $this->response->getStatusCode(),
+            $this->response->getReasonPhrase()
+        );
+        header($http_line, true, $this->response->getStatusCode());
```
mvorisek:
```suggestion
        http_response_code($this->response->getStatusCode());
```
I think `header` will not set the internal php status code and read by `http_response_code()` will not be possible otherwise (in theory, I know, we are about to exit anyway...)

src/App.php
```diff
@@ -1158,4 +1169,31 @@ private function getRenderedPortals(): array
...
+        foreach ($this->response->getHeaders() as $name => $values) {
+            foreach ($values as $value) {
+                header("$name: $value", false);
```
mvorisek:
```suggestion
                header($name: ' . $value, false);
```
just preferred vs

src/App.php
```diff
@@ -1158,4 +1169,31 @@ private function getRenderedPortals(): array
 
         return $portals;
     }
+
+    protected function emitResponse(): void
```
mvorisek:
```suggestion
    /**
     * @return never
     */
    protected function emitResponse(): void
```

If this method is expected to be terminating, use this phpdoc and put `exit;` at the end of the method

but not sure about SSE

abbadon1334: IMHO we cannot exit like this, this method only output the response but the app must terminate as before, with stop execution or catching ExitApplicationError in callExit, called by App::run() or App::terminate(), if we exit there HOOK_BEFORE_EXIT is not fired and in other context will break the handler flow.

What you think about making it private and add a boolean flag to App like $emit_response?

src/App.php
abbadon1334: Ok, but probably in this way : `header($name . ': ' . $value, false);`

src/App.php
abbadon1334: OK

src/App.php
mvorisek: yes ;-)

src/App.php
mvorisek: I see, never cannot be used...

abbadon1334:
@mvorisek did you remember why you need this? https://github.com/atk4/ui/blob/239db622479b85d93ba7e3abc1b0e0e4099c78f9/src/App.php#L1068

Because now with response i think is not needed anymore, if i remove that i can get all headers correctly and 100% Unit Tests pass

mvorisek:
> @mvorisek did you remember why you need this?
> 
> https://github.com/atk4/ui/blob/239db622479b85d93ba7e3abc1b0e0e4099c78f9/src/App.php#L1068
> 
> Because now with response i think is not needed anymore, if i remove that i can get all headers correctly and 100% Unit Tests pass

lateError is not tested, but it prevents phpunit to exit/display lateErrors

abbadon1334:
some Tests are failing at the moment due to github problems, but I think all tests are successful.

src/App.php
```diff
@@ -1063,25 +1072,50 @@ function () {
...
+        }
+
         $isCli = \PHP_SAPI === 'cli'; // for phpunit
 
         if (!headers_sent() || $isCli) {
             foreach ($headersNew as $k => $v) {
                 if (!$isCli) {
                     if ($k === self::HEADER_STATUS_CODE) {
-                        http_response_code($v === (string) (int) $v ? (int) $v : 500);
+                        $this->response = $this->response->withStatus($v === (string) (int) $v ? (int) $v : 500);
```
mvorisek: if response is immutable, then will have too much overhead, all headers needs to be set at once, eg. response should be build and set once per `outputResponseUnsafe` call

src/App.php
```diff
@@ -1063,25 +1072,50 @@ function () {
      */
     protected function outputResponseUnsafe(string $data, array $headersNew): void
     {
+        $isSse = $this->response->getHeaderLine('Content-Type') === 'text/event-stream';
+
+        if (headers_sent() && $isSse) {
+            echo $data;
```
mvorisek: there is no streaming in PSR-7 response?

otherwise we should build it/write to it as for normal response

also, what is the point of building PSR-7 response here when we echo it a few lines later?

abbadon1334: There is many discussion on this topic :
https://github.com/slimphp/Slim/issues/2481
https://github.com/laminas/laminas-diactoros/issues/16

I see many suggests to use callback stream but in this case we have to make changes at JsSse level, Slim have a different approach which uses this : https://github.com/slimphp/Slim-Psr7/blob/master/src/NonBufferedBody.php

In Atk JsSse is blocking and we cannot return the response, while JsSse is in callback execution, so we can take this KISS approach and output following request to output data or we can use something like this : https://github.com/phly/psr7examples/blob/master/src/CallbackStream.php and execute the Sse in returned stream, which is IMHO better but we have to set special response directly in JsSse, and callback firing will be done when response will be emitted.

I like the second one, but i try to make as less changes as possible, to avoid conflict with other open PR.

src/App.php
abbadon1334: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-7-http-message.md
The interface is clear add a single header and gets a new response with the new header, that's it, a different approach will do what I write there and I have a fear that creating a totally new response as no reason and is bad for future implementation (ex. JsSse) where i need to setup a response before creation.

mvorisek: so for the 1st time (and it is normally called only once) constructor can be used - https://github.com/Nyholm/psr7/blob/master/src/Response.php#L42

https://github.com/Nyholm/psr7/blob/master/src/MessageTrait.php#L81 `withHeader` clones all the time

src/App.php
mvorisek: the nonBufferedBody seems not usefull at all

> CallbackStream

seems all to address read-once concept, which is not what we actually need

> KISS approach and output following request to output data

with custom Response class (that we can detect using instanceof) this seems the best to me and individual SSE responses there

I think it is ok for now if we keep the SSE impl. minimal like now and implement it only when there is a valid usecase and we actually know what the usecase needs.

One major question still: if target use still need to extent App (to override `emitResponse`, othewise App will echo), why is PSR-7 helpfull to have in App if the target use can override `outputResponseUnsafe` in a simillar fashion?

If I do not missed something, having `response` in an app is not usefull at all (as it is built once per request anyway, we do not need it as a place to add a data to) :)

tests/DemosTest.php
```diff
@@ -143,8 +143,8 @@ protected function getClient(): Client
             $localPath = static::ROOT_DIR . $request->getUri()->getPath();
 
             ob_start();
+            $app = $this->createTestingApp();
```
mvorisek: this must be kept in try, otherwise try/finally for buffering is not atomic!!!

src/App.php
mvorisek: We should integrate https://github.com/abbadon1334/atk4-psr7/tree/master/src into this repo. The `PSR7App` should be transformed to a trait with `outputResponseUnsafe` method and `protected getDefaultsForPsr7Response` method to get the needed default properties which should be merged with the defaults in an user App class constructor.

tests/DemosTest.php
abbadon1334: later i need to call `$app->getResponse()` to check it, i do it only to avoid errors in static code analysis, how to instruct phpstan that is not an error? is the only solution `@phpstan-ignore-line`

src/App.php
abbadon1334: agree, but now if we remove the response and remove the last added method `emitResponse`, we can output directly the response in `outputResponseUnsafe` as we do before.
At this point if someone needs PSR7 implementation, we can avoid copy/paste the whole method `outputResponse` just to change 2 lines, now is all much clear with `outputResponseUnsafe`.

About useful or not useful, Psr7 if implemented correctly will allow integration of Atk4, in every framework nearly without touching it, but practically, the emitter of response is not needed, Psr7 is for Request/Response standardization to allow implementation of routing + middleware.

About integration, I think I need to find a correct way to implement JsSse after that we speak about integration, at now except for usage of NonBufferedBody, which equally breaks the output stream, this PR and atk4-psr7 are at the same point.

src/App.php
abbadon1334: Ok, it works for sure, but behind is nearly the same code : https://github.com/Nyholm/psr7/blob/fafc2faf8d49b979f922fe416dc8805462dc1fe8/src/MessageTrait.php#L138

Can be a problem because interface don't define a contract for `__construct`, so construct can be different from an implementation to another

mvorisek: You are right, we must follow the interface (as the optimized constructor is not present).

On the otherside, if this is expected to be called only once per app lifecycle we should actually instance the response class here and then assign. `responseClass` is not helpfull at all, if middleware needs a specific class, it should convert from our (PSR-7 Response interfaced) response object.

abbadon1334: In my experience, whatever response we use if implements ResponseInterface is OK, so from what i see nyholm is the much stricter response library, i think we can stick to it and think is not our problem until someone will raise an issue, i think is a remote case because is the beauty of working with interface.
So we can remove the $responseClass and raise visibility to protected of $response to allow DI, what you think?

src/App.php
abbadon1334: > If I do not missed something, having response in an app is not usefull at all (as it is built once per request anyway, we do > not need it as a place to add a data to) :)

Rethinking about your affirmation, i think with Response we have another way to simplify some code in App.
if we remove normalization and other things we do to set/check headers in App response and redirect all directly to Response to build it? At the end we have only to emit the response in place of recover status + headers and build and emit

tests/DemosTest.php
mvorisek: then you do something wrong ;-)

src/App.php
mvorisek: Well, it is all quite handly and well written there... The normalization and diffing there is better than what it can be then. So I prefer to keep it, maybe one day once all other ugly things are fixed :)

tests/DemosTest.php
abbadon1334: > then you do something wrong ;-)

this time I don't think so, I move back in try/catch, phpstan will fail. anyway i can added 2 `@phpstan-ignore-line`

abbadon1334: Fixed

src/App.php
abbadon1334: Totally agree, but FYI in Nyholm Response when you call withHeader internally normalize the header, clearly doing what we do in App https://github.com/Nyholm/psr7/blob/fafc2faf8d49b979f922fe416dc8805462dc1fe8/src/MessageTrait.php#L59-L89

tests/DemosTest.php
```diff
@@ -189,6 +189,17 @@ class_exists(Utils::class) ? Utils::streamFor($body) : \GuzzleHttp\Psr7\stream_f
                 $body->rewind();
             }
 
+            $this->assertSame(
```
mvorisek: I prefer assertions on the same line if not too long otherwise the LOC it too high and file overview is limited.

To your phpstan strugles:
```
ob_start();
try {
create app
try {} catch {}
} finally {
ob end...
}
```

src/App.php
mvorisek: > internally normalize the header, clearly doing what we do in App

then go ahead with it, no need to have the code twice!

composer.json
```diff
@@ -54,7 +54,8 @@
     ],
     "require": {
         "php": ">=7.4 <8.2",
-        "atk4/data": "dev-develop"
+        "atk4/data": "dev-develop",
+        "nyholm/psr7": "^1.4"
     },
     "require-release": {
```
mvorisek, line 63: make sure `nyholm/psr7` is added here as well, 1:1

abbadon1334: i dunno why is missing, i do a rollback, after some changes to merge App.php diff correctly and push was approved.

src/App.php
```diff
@@ -1065,42 +1055,29 @@ function () {
...
      * Output Response to the client.
      */
     protected function outputResponse(string $data, array $headers): void
     {
-        $this->response_headers = $this->normalizeHeaders($this->response_headers);
-        $headersAll = array_merge($this->response_headers, $this->normalizeHeaders($headers));
-        unset($headers);
-        $headersNew = array_diff_assoc($headersAll, self::$_sentHeaders);
-        unset($headersAll);
+        foreach ($headers as $name => $value) {
```
abbadon1334: duplicate code, 3 times.

abbadon1334:
the next step is to replace all direct calls to $_GET and $_POST. for ex :
https://github.com/atk4/ui/blob/09ec6e11c3c709cb2e5ac78b71ceee65d836b3d0/src/Accordion.php#L151

can be replaced with : `$this->getApp()->issetRequestQueryParam('__atk-dyn-section')` and this is ok, the problem is where is no reference to App, for ex. :
https://github.com/atk4/ui/blob/09ec6e11c3c709cb2e5ac78b71ceee65d836b3d0/src/Persistence/Post.php#L19

which cannot be changed with : `$this->getApp()->getRequestFormParam($field)` because Persistence has no reference to App

IMHO every external param must be recovered from the request object, getting request vars directly from global $_GET or $_POST is wrong or at least makes usage of sanitizer or intrusion check middlewares useless.

@mvorisek what is your opinion on this?

tests/AppPsr7Test.php
```diff
@@ -0,0 +1,55 @@
...
+use Atk4\Core\Phpunit\TestCase;
+use Atk4\Ui\App;
+use Atk4\Ui\Exception\ExitApplicationError;
+use Atk4\Ui\Layout;
+
+class AppPsr7Test extends TestCase
+{
+    protected function getApp(): App
+    {
+        gc_collect_cycles();
```
mvorisek: why is explicit GC here needed?

src/App.php
```diff
@@ -1200,4 +1177,79 @@ private function getRenderedPortals(): array
 
         return $portals;
     }
+
+    public function getResponse(): ResponseInterface
```
mvorisek: The App class ode is quite long. Do you think it is a good idea to move all request/response code to `App\Request` and `App\Response`?

abbadon1334: I had the same idea, but we can opt for something like App\Http\Response|Request to avoid new classes on root namespace, what do you think?

tests/AppPsr7Test.php
abbadon1334: probably a typo i will remove later and see if the test passes, but i see that some **first** tests fails.

src/App.php
mvorisek: `App\Request` put into `App` directory seems exactly that solution ;-)

tests/DemosTest.php
```diff
@@ -512,7 +485,7 @@ public function testDemoCallbackError(string $path, string $expectedExceptionMes
         $response = $this->getResponseFromRequest5xx($path);
 
         $this->assertSame(500, $response->getStatusCode());
-        $responseBodyStr = $response->getBody()->getContents();
+        $responseBodyStr = (string) $response->getBody();
```
mvorisek: It seems the body content is not rewound somewhere.

__toString() calls rewinds, but we should support `getContents` if possible

abbadon1334: LGTM

mvorisek: original `getContents` returns an empty string

I belive it is an issue in the impl, probably the output buffer is not rewound somewhere.

mvorisek: `rewind` in in DemosTest added, fixed

src/App.php
mvorisek: Let's keep it as is for now...

src/App.php
```diff
@@ -139,12 +143,37 @@ class App
...
+    private $response;
+
+    /** @var ServerRequestInterface */
+    private $request;
+
     public function __construct(array $defaults = [])
     {
         $this->setApp($this);
 
+        $this->response = $defaults['response'] ?? new Response();
```
mvorisek: we should support atk4 array seeding

mvorisek:
> IMHO every external param must be recovered from the request object, getting request vars directly from global $_GET or $_POST is wrong or at least makes usage of sanitizer or intrusion check middlewares useless.
> 
> @mvorisek what is your opinion on this?

Definitely, maybe even unset such env. vars to make sure, all accesses are done thru the App/PSR request interface

src/App.php
```diff
@@ -144,12 +148,45 @@ class App
...
+        if (isset($defaults['response'])) {
+            unset($defaults['response']);
+        }
+
+        // TODO remove this before PSR7 PR merge, fix tests to pass /wo this develop merge hack
```
mvorisek: TODO before merge

mvorisek: @abbadon1334 I have merged develop branch in and removed this TODO, please adjust the code accordingly to make CI gree again

src/App.php
```diff
@@ -341,39 +341,6 @@ public function getResponse(): ResponseInterface
...
-    /**
-     * Return $_POST param by key or null if not exists.
-     *
-     * @return mixed
-     */
-    public function getRequestPostParam(string $key)
-    {
-        return $this->request->getParsedBody()[$key] ?? null;
-    }
-
```
mvorisek: to be reintroduced in #2009

mvorisek:
Thank you @abbadon1334!

mvorisek:
Thank you @abbadon1334 ❤️

PR #1710, opened 2021-11-29
Handle late error when exception handler is active

mvorisek:
steps to reproduce:

1. output some unwatend output (`echo 'xx';`)
2. throw some exception

test for this should be added before merge

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -1103,6 +1103,7 @@ class App
                 $lateError = new LateOutputError('Unexpected output detected');
                 if ($this->catch_exceptions) {
                     $this->caughtException($lateError);
+                    $this->outputLateOutputError($lateError);
                 }
 
                 throw $lateError;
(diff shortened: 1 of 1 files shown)


src/App.php
```diff
@@ -1100,14 +1100,24 @@ protected function outputResponse(string $data, array $headers): void
 
         foreach (ob_get_status(true) as $status) {
             if ($status['buffer_used'] !== 0) {
-                throw new LateOutputError('Unexpected output detected');
+                $lateError = new LateOutputError('Unexpected output detected');
+                if ($this->catch_exceptions) {
+                    $this->caughtException($lateError);
```
mvorisek, line 1105: should terminate (and if not overrode it always will), if it does not, throw line below save us

PhilippGrashoff:
Thumbs up for this one, very sensible.

abbadon1334 (approved):
Preapprove it, agree to add a test, we need it to check for PSR7

mvorisek:
Merging, impossible to unit test as error/exception handler cannot be removed easily, as we do not know if the latest handler is the one set from the App - https://www.php.net/manual/en/function.restore-error-handler.php

Late errors can be tested with Behat only.

PR #1711, opened 2021-11-29
Fix form2.php demo

### LINKED ISSUE #1704: Field names missing in Model->getField, causing corrupt Exception message
In recent develop, example `demos/form/form2.php` don't work:
- First form can never be saved, as iso ~cannot be empty, nor~ cannot it be changed
- If you deactivate all validation handling in onSubmit to make it work again, still it will throw an error `Javascript Error
Syntax error, unrecognized expression: #`

The second form throws error `Atk4\Ui\Exception: Callback requested, but never reached. You may be missing some arguments in request URL.` on save.

ibelar:
Fix: #1704

### DIFF
--- a/demos/form/form2.php
+++ b/demos/form/form2.php
@@ -44,9 +44,9 @@ $formCodes->addControl(Country::hinting()->fieldName()->phonecode, ['width' => '
 
 // form names field group
 $formNames = $form->addGroup(['More Information about you']);
-$formNames->addControl('f_name', ['width' => 'five', 'caption' => 'First Name']);
-$formNames->addControl('m_name', ['width' => 'five', 'caption' => 'Middle Name']);
-$formNames->addControl('l_name', ['width' => 'six', 'caption' => 'Last Name']);
+$formNames->addControl('first_name', ['width' => 'five', 'caption' => 'First Name']);
+$formNames->addControl('middle_name', ['width' => 'five', 'caption' => 'Middle Name']);
+$formNames->addControl('last_name', ['width' => 'six', 'caption' => 'Last Name']);
 
 // form on submit
 $form->onSubmit(function (Form $form) {
@@ -60,11 +60,11 @@ $form->onSubmit(function (Form $form) {
 
     // In-form validation
     $errors = [];
-    if (mb_strlen($form->model->get('f_name')) < 3) {
-        $errors[] = $form->error('f_name', 'too short, ' . $form->model->get('f_name'));
+    if (mb_strlen($form->model->get('first_name')) < 3) {
+        $errors[] = $form->error('first_name', 'too short, ' . $form->model->get('first_name'));
     }
-    if (mb_strlen($form->model->get('l_name')) < 5) {
-        $errors[] = $form->error('l_name', 'too short');
+    if (mb_strlen($form->model->get('last_name')) < 5) {
+        $errors[] = $form->error('last_name', 'too short');
     }
 
     // Model validation. We do it manually because we are not using Model::save() method in demo mode.
--- a/demos/form/form2.php
+++ b/demos/form/form2.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Atk4\Ui\Demos;
 
 use Atk4\Ui\Form;
+use Atk4\Ui\JsToast;
 
 /** @var \Atk4\Ui\App $app */
 require_once __DIR__ . '/../init-app.php';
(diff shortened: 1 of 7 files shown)


mvorisek:
@ibelar how can this fix #1704 when there is no JS related code fixed?

`Javascript Error Syntax error, unrecognized expression: #` is something that should never happen as it signals some non-100% quoting somewhere

demos/form/form2.php
```diff
@@ -21,55 +22,61 @@
...
+$formCodes->addControl(Country::hinting()->fieldName()->iso3, ['width' => 'four'])->iconLeft = 'flag';
+$formCodes->addControl(Country::hinting()->fieldName()->numcode, ['width' => 'four'])->iconLeft = 'flag';
+$formCodes->addControl(Country::hinting()->fieldName()->phonecode, ['width' => 'four'])->iconLeft = 'flag';
 
 // form names field group
 $formNames = $form->addGroup(['More Information about you']);
-$formNames->addControl('first_name', ['width' => 'eight']);
-$formNames->addControl('middle_name', ['width' => 'three']);
-$formNames->addControl('last_name', ['width' => 'five']);
+$formNames->addControl('f_name', ['width' => 'five', 'caption' => 'First Name']);
```
mvorisek: original `first_name` looks much better to me

src/Form.php
```diff
@@ -307,6 +307,9 @@ public function onSubmit(\Closure $callback)
             } catch (\Atk4\Data\ValidationException $val) {
                 $response = [];
                 foreach ($val->errors as $field => $error) {
+                    if (!$field) {
+                        throw ($val);
```
mvorisek:
```suggestion
            } catch (\Atk4\Data\ValidationException $e) {
                $response = [];
                foreach ($val->errors as $field => $error) {
                    if (!$field) {
                        throw $e;
```

src/JsCallback.php
```diff
@@ -115,12 +115,8 @@ public function set($fx = null, $args = null)
...
 
                 $this->terminateAjax($ajaxec);
-            } catch (\Atk4\Data\ValidationException $e) {
-                // Validation exceptions will be presented to user in a friendly way
-                $msg = new Message($e->getMessage());
-                $msg->addClass('error');
-
-                $this->terminateAjax(null, $msg->getHtml(), false);
+            } catch (\Exception $e) {
+                $this->terminateAjax(null, $this->getApp()->renderExceptionHtml($e), false);
```
mvorisek: one question - why catch, when we can leave to app exception handler?

ibelar: Do you mean `throw ($e)`?

ibelar: or remove entire try - catch block?

src/Form.php
ibelar: missing change for $val->errors - I will update

ibelar:
@mvorisek - any idea why all MSSQL test are failing?

src/JsCallback.php
mvorisek: remove entire catch, I think it is not needed as renderExceptionHtml is the normal exception render when an exception is caught in app, but of course check if the behaviour is the same...

mvorisek:
> @mvorisek - any idea why all MSSQL test are failing?

yes, merge the latest develop (standard DBAL DSN is required in GH CI config)

ibelar:
> yes, merge the latest develop (standard DBAL DSN is required in GH CI config)

Ah - thanks

src/JsCallback.php
ibelar: Done!

src/Behat/Context.php
```diff
@@ -45,7 +45,9 @@ public function waitUntilLoadingAndAnimationFinished(AfterStepScope $event): voi
     {
         $this->jqueryWait();
         $this->disableAnimations();
-        $this->assertNoException();
+        if ($event->getFeature()->getTitle() !== 'Exception') {
```
mvorisek: we need to have a special step for testing exceptions, this is a dirty hack, I will fix it

src/Form.php
```diff
@@ -304,9 +304,12 @@ public function onSubmit(\Closure $callback)
                 }
 
                 return $response;
-            } catch (\Atk4\Data\ValidationException $val) {
+            } catch (\Atk4\Data\ValidationException $e) {
                 $response = [];
-                foreach ($val->errors as $field => $error) {
+                foreach ($e->errors as $field => $error) {
+                    if (!$field) {
```
mvorisek: as @mkrecek234 pointed out, this needs to be fixed in data, however, I belive, there is a usecase when field is empty when the source of the validation exception is unknown

mvorisek: fixed in https://github.com/atk4/data/pull/942

src/JsCallback.php
```diff
@@ -102,26 +102,18 @@ public function set($fx = null, $args = null)
...
-                }
-
-                $response = $fx(...array_merge([$chain], $values));
+            $values = [];
+            foreach ($this->args as $key => $value) {
+                $values[] = $_POST[$key] ?? null;
+            }
 
-                $ajaxec = $response ? $this->getAjaxec($response, $chain) : null;
+            $response = $fx(...array_merge([$chain], $values));
```
mvorisek:
```suggestion
            $response = $fx($chain, ...$values);
```

PR #1714, opened 2021-12-01
Setup Flatpickr locale from ui persistence data

mvorisek:
fixes setup of `firstDayOfWeek` from ui persistence, which was not honored (setup using static Calendar method was required)

### DIFF
--- a/src/Form/Control/Calendar.php
+++ b/src/Form/Control/Calendar.php
@@ -4,9 +4,7 @@ declare(strict_types=1);
 
 namespace Atk4\Ui\Form\Control;
 
-use Atk4\Ui\App;
 use Atk4\Ui\Jquery;
-use Atk4\Ui\JsChain;
 use Atk4\Ui\JsExpression;
 
 /**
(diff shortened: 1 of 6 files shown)


src/Form/Control/Calendar.php
```diff
@@ -33,62 +31,30 @@ public function setOption($name, $value)
...
-        $app->html->js(true, (new JsExpression('flatpickr.l10ns.default.firstDayOfWeek = [day]', ['day' => $day])));
-    }
-
-    /**
-     * Load flatpickr locale file.
-     * Pass it has an option when adding Calendar input.
-     *  Form\Control\Calendar::requireLocale($app, 'fr');
-     *  $form->getControl('date')->options['locale'] = 'fr';.
-     */
-    public static function requireLocale(App $app, string $locale)
```
ibelar, line 50: Wy not keep this static method in order to easily load locale setting within a multi language environment?

mvorisek: because "multi language environment" in a single app/render is non-sense :)

ui persistence `locale` can be easily changed once per app render

mvorisek: @ibelar does it make sense, ok to merge?

ibelar (approved):
LGTM

Issue #1689, opened 2021-11-05
Grid Filter Throwing Exception on empty Date filter


mkrecek234:
In a grid view where you activate the filter function, do the following step to reproduce an error:

1. Have a grid view with a DateTime column
2. Add a filter on a datetime column, and just press "Set" (leave the filter empty).

This will throw an error `TypeError: DateTime::__construct(): Argument #1 ($datetime) must be of type string, null given`
As the filter is stored in a cookie, this error will never go away on reload of the page. The user is trapped, unless he deletes cookies.

Verified in 2.4 and current dev-develop branches

mvorisek:
please reproduce on demos

mkrecek234:
@mvorisek Please check my PR on this - the filter is not capable handling an empty/null input field which should be resolved using the attached PR.

mvorisek:
fixed in https://github.com/atk4/ui/pull/1718

PR #1717, opened 2021-12-05
Fix datetime filter with empty input

mkrecek234:
Fix grid error when applying an empty value to a datetime field, see issue #1689


mvorisek:
rebase on the latest develop, also, `(` `)` outside the expr is not needed

PR #1718, opened 2021-12-05
Fix datetime grid filter bug on empty input

mkrecek234:
@mvorisek As requested, sorry for the earlier PR - solves issue #1717

### DIFF
--- a/src/Table/Column/FilterModel/TypeDate.php
+++ b/src/Table/Column/FilterModel/TypeDate.php
@@ -123,7 +123,7 @@ class TypeDate extends Column\FilterModel
 
                 break;
             default:
-                $date = new DateTime($dateModifier);
+                $date = $dateModifier ? new DateTime($dateModifier) : null;
 
                 break;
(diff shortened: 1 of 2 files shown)


mvorisek:
next time, better to keep the same PR...

how to test: issue can be tested on `collection/tablefilter.php` demo by replacing `CountryLock` with `Stat`

this PR is bad, a) this does not fix the same issue in other Dt filters (TypeDate, TypeDateTime, TypeTime), b) code of these filters should be deduplicated

mkrecek234:
@mvorisek Sorry, but I try just to help fixing bugs even if it is not as exhaustive as you would fancy (to your point that PR is "bad"). I think contributing should not be criticised as "bad" or "good". Every contribution identifying or fixing even the most minor bug should be considered "helpful". If you can make it even better, e.g., by even improving code by deduplicating, everyone will appreciate this a lot. 

Please check commit for datetime fixing the same issue

PR #1719, opened 2021-12-24
Add late output error test

mvorisek:
extracted from https://github.com/atk4/ui/pull/1706 , originally authored by @abbadon1334

I belive php instance state cannot be reverted after some late error tests, eg. the side effects of unamanged headers/output sent can be tested with HTTP tests only.

Also fixes 500 status code when unmanaged output/headers are flushed by an user manually.


demos/_unit-test/late-output-error.php
```diff
@@ -0,0 +1,39 @@
...
+$cb2 = CallbackLater::addTo($app);
+$cb2->setUrlTrigger('trigger_err_unexpected_output');
+
+$modal2 = \Atk4\Ui\Modal::addTo($app, ['cb' => $cb2]);
+$modal2->name = 'm_header';
+
+$modal2->set(function () {
+    ob_start();
+    // do we detect only flushed content or unflushed too?
+    echo 'unmanaged output';
```
mvorisek, line 32: I belive this cannot be tested with JSON format, as "header sent" will be triggered first

PR #1721, opened 2021-12-29
Fix user action usage /w entity

mvorisek:
adjust for https://github.com/atk4/data/pull/943 changes

### DIFF
--- a/demos/_includes/DemoActionsUtil.php
+++ b/demos/_includes/DemoActionsUtil.php
@@ -86,7 +86,7 @@ class DemoActionsUtil
             [
                 'caption' => 'Edit ISO3',
                 'description' => function (UserAction $action) {
-                    return 'Edit ISO3 for country: ' . $action->getEntity()->getTitle();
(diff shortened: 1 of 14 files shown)


demos/_includes/DemoActionsUtil.php
```diff
@@ -86,7 +86,7 @@ public static function setupDemoActions(CountryLock $country): void
             [
                 'caption' => 'Edit ISO3',
                 'description' => function (UserAction $action) {
-                    return 'Edit ISO3 for country: ' . $action->getEntity()->getTitle();
+                    return 'Edit ISO3 for country: ' /* TODO . $action->getEntity()->getTitle()*/;
```
mvorisek, line 89: originally, (empty) entity was created when it was not set and `getEntity()` was called

with https://github.com/atk4/data/commit/ec4a51a1d942d3df9e436b93185cfeba48acf281#diff-5588f8a6d190c5213f436c880d33b4790cf9f58d2e5f04c7b1e64a7301de15a7L228 no entity is automatically created and this was failing, no replacement avaialble, as the usage is wrong

PR #1725, opened 2022-01-04
Remove Lock demo Models

mvorisek:


demos/init-db.php
```diff
@@ -20,8 +20,48 @@
...
+{
+    public function atomic(\Closure $fx)
+    {
+        $eForRollback = new \Exception('Prevent modification');
+        $res = null;
+        try {
+            parent::atomic(function () use ($fx, $eForRollback, &$res) {
+                $res = $fx();
+
+                throw $eForRollback;
```
mvorisek, line 33: FYI this guarantees all save (insert/update/delete) operations will be executed, but always rolled back

demos/init-db.php
```diff
@@ -20,8 +20,48 @@
...
+{
+    public function atomic(\Closure $fx)
+    {
+        $eRollback = new \Exception('Prevent modification');
+        $res = null;
+        try {
+            parent::atomic(function () use ($fx, $eRollback, &$res) {
+                $res = $fx();
+
+                throw $eRollback;
```
mvorisek, line 33: FYI this guarantees all save (insert/update/delete) operations will be executed, but always rolled back

PR #1743, opened 2022-03-01
Fix Lookup add new button

### LINKED ISSUE #1731: Lookup "plus" feature throws error
This line causes an error when you choose to have the Lookup with Lookup::plus = true:

https://github.com/atk4/ui/blob/3cb026a3649efc1841d5d73a7608145da6212971/src/Form/Control/Lookup.php#L257

Error is "Atk4\Data\Exception: Expected model, but instance is an entity".

mkrecek234:
fixes https://github.com/atk4/ui/issues/1731

### DIFF
--- a/src/Form/Control/Lookup.php
+++ b/src/Form/Control/Lookup.php
@@ -251,10 +251,10 @@ class Lookup extends Input
 
         $vp->set(function ($page) {
             $form = \Atk4\Ui\Form::addTo($page);
-
-            $entity = $this->model->createEntity();
-
-            $form->setModel($entity->setOnlyFields($this->plus['fields'] ?? []));
+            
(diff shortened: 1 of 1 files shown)


src/Form/Control/Lookup.php
```diff
@@ -251,10 +251,10 @@ protected function initQuickNewRecord()
 
         $vp->set(function ($page) {
             $form = \Atk4\Ui\Form::addTo($page);
-
-            $entity = $this->model->createEntity();
-
-            $form->setModel($entity->setOnlyFields($this->plus['fields'] ?? []));
+
```
mvorisek: trailing spaces need to be removed, see CS fixer diff

mkrecek234: @mvorisek See new commit

mvorisek:
looks good, thank you!

PR #1747, opened 2022-03-04
Add table column filter support for JSON type

### LINKED ISSUE #1744: Table\Column\FilterModel throws error on json type fields
If we have `type=>json` field in model and for some strange reason add it as column in grid and enable filtering, then `Table\Column\FilterModel` throws error
```
Critical Error
ErrorException: Undefined index: json

in vendor/atk4/ui/src/Table/Column/FilterModel.php:75
```

That's because filtering for this field type is not implemented and that's kind of understandable, but at least it shouldn't throw such ugly exception. Better would be to simply not add filtering in this column.

Probably same happens with fields `type=>object` and maybe `enum` and `values` type of data fields.

### LINKED ISSUE #1746: Table\Column\FilterModel do not support "Is Not" for string columns
For ordinary string columns we have these conditions:
image

Would it be possible to also add "Is Not" there like we have for date field for example?
I wanted to filter all rows where string column `http_status <>"OK"`

DarkSide666:
fix #1744
fix #1746


src/CardDeck.php
```diff
@@ -420,7 +420,7 @@ protected function initPaginator()
         $count = $this->model->action('count')->getOne();
```
mvorisek, line 420: cast here

src/Table/Column/FilterModel.php
```diff
@@ -65,6 +65,9 @@ public static function factoryType(App $app, Field $field): self
             Types::DATETIME_MUTABLE => FilterModel\TypeDatetime::class,
             Types::DATETIME_IMMUTABLE => FilterModel\TypeDatetime::class,
 
+            Types::JSON => FilterModel\TypeString::class,
```
mvorisek, line 68: These types are casted to non-string in PHP, the php values are never used in these filters?

src/Table/Column/FilterModel/TypeString.php
```diff
@@ -13,7 +13,7 @@ protected function init(): void
     {
         parent::init();
 
-        $this->op->values = ['is' => 'Is', 'contains' => 'Contains', 'start' => 'Start with', 'end' => 'End with'];
+        $this->op->values = ['is' => 'Is', 'is not' => 'Is Not', 'contains' => 'Contains', 'start' => 'Start with', 'end' => 'End with'];
```
mvorisek, line 16: and not contains, not starts... Not sure if this is wanted...

DarkSide666: At least simple "Is Not" definitely is good to have and also - it was already implemented there, but wasn't added in op->values array :)

src/Table/Column/FilterModel.php
DarkSide666: Yes it looks like that. For example, type=json field shows as text in grid column and looks like filter works on top of it.
Probably @ibelar could check this in more details too.
Of course, filtering as string will not be the best option, but at least it will not throw exception then and at least partly work as filter.

mvorisek: I also wonder if such filtering is even cross-DB compatible when the DB column type is native like JSON.

If it is, then maybe we can add `FilterModel\TypeString::class` as a fallback for any unspecified type

DarkSide666: If not compatible with all DB engines, then maybe we can somehow not add filter on these columns at all (automatically)?
What I don't like there is that ugly exception we get if json column is added in Grid with filters.

src/CardDeck.php
```diff
@@ -417,7 +417,7 @@ private function getModelActions(string $appliesTo): array
      */
     protected function initPaginator()
     {
-        $count = $this->model->action('count')->getOne();
+        $count = (int) $this->model->action('count')->getOne();
```
mvorisek, line 420: replace all occurences, see https://github.com/atk4/data/pull/1008, no cast needed then

mvorisek (approved):
LGTM, waiting for a green CI

mvorisek (changes requested):
image

for some reasons, Behat /w Coverage does not pass, it seems unrelated, but better to fix it first

mvorisek:
Thank you!

PR #1756, opened 2022-03-06
Using flag decorator in card table

DarkSide666:


src/Table/Column/Flag.php
```diff
@@ -0,0 +1,33 @@
...
+use Atk4\Data\Model;
+use Atk4\Ui\Table;
+
+/**
+ * Column for formatting country code as flags.
+ */
+class Flag extends Table\Column
+{
+    /** @var string Name of model field which contains country ALPHA-2 (2 letter) codes. */
+    public $code_field;
```
PhilippGrashoff: Didn't we switch to camelCase property names?

mvorisek: yes, we did everywhere!

unfortunatelly, PHP CS Fixer has no rule for it, so we need to rely on reviews, thanks!

src/Table/Column/Flag.php
```diff
@@ -0,0 +1,34 @@
...
+
+    public function getHtmlTags(Model $row, ?Field $field): array
+    {
+        $countryCode = $row->get($this->codeField);
+        $countryName = $this->nameField ? $row->get($this->nameField) : null;
+
+        return [
+            $field->shortName => $countryCode === null ? '' : $this->getApp()->getTag('i', [
+                'class' => strtolower($countryCode) . ' flag',
+                'title' => $countryCode . ($countryName === null ? '' : ' - ' . $countryName),
```
mvorisek: `$countryCode` should be in UC

src/Table.php
```diff
@@ -482,7 +482,7 @@ public function renderRow(): void
                 if (!is_array($columns)) {
                     $columns = [$columns];
                 }
-                $field = !is_int($name) && $this->model->hasField($name) ? $this->model->getField($name) : null;
+                $field = is_int($name) ? null : $this->model->getField($name);
```
mvorisek: ❤️

src/Table/Column/Flag.php
DarkSide666: it is already in UC here in data, but to be sure I'll add strtoupper

DarkSide666:
DO NOT MERGE THIS

mvorisek:
resubmit as https://github.com/atk4/ui/pull/1959

PR #1757, opened 2022-03-08
Fix timezone bug in UI

mkrecek234:
ATK4/Data handling of timezones is correct, converting fields into the default PHP timezone is reverted back to persist_timezone erroneously.


mvorisek (changes requested):
As commented on the forum, I belive we need an UI TZ, the change presented in this PR can be limiting for others and also does not change the TZ handling for data loading (from UI to code).

PR #1763, opened 2022-03-29
Add custom decimal separator support for atk4_money type

mkrecek234:
Introduce support for configuring decimal and thousands separators for `atk4_money` type.

by default we use:
- `.` (point) as a decimal point separator as this is more or less international (in EU, `,` is default, but in `US`, it may be interchanged to thousands deparator)
- ` `&nbsp;(NBSP) as a thousands separator as this is readable across the world and also pastable directly to MS Excel:
  image
  \* NBSP does not work implicitly in MS Excel, but when copyed from web, NBSP is transformed to regular SP by the most common webbrowsers like Chrome or Firefox
  \* also MS Excels accept spaces strictly only as a thousands separator (at least 3 num digits must follow), leading/trailing space, space after currency symbol, but not anywhere else (like `1 2.3`, `1.888 999`)

### DIFF
--- a/src/Form/Control/Money.php
+++ b/src/Form/Control/Money.php
@@ -12,8 +12,9 @@ class Money extends Input
     public function getValue()
     {
         $res = parent::getValue();
+        $res = str_replace("\u{00a0}" /* Unicode NBSP */, ' ', $res);
 
-        return $res === null ? null : trim(str_replace($this->getApp()->ui_persistence->currency, '', $res), ' ' . "\u{00a0}" /* Unicode NBSP */);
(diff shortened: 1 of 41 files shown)


src/Form/Control/Money.php
```diff
@@ -17,7 +17,7 @@ public function getValue()
             return;
         }
 
-        return number_format($v, $this->getApp()->ui_persistence->currency_decimals);
+        return number_format($v, $this->getApp()->ui_persistence->currency_decimals, $this->getApp()->ui_persistence->decimal_separator, $this->getApp()->ui_persistence->thousands_separator);
```
mvorisek, line 20: I wonder why we do not use the UI persistence typecasst here

PR #1765, opened 2022-04-08
Fix nested modal support

mvorisek:
fixes #1732
fixes #1699
and highly improves the Behat testing quality

Steps to reproduce:

1. open debug example `demos/_unit-test/crud_gh1732.php`
2. click "flag" icon on the 1st row
3. click "edit" icon in the opened modal and notice `Mustard` name
4. click outside the modal to close it (the Product modal)
5. click outside the modal to close it (the Category modal)
7. repeat 2. step with 2nd row
8. repeat 3. step, notice, there is still `Mustard` name instead of `Cola`

PR #1769, opened 2022-04-09
Remove View::id prop in favor of NameTrait::name

mvorisek:
part of https://github.com/atk4/ui/pull/1765

`View::id` was basically a copy of `NameTrait::name`, but it was not used consistently

Final deduplicated name might be not the most intuitive one, but it is given by core trait `NameTrait::name` and is also much more used than `View::id` in the current code base. In the future, we might introduce a more intuitive getter for it with stricter global uniqueness check.


PhilippGrashoff:
Thumbs up!

PR #1770, opened 2022-04-09
Add/remove specific HTML class /w "class.xxx" seed

mvorisek:
related with https://github.com/atk4/core/issues/206

also much safer, as invalid seeds with bool values for non-existent keys will throw now

### DIFF
--- a/demos/_includes/DemoLookup.php
+++ b/demos/_includes/DemoLookup.php
@@ -29,7 +29,7 @@ class DemoLookup extends \Atk4\Ui\Form\Control\Lookup
 
         $buttonSeed = is_string($buttonSeed) ? ['content' => $buttonSeed] : $buttonSeed;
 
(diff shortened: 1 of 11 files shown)


demos/_includes/DemoLookup.php
```diff
@@ -29,7 +29,7 @@ protected function initQuickNewRecord()
 
         $buttonSeed = is_string($buttonSeed) ? ['content' => $buttonSeed] : $buttonSeed;
 
-        $defaultSeed = [\Atk4\Ui\Button::class, 'disabled' => ($this->disabled || $this->readonly)];
+        $defaultSeed = [\Atk4\Ui\Button::class, 'class.disabled' => ($this->disabled || $this->readonly)];
```
mvorisek, line 32: ⚠️ be careful, some components like `Atk4\Ui\Form\Control` has native `disabled` property

PR #1797, opened 2022-06-07
Adjust atk4/data Model::tryLoad returning null

mvorisek:
related with https://github.com/atk4/data/pull/996, all `->tryLoad` usages reviewed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,9 @@
     },
     "require-release": {
         "php": ">=7.4 <8.2",
-        "atk4/data": "~4.0.0"
+        "atk4/data": "~4.0.0",
+        "symfony/filesystem": "^4.4 || ^5.3",
+        "symfony/http-foundation": "^4.4 || ^5.3"
     },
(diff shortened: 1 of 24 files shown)


demos/interactive/scroll-grid-container.php
```diff
@@ -28,7 +28,7 @@
 $g1->addActionButton('red', function ($js) {
     return $js->closest('tr')->css('color', 'red');
 });
-// THIS SHOULD GO AFTER YOU CALL addAction() !!!
+// THIS SHOULD GO AFTER YOU CALL Grid::addActionButton() !!!
```
mvorisek, line 31: @DarkSide666 why? added in https://github.com/atk4/ui/pull/635/files#diff-77da0ad02520ab833baf119292167fc528e59c732fcf530278535ce4a101c426R26

also, when the demo is scrolled, the newly loaded rows does not have the buttons:

image

also the width is broken - an SUI issue?

PR #1805, opened 2022-07-21
Fix empty string to null normalization in UI persistence

mvorisek:


mvorisek:
Hi @DarkSide666, @abbadon1334, @ibelar, @mkrecek234, what are your thoughts on typecasting empty string from UI to php value?

In data, we distinguish between an empty string and null value strictly. But in ui, an empty field cannot be differenciated from null value, as html input does not represent null value differently than empty string.

I am not completely convinced by typecasting empty string to null when typecasting the UI value to php value fully. Even with the logic when field cannot be nullable, then empty string value is retained.

What is your practice and what behaviour would you expect?

mkrecek234:
@mvorisek I fancy the approach to make empty string null where possible in UI, as I never had any distinguishing. Having to handle both null and empty is really not helpful. I never had any use case where I intentionally wanted to separate the two from a UI perspective.

DarkSide666:
I completely agree with @mkrecek234. And not just in UI, but better even in Data level.
In my opinion empty string value do not carry any useful, and other than null, information. It's not like zero or false values which have some meaning other than null.
Also using workaround to store empty string values in DB instead of null just because DB column is NOT NULL should not be a way to do this. Also some DB engines (or settings) automatically replace empty string value to NULL.

So I vote for removing empty string values (setting them as null) from all toolkit repos not just UI. It should be included in lower level normalization.

Also can check this:
- https://dev.mysql.com/doc/refman/8.0/en/problems-with-null.html

mvorisek:
> empty string value do not carry any useful, and other than null, information

That is not completely true. Consider storing a CLI param. Null represents "no value" (0 params) and empty string "empty value" (1 param).

Another example is from the MySQL docs you posted:
```
SELECT CONCAT(`first_name`, `last_name`) AS `name`;
```
will return null if at least one name is null (which will be, if both names are not non-empty and if empty strings will be converted to null by default).

> Also some DB engines (or settings) automatically replace empty string value to NULL.

This is only true for Oracle DB. And I hate it, as there is no easy workaround instead of storing empty string as a special value - https://github.com/atk4/data/blob/8d738d8a866d18dfa2d93af1819949d762623dc2/tests/Schema/MigratorTest.php#L155

This is actually the strongest argument for NOT converting empty string to null. If atk4/data should support "any data", we must support storing empty string (for nullable/not-nullable field).

> Also using workaround to store empty string values in DB instead of null just because DB column is NOT NULL should not be a way to do this. So I vote for removing empty string values (setting them as null) from all toolkit repos not just UI. It should be included in lower level normalization.

I do not like any normalization/typecast parametrization on anything else than the UI persistence config and field type either.

But as long as the case above is valid, it cannot be done in atk4/data by default.

In summary, we have two main cases:
- null is not allowed (`Field::mandatory` and/or `Field::required` is set) - then any conversion to `null` is wrong, as it will imply invalid value (and we should never convert `null` to empty string in atk4/data, in atk4/ui we can as it is impossible to distinguish between)
- null is allowed - I reach you guys again, do you propose any solution how can the "empty string to null" can be optional?

I think there are several possibilities (to impl. on atk4/data level):
- custom `emptyable_string` type - this will still "corrupt" empty string data with standard `string` DBAL type
- custom Field flag `convertEmptyStringToNull` - this will be easy to understand, but the flag will be string specific and not aplicable on other DBAL types
- set `Field::mandatory` to `true` by default - this is default in doctrine/orm - but when field is explicitly allowed to be null, we are at the beginning (this PR for UI, but with data field not-nullable by default, thus "empty input" can be converted to "empty string" only)

I am not inclined to any solution at this moment, but some ideas can be rejected for the presented bad implications.

I am slightly in favor of actually making every model field mandatory by default. Non-nullable field does not hold this ambiguity and the developer have to check only for non-emptiness.

mvorisek:
This PR can solve some usecases, but will break other. We should actually add a support for 3-state strings - null, empty and non-empty by adding a checkbox to the field to be able to specify null if nullable.

mvorisek:
reopened purely to rebase on develop

PR #1810, opened 2022-08-02
Unify Control readOnly property case to camelCase

mkrecek234:
change readonly to readOnly

### DIFF
--- a/demos/form-control/input2.php
+++ b/demos/form-control/input2.php
@@ -14,7 +14,7 @@ use Atk4\Ui\HtmlTemplate;
 /** @var \Atk4\Ui\App $app */
 require_once __DIR__ . '/../init-app.php';
 
-\Atk4\Ui\Header::addTo($app, ['Disabled and read only form controls (normal / readOnly / disabled)']);
+\Atk4\Ui\Header::addTo($app, ['Disabled and read only form controls (normal / read only / disabled)']);
 
 $form = Form::addTo($app);
 
--- a/demos/form-control/input2.php
+++ b/demos/form-control/input2.php
@@ -14,24 +14,24 @@ use Atk4\Ui\HtmlTemplate;
 /** @var \Atk4\Ui\App $app */
 require_once __DIR__ . '/../init-app.php';
 
-\Atk4\Ui\Header::addTo($app, ['Disabled and read only form controls (normal / readonly / disabled)']);
+\Atk4\Ui\Header::addTo($app, ['Disabled and read only form controls (normal / readOnly / disabled)']);
 
 $form = Form::addTo($app);
 
 // Test all kinds of input fields
 $group = $form->addGroup('Line');
 $group->addControl('line_norm')->set('editable');
-$group->addControl('line_read', ['readonly' => true])->set('read only');
+$group->addControl('line_read', ['readOnly' => true])->set('read only');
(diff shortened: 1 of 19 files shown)


demos/form-control/input2.php
```diff
@@ -14,24 +14,24 @@
 /** @var \Atk4\Ui\App $app */
 require_once __DIR__ . '/../init-app.php';
 
-\Atk4\Ui\Header::addTo($app, ['Disabled and read only form controls (normal / readonly / disabled)']);
+\Atk4\Ui\Header::addTo($app, ['Disabled and read only form controls (normal / readOnly / disabled)']);
```
mvorisek: Where does `normal` come from?

To the SUI/FUI case - we should use the canonical form from SUI/FUI docs.

src/Form/Control.php
```diff
@@ -70,7 +70,7 @@ class Control extends View
      *
      * @var bool
      */
-    public $readonly = false;
+    public $readOnly = false;
```
mvorisek: This property is definitely `bool` only, please strong type it, also `disabled`.

demos/form-control/input2.php
mkrecek234: @mvorisek This is just a description for the demo to show normal controls, or readonly or disabled controls. No code implication.

mvorisek: then the original should stay or it should be changed to "read only"

mvorisek:
@mkrecek234 what is your opinion on `readonly`? It seems it is one word keyword used in LC across many languages (PHP, C#, TS, ...).

mkrecek234:
@mvorisek you can argue that "readonly" is common for many languages, but many languages don't camelCase. Thus, I clearly prefer the camelCase as it is according to the rules. In English, proper writing is "read-only", which also supports to camelCase it. Any other opinion @abbadon1334 @ibelar ? Not a major topic, but let's make ATK4 even more beautiful ;-)

src/Form/Control/Dropdown.php
```diff
@@ -136,7 +136,7 @@ public function getInput()
             'type' => $this->inputType,
             'id' => $this->name . '_input',
             'value' => $this->getValue(),
-            'readonly' => $this->readonly ? 'readonly' : false,
+            'readOnly' => $this->readOnly ? 'readonly' : false,
```
mvorisek: this is definitely wrong, `readonly` is canonical form of HTML attribute

mvorisek:
> @mvorisek you can argue that "readonly" is common for many languages, but many languages don't camelCase. Thus, I clearly prefer the camelCase as it is according to the rules. In English, proper writing is "read-only", which also supports to camelCase it. Any other opinion @abbadon1334 @ibelar ? Not a major topic, but let's make ATK4 even more beautiful ;-)

In atk4/* the camelCase naming is defined by the "most natural snake_case". Both "read_only" and "readonly" variants make sense (vs. "checkbox", almost noone would write "check_box").

I have done more reseach on "readonly" vs "readOnly" with Google. When I was renaming the property in atk4/data, I did not think much of it and simply converted the snake_case to camelCase. The case used in other projects in quite mixed, LC is used mainly when it is a "language attribute" which is almost always LC (case insensitive) only. When used as a property name, they are projects which use it like proposed. To unify the naming with atk4/data (atk4 ecosystem in general), I am ok with this change.

PR #1816, opened 2022-08-18
Fix '0' instead of '' default value for Lookup

### LINKED ISSUE #1813: Lookup control writes 0 instead of NULL for nullable field
- You have any Lookup for a field which is nullable in a Form with a connected Model (e.g. via hasOne reference)
- You have selected an entry of the Lookup and saved.
- You then select the empty entry of that Lookup and save again

Result: Atk4/Data tries to store 0, instead of NULL. This is wrong, it has to be NULL.
Sorry can't provide a quick demo, but this should be straightforward.
Funnily, in Dropdown control it works, only in Lookup not.

mvorisek:
fixes https://github.com/atk4/ui/issues/1813

### DIFF
--- a/demos/_includes/Demo.php
+++ b/demos/_includes/Demo.php
@@ -18,6 +18,7 @@ class Demo extends \Atk4\Ui\Columns
     protected function init(): void
     {
         parent::init();
+
         $this->addClass('celled');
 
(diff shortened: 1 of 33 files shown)


mvorisek:
@mkrecek234 thank you for the great repro code!

PR #1818, opened 2022-08-20
Fix Multiline /w SQL expr

mvorisek:


mvorisek:
currency (/w `develop` branch/Array persistence) the total does not work also when the `# of Boxes` is set `0`, for some reasons, it is assumed as `1`, example:

image

Issue #1728, opened 2022-01-20
Activate Accordion or Tab on validation error


mkrecek234:
I am using heavily Acccordion sections to structure forms (e.g. Customer, Address, Payment, Shipping, Tax) - the way I do this is to make one form, and inside the form we have the Accordions.

However if a Validation for a field inside a closed Accordion fails, it seems to the user nothing happens on Submit. Can we automatically open the Accordion(s) where fields with failed Validation messages are present?

mvorisek:
Is there any current demo on which this issue can be reproduced or can you post a minimalistic code to reproduce?

mkrecek234:
@mvorisek No, there is not yet an example but it is very simple: If you have a form with an Accordion layout inside, and one field does not validate, the form is reloaded on save, the error message is attached to the failing field, but the user can't see it if the field is in one of the (closed) Accordions. Very confusing for the user. This also would apply to tabs I assume.

Improved behaviour: If validation fails, for the first invalid field that is inside an Accordion, that Accordion is activated automatically.

@ibelar something for you eventually?

mvorisek:
Accordion Section activate on error is/was supported and was fixed by https://github.com/atk4/ui/pull/1863.

This issue is about general support. A test for Accordion and Tab should be added. And both must be activated even if insined nested structure, like Accordion->Tab->Tab->Accordion->section with an error.

mvorisek:
https://app.codecov.io/gh/atk4/ui/pull/2032/blob/src/Form/Layout/Section/Accordion.php - needs to be covered by a Behat test

Issue #1821, opened 2022-08-20
Accordion Section is unattached after form is saved in demo


mvorisek:
Steps to reproduce:

1. open `/demos/form/form-section.php`
2. click `Save` in 2nd form `Accordion Section in Form`
3. click `Section 2` - for 1st time, section is expanded
4. repeat step 3. - for this time, section is not expanded at all

it seems the problem exists in 1st click/section expand already, console output:

![image](https://user-images.githubusercontent.com/2228672/185765943-c98f3e78-f6a1-4c2c-8f38-fafab0856eb3.png)

![image](https://user-images.githubusercontent.com/2228672/185765963-005beb59-474d-4f10-b0ef-cd5f11976c1e.png)

mvorisek:
fixed/tested in Fomantic 2.9.0

PR #1679, opened 2021-10-11
Add dynamic return type seed support

mvorisek:

PR #1835, opened 2022-08-28
Mark all parent accordions/tabs if validation exception of control inside

mkrecek234:
Introduces  #1728 

Demo can be seen by opening `demos/form/form.php` demo, pressing save on the second form ("but very flexibly"). The field `status integer not nullable` will then not validate and you can see that the Accordion is automatically opened.

https://github.com/atk4/ui/blob/621d33fb35e563e3154a067d4fb2f1ec997131f2/demos/form/form.php#L60

### DIFF
--- a/demos/form/form.php
+++ b/demos/form/form.php
@@ -38,6 +38,7 @@ $tab = $tabs->addTab('Basic Use');
 Header::addTo($tab, ['Very simple form']);
 
 $form = Form::addTo($tab);
+
 $form->addControl('email');
 $form->onSubmit(function (Form $form) {
     // implement subscribe here
@@ -56,13 +57,16 @@ $group->addControl('name');
 $group->addControl('surname');
 $group->addControl('gender', [Form\Control\Dropdown::class, 'values' => ['Female', 'Male']]);
 
+$accordionLayout = $form->layout->addSubLayout([\Atk4\Ui\Form\Layout\Section\Accordion::class]);
+$details = $accordionLayout->addSection('Validating Field in Accordion');
+
 // testing 0 value
 $values = [0 => 'noob', 1 => 'pro', 2 => 'dev'];
 $form->addControl('description', [Form\Control\Textarea::class])->set(0);
 $form->addControl('no_description', [Form\Control\Textarea::class])->set(null);
 $form->addControl('status_optional', [Form\Control\Dropdown::class, 'values' => $values]);
-$form->addControl('status_string_not-nullable', [Form\Control\Dropdown::class], ['type' => 'string', 'values' => $values, 'nullable' => false]);
-$form->addControl('status_integer_not-nullable', [Form\Control\Dropdown::class], ['type' => 'integer', 'values' => $values, 'nullable' => false]);
+$details->addControl('status_string_not-nullable', [Form\Control\Dropdown::class], ['type' => 'string', 'values' => $values, 'nullable' => false]);
+$details->addControl('status_integer_not-nullable', [Form\Control\Dropdown::class], ['type' => 'integer', 'values' => $values, 'nullable' => false]);
 $form->addControl('status_string_required', [Form\Control\Dropdown::class], ['type' => 'string', 'values' => $values, 'required' => true]);
 $form->addControl('status_integer_required', [Form\Control\Dropdown::class], ['type' => 'integer', 'values' => $values, 'required' => true]);
 
--- a/demos/form/form.php
+++ b/demos/form/form.php
@@ -38,7 +38,6 @@ $tab = $tabs->addTab('Basic Use');
 Header::addTo($tab, ['Very simple form']);
 
 $form = Form::addTo($tab);
-
 $form->addControl('email');
 $form->onSubmit(function (Form $form) {
     // implement subscribe here
@@ -57,7 +56,7 @@ $group->addControl('name');
 $group->addControl('surname');
 $group->addControl('gender', [Form\Control\Dropdown::class, 'values' => ['Female', 'Male']]);
 
-$accordionLayout = $form->layout->addSubLayout([\Atk4\Ui\Form\Layout\Section\Accordion::class]);
+$accordionLayout = $form->layout->addSubLayout([Form\Layout\Section\Accordion::class]);
 $details = $accordionLayout->addSection('Validating Field in Accordion');
 
 // testing 0 value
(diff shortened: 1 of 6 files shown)


demos/form/form.php
```diff
@@ -38,6 +38,7 @@
 Header::addTo($tab, ['Very simple form']);
 
 $form = Form::addTo($tab);
+
```
mvorisek: why such extra space?

demos/form/form.php
```diff
@@ -56,13 +57,16 @@
 $group->addControl('surname');
 $group->addControl('gender', [Form\Control\Dropdown::class, 'values' => ['Female', 'Male']]);
 
+$accordionLayout = $form->layout->addSubLayout([\Atk4\Ui\Form\Layout\Section\Accordion::class]);
```
mvorisek: CS fixer is a little short here, we prefer `Form\...` import here, never FQCN, except single FQCN like `\Throwable`. You will get feeling for the CS we use by reading more code.

demos/form/form.php
```diff
@@ -56,13 +57,16 @@
...
+$details = $accordionLayout->addSection('Validating Field in Accordion');
```
mvorisek, line 60: The phpstan error is correct here and I am still not fully decided id we want https://github.com/atk4/ui/pull/1679 or not, thus add this to `phpstan.neon.dist` at the "most logical place" - around the same phpstan errors/level and file names are sorted alphabetically

src/Form.php
```diff
@@ -258,8 +258,17 @@ public function onSubmit(\Closure $callback)
                 return $response;
             } catch (ValidationException $e) {
                 $response = [];
+                $openFirstSectionOnError = false;
+
                 foreach ($e->errors as $field => $error) {
                     $response[] = $this->error($field, $error);
+
+                    // If field inside Accordion section does not validate, open first AccordionSection
+                    if (!$openFirstSectionOnError && $this->getControl($field)->getOwner()->getOwner() instanceof \Atk4\Ui\AccordionSection) {
```
mvorisek: this smells to me - `getOwner()->getOwner()` and below even triple times it seems it is single purpose code only and for more layout nesting it will not work

also, ideally it should not be handled in `Form`, as form should not care about it's childs too much

also Behat test should be added, so this code is covered - otherwise the logic might break soon - once you will satisfy all CI tests, you should see this in CodeCov report

mvorisek:
Also, what would happen if you have errors in more tabs/accordions? I would expect some error icon on the tabs/accordion menus - the "first accordion error" might be an error comming from other tab/accordion.

mkrecek234:
Certainly this is not the perfect solution, also Tabs in Forms are not covered. Do you have a better solution in mind? This is clearly a minor field which would not justify to refactor the whole rendering logic, but is a critical situation for users - if there is an error only with a closed accordion, the user does not get any feedback. This would be a pragmatic solution, and unfortunately Ui does not offer to add a badge to the closed Accordion...

If there is an error also in another Accordion, at next save it would open that second Accordion.

src/Form.php
mkrecek234: getOwner multiple times also is used in other view, like Crud. You are right, nested Accordions are then not supported - but since today it just won't work at all, this is a huge improvement to usability. And to me, nested Accordions in forms really is an edge case which would not be priority to me to support yet.

mvorisek (changes requested):
The main UX problem is user must be aware of all errors. Currently, user must click thru all sections which is definitely an issue. However, when the first section with an error is opened, the problem remains unsolved.

I would like the error badge instead which can also cascade up easily.

A badge/icon is already used in:
image
in `demos/layout/layout-panel.php` demo

Opening first error might not be even wanted, if there are more errors on more pages and user is currently on the non-first page /w error he just entered.

mkrecek234:
@mvorisek To my knowledge the Accordion Section in Fomantic UI does not easily allow to add an icon, at least not that I know of or that Atk4 supports it. Sure this would be better - in the current state however, it is a big issue - the user believes he saved, because he does not see ANY error message if only a field inside an accordion is invalid. We had these issues with users many times. Can we cut this into two steps 

1) We just open the first Accordion for now, so the error is displayed in the incorrect field 

and later, - if we there is an easy solution with Fomantic Ui to add icons to Accordion Sections - we 

2) mark all sections with errors with that icon. 

Even in the scenario of multiple sections to have errors, you want the first section to be opened automatically. One click saved, and much easier to grasp for the user what he is expected to do.

mvorisek:
I fully confirm this issue and agree it must be fixed.

I however do not agree the implementation with nested `getOwner` is valid and even belive it can crash the App, if 2/3+ owners are not defined. (imagine Form with Control only and no App/Form Layout)

As discussed in previous posts, the "opening the 1st section" solution is not a general solution and is Accordion specific. Component consistency is important and no fix should be merged until it fixes Tabs too.

Also, while testing this PR, I hit https://github.com/atk4/ui/issues/1821. So this issue should be fixed first. Repro steps: save form and then click section header to expand (you can click multiple times to open/close) and you will see messages in DOM browser console.

mkrecek234:
May be we use jQuery to find all parent accordion sections to get rid of the triple getOwner logic, see here: https://stackoverflow.com/questions/5333426/how-to-find-a-parent-with-a-known-class-in-jquery

We could also search for tabs or only for the warning icon class.

This would open also all nested parents in a row/ mark all nested parents with a warning sign ‘d.parents('.a').attr('id')‘

Or even better all parents up to/inside current form: https://api.jquery.com/parentsUntil/ so we do not go outside current form.

mvorisek:
yes, but in PHP, no need for jQuery, there are even some helper method for it present

mkrecek234:
> yes, but in PHP, no need for jQuery, there are even some helper method for it present

I know, but we could make use of the jQuery function - calling it through the Atk4 jQuery helper code. Do you like the logic in order to cover nested warning icons/tabs/accordions?

mvorisek:
> > yes, but in PHP, no need for jQuery, there are even some helper method for it present
> 
> I know, but we could make use of the jQuery function - calling it through the Atk4 jQuery helper code. Do you like the logic in order to cover nested warning icons/tabs/accordions?

yes, I meant `View::getClosestOwner` method

yes, nested, multiple errors...

mkrecek234:
Warning icon should be red as all validation errors. The orange one is used for canLeave=true on changes

PR #1839, opened 2022-08-29
Improve App::getTag()

### LINKED ISSUE #1838: Readonly/disabled Dropdown form control must not be clickable/changeable
Steps to reproduce:

1. open `demos/form-control/input2.php`
2. click on readonly or disabled `File Upload` control:
  image
3. notice a file select dialog is shown and file was changed

mvorisek:
fixes #1838

fix too many "form elements" tags introduced in https://github.com/atk4/ui/pull/1275

minor BC break implied by some no longer supported inputs


src/Form/Control/Lookup.php
```diff
@@ -378,22 +378,22 @@ protected function initDropdown($chain): void
...
-            $this->template->set('multiple', 'multiple');
+            $this->template->set('multiple', 'multiple="multiple"');
         }
 
         if ($this->disabled) {
             $this->settings['showOnFocus'] = false;
             $this->settings['allowTab'] = false;
 
-            $this->template->set('disabled', 'disabled');
+            $this->template->set('disabled', 'disabled="disabled"');
```
mvorisek, line 388: must be set as html:

image

must investigate, `"` does need to be escaped only in attributes

mvorisek: fixed, mixed tag use in Lookup template

src/Form.php
```diff
@@ -532,13 +532,12 @@ protected function renderTemplateToHtml(string $region = null): string
...
-        return $this->fixFormInRenderedHtml($output);
+        return $this->fixOwningFormAttrInRenderedHtml($output);
     }
 
-    public function fixFormInRenderedHtml(string $html): string
+    public function fixOwningFormAttrInRenderedHtml(string $html): string
     {
-        $innerFormTags = ['button', 'datalist', 'fieldset', 'input', 'keygen', 'label', 'legend',
-            'meter', 'optgroup', 'option', 'output', 'progress', 'select', 'textarea', ];
+        $innerFormTags = ['button', 'fieldset', 'input', 'output', 'select', 'textarea'];
```
mvorisek: all other removed tags were added wrongly, they have no `form` nor `for` attribute

`label` is non-value tag /w `for` attribute, removing now, hopefully not needed - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

PR #1840, opened 2022-08-29
Fix coverage static analysis cache and move util for reuse

mvorisek:
related upstream performance fixes:
- https://github.com/sebastianbergmann/php-code-coverage/pull/928
- https://github.com/sebastianbergmann/php-code-coverage/pull/929
- https://github.com/sebastianbergmann/php-code-coverage/pull/930

reduces Unit/Behat testing times /w coverage by about 3x

Unit time: before ~6 mins, now ~2 mins
Behat time: before ~4.5 mins, now ~2 mins


codecov.yml
```diff
@@ -1,14 +1,9 @@
...
-  - docs
-  - tools
 comment: false
 coverage:
   status:
     project:
       default:
         target: auto
-        threshold: 0.025
+        threshold: 0.015
```
mvorisek, line 7: 0.015% equals 1 uncovered line pass, but 2 emit an CI error badge

non-zero is needed to prevent rounding/random errors by Covecov

mvorisek:
this PR can possibly improve also the random Behat failures, but still does not fix all of them, needs deeper investigation why they happen

PR #1841, opened 2022-08-29
Fix 1836 AccordionSection activate

### LINKED ISSUE #1836: In forms, Accordion->activate(...) stopped working
For Accordions in forms, the activate function to open Accordions on load worked until very recently (must have been introduced only max. in the last 7-10 days, as this commit still worked f8278d6934559eb8efd8f86013f39cc95bdb1475

This stopped working now, see demo - the first Accordion should open:

https://github.com/atk4/ui/blob/a120f0f4d1593a4dd10b8b3ad32668c6107d191b/demos/form/form-section-accordion.php#L66

mvorisek:
fix #1836

the problem was introduced in removed method override https://github.com/atk4/ui/commit/01d39d24694e4b4f63a576d090e55e0e97aaa9e4#diff-08f1560acb847ce226eb00aca9621dc3412618092cd61be204df26aaf98a7ecaL62

investigate the created View structure and preferably come up with a solution that does not depend on a nesting level

a Behat test would be great too

### DIFF
--- a/src/Accordion.php
+++ b/src/Accordion.php
@@ -136,11 +136,9 @@ class Accordion extends View
     /**
      * Return the index of an accordion section in collection.
      *
-     * @param AccordionSection $section
-     *
      * @return int
      */
-    public function getSectionIdx($section)
+    public function getSectionIdx(AccordionSection $section)
     {
         $idx = -1;
         foreach ($this->sections as $key => $accordion_section) {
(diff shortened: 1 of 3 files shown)


mkrecek234:
Warning icon should be red as all validation errors. The orange one is used for canLeave=true on changes

PR #1845, opened 2022-08-31
Fix CardDeck action /w entity

### LINKED ISSUE #1844: CardDeck default UserAction not working
Steps to reproduce:
- add default CardDeck to any view
- set any model to CardDeck
- Press default UserAction 'Add' and enter entity

Error:
- Expected entity, but instance is a model
https://github.com/atk4/ui/blob/a8a08cfed516c5ffc1e7e0c54db8ab3e8122b0c4/src/CardDeck.php#L262

and here

https://github.com/atk4/ui/blob/a8a08cfed516c5ffc1e7e0c54db8ab3e8122b0c4/src/CardDeck.php#L296

Replacing $action->getModel() by $action->getEntity() also won't fix it...

mkrecek234:
fix #1844

### DIFF
--- a/demos/collection/card-deck.php
+++ b/demos/collection/card-deck.php
@@ -47,6 +47,6 @@ $infoAction->args = [
     'country' => ['required' => true, 'ui' => ['form' => [Form\Control\Lookup::class, 'model' => new Country($app->db), 'placeholder' => 'Please select a country.']]],
 ];
 
-$deck = CardDeck::addTo($app, ['noRecordScopeActions' => ['request_info'], 'singleScopeActions' => ['book']]);
+$deck = CardDeck::addTo($app);
 
 $deck->setModel($countries, ['Cost'], [$countries->fieldName()->iso, $countries->fieldName()->iso3]);
--- a/demos/collection/card-deck.php
+++ b/demos/collection/card-deck.php
@@ -47,6 +47,6 @@ $infoAction->args = [
     'country' => ['required' => true, 'ui' => ['form' => [Form\Control\Lookup::class, 'model' => new Country($app->db), 'placeholder' => 'Please select a country.']]],
 ];
 
-$deck = CardDeck::addTo($app);
+$deck = CardDeck::addTo($app, ['noRecordScopeActions' => ['request_info'], 'singleScopeActions' => ['book']]);
 
 $deck->setModel($countries, ['Cost'], [$countries->fieldName()->iso, $countries->fieldName()->iso3]);
--- a/demos/init-db.php
+++ b/demos/init-db.php
@@ -48,10 +48,14 @@ trait ModelPreventModificationTrait
     {
         $originalCallback = $action->callback;
         $action->callback = function (Model $model, ...$args) use ($action, $originalCallback, $outputCallback) {
+            if ($model->isEntity()) {
+                $action = $action->getActionForEntity($model);
+            }
+
             $callbackBackup = $action->callback;
             try {
                 $action->callback = $originalCallback;
-                $action->execute($model, ...$args);
+                $action->execute(...$args);
             } finally {
                 $action->callback = $callbackBackup;
             }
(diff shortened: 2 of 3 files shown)


demos/init-db.php
```diff
@@ -133,7 +133,7 @@ protected function init(): void
 
         parent::init();
 
-        $this->initPreventModification();
+        //$this->initPreventModification();
```
mvorisek: ⚠️ THIS MUST BE REVERTED BEFORE MERGE

We want to keep no-save beaviour for local too, as it makes the demos stateless, which significantly helps while debugging some issue.

I will check why this demo needed this uncommented to reproduce the issue, maybe we can improve this for the UserAction somehow.

src/CardDeck.php
```diff
@@ -293,9 +293,9 @@ protected function findCard(Model $model)
             }
         }
 
-        if (in_array($model->getId(), array_map($mapResults, $model->export([$model->idField])), true)) {
+        if (in_array($entity->getId(), array_map($mapResults, $entity->getModel()->export([$entity->idField])), true)) {
```
mvorisek: I know you are proposing a bug fix and it looks right. But exporting all model rows in order to find out if an ID is present is horrible!

mkrecek234: findCard was coded like that before, so if you want to change the logic, please propose how. Have to be focused on fixing broken things due to the changes in atk4 due to time-constraints

demos/init-db.php
mvorisek: done

src/CardDeck.php
mvorisek: done + checked if there is no simillar bad usage

src/CardDeck.php
```diff
@@ -255,11 +255,11 @@ protected function getNotifier(string $msg = null, Model\UserAction $action = nu
     /**
      * Js expression return when action afterHook executor return a Model.
      */
-    protected function jsModelReturn(Model\UserAction $action = null, string $msg = 'Done!'): array
+    protected function jsModelReturn(Model\UserAction $action = null, Model $returnEntity, string $msg = 'Done!'): array
```
mvorisek: This is complex topic if UA should be immutable in sense of model of if it should be cloned when fetched on entity - related PR https://github.com/atk4/data/pull/943.

As there is no clear solution not direction now, I will accept this change now.

demos/collection/card-deck.php
```diff
@@ -47,6 +47,6 @@
     'country' => ['required' => true, 'ui' => ['form' => [Form\Control\Lookup::class, 'model' => new Country($app->db), 'placeholder' => 'Please select a country.']]],
 ];
 
-$deck = CardDeck::addTo($app, ['noRecordScopeActions' => ['request_info'], 'singleScopeActions' => ['book']]);
+$deck = CardDeck::addTo($app);
```
mvorisek: I will revert this repro code not because it is solely bad, but we need it for more coverage, otherwise `noRecordScopeActions` would be untested.

src/CardDeck.php
mvorisek: actually this should use `getActionForEntity` somewhere, otherwise the UA could be for a different model - related https://github.com/atk4/data/pull/947

mvorisek: done

mvorisek (changes requested):
there are 2 problems:

- RO CardDeck demo does not reload after record is added
- (from chat, unverified) jsExecute is not executed upon edit or delete, only on add

mvorisek:
> RO CardDeck demo does not reload after record is added

this is a missing feature, opened https://github.com/atk4/ui/issues/1848, not easy to fix

> (from chat, unverified) jsExecute is not executed upon edit or delete, only on add

not sure what you meant @mkrecek234, the `jsExecute` method should be executed, it has the same `Model\UserAction::APPLIES_TO_SINGLE_RECORD` type as `add` action - no need for further investigation, the whole execute -> ajax update part must be reworked for https://github.com/atk4/ui/issues/1848

merging, as the subjected https://github.com/atk4/ui/issues/1844 issue is fixed

PR #1847, opened 2022-08-31
Add support to add label to menu group

mkrecek234:
Situation:
Today, only icons are supported for menu groups. Menu items however in Admin or Maestro layout can show either an icon or a label.

PR:
Introduce to add label also for menu groups.

Sample Usage:
`$layout->addMenuGroup(['Tickets', 'label' => ['1', 'class.red' => true]]);`

Look:
image

Caveat: It would be better if the items were right-aligned (the icon or the label) - for menu items (see below), they are right aligned, so the label spreads out to the left. If anyone has a hint, how to set classes for that please contribute.

### DIFF
--- a/src/Menu.php
+++ b/src/Menu.php
@@ -131,6 +131,8 @@ class Menu extends View
 
         if (isset($name['icon'])) {
             Icon::addTo($group, [$name['icon']], ['Icon'])->removeClass('item');
+        } elseif (isset($name['label'])) {
+            Label::addTo($group, [$name['label']], ['Icon'])->removeClass('item')->addClass('mini');
         }
 
         return $group;
--- a/src/Menu.php
+++ b/src/Menu.php
@@ -132,7 +132,7 @@ class Menu extends View
         if (isset($name['icon'])) {
             Icon::addTo($group, [$name['icon']], ['Icon'])->removeClass('item');
         } elseif (isset($name['label'])) {
-            Label::addTo($group, [$name['label']], ['Icon'])->removeClass('item')->addClass('mini');
+            Label::addTo($group, [$name['label']], ['Icon'])->removeClass('item')->addClass('tiny');
         }
 
         return $group;
--- a/src/Menu.php
+++ b/src/Menu.php
@@ -91,7 +91,7 @@ class Menu extends View
             $name = [$name];
         }
 
-        $label = $name['title'] ?? $name['text'] ?? $name['name'] ?? $name[0] ?? null;
+        $label = $name['title'] ?? $name[0] ?? null;
 
         if ($label !== null) {
             $subMenu->template->set('label', $label);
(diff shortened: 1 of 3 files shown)


src/Menu.php
```diff
@@ -121,6 +121,8 @@ public function addGroup($name, string $template = 'menugroup.html')
 
         if (isset($name['icon'])) {
             Icon::addTo($group, [$name['icon']], ['Icon'])->removeClass('item');
+        } elseif (isset($name['label'])) {
+            Label::addTo($group, [$name['label']], ['Icon'])->removeClass('item')->addClass('mini');
```
mvorisek: you say this is supported somewhere and somewhere not - can we dedup the code at least with private method?

mvorisek (changes requested):
need at least a test in demos - to cover the added lines

mkrecek234:
It is supported for menu items, but not for groups - groups were added outside of Fomantic UI, whereas menuitems are rendered based on standard FUI. Will work on demo

src/Menu.php
```diff
@@ -123,14 +123,16 @@ public function addGroup($name, string $template = 'menugroup.html')
...
+        $title = $name['title'] ?? $name[0] ?? null;
 
         if ($title !== null) {
             $group->template->set('title', $title);
         }
 
         if (isset($name['icon'])) {
             Icon::addTo($group, [$name['icon']], ['Icon'])->removeClass('item');
+        } elseif (isset($name['label'])) {
+            Label::addTo($group, [$name['label']], ['Icon'])->removeClass('item')->addClass('tiny');
```
mvorisek, line 135: for menu item the size is standard (class is unspecified) - ie. the font size is 12 px

for menu group, the size should not be smaller (tiny is 9 px)

related with https://github.com/atk4/ui/pull/1847/files#r1548874861 which should be solved first/instead

src/Menu.php
```diff
@@ -123,14 +123,16 @@ public function addGroup($name, string $template = 'menugroup.html')
...
-        $title = $name['title'] ?? $name['text'] ?? $name['name'] ?? $name[0] ?? null;
+        $title = $name['title'] ?? $name[0] ?? null;
 
         if ($title !== null) {
             $group->template->set('title', $title);
         }
 
         if (isset($name['icon'])) {
             Icon::addTo($group, [$name['icon']], ['Icon'])->removeClass('item');
+        } elseif (isset($name['label'])) {
```
mvorisek, line 134: > Caveat: It would be better if the items were right-aligned (the icon or the label) - for menu items (see below), they are right aligned, so the label spreads out to the left. If anyone has a hint, how to set classes for that please contribute.

The solution is probably to implement the "menu group item" the same way as "menu item", currently the HTML/CSS for "menu group item" is using CSS/FUI columns, but "menu item" right float positioning.

mvorisek: in https://github.com/atk4/ui/pull/1847/commits/93ac62f1aa8a1bd2678138e2836d9fb99bbe55c5 I have commit debug code to debug/unify this

also IMO Maestro layout should be integrated into parametrized Admin layout

PR #1854, opened 2022-09-05
Fix Reference::getTheirFieldName getter usage

### LINKED ISSUE #1853: ScopeBuilder with hasOne reference not working
Steps to reproduce:

1. Define two models with MySQL persistence, e.g. customer and agent, where contract has a hasOne reference to agent.
2. The hasOne reference needs to be set like this: `$this->hasOne('agent_id', ['model' => [Agent::class], 'required' => true])->addTitle()`;
2. Create a view with a form `$form` and add a ScopeBuilder to it.

When rendering, I receive this error:
image

If you set the hasOne reference like this:
`$this->hasOne('agent_id', ['model' => [Agent::class], 'theirField' => 'id', 'required' => true])->addTitle();`
the exception won't be triggered. I think atk4/data should make sure to continue to automatically assume theirField to be `id` if it is not set.

Inside getLookupProps when calling getFieldItems and then `$items[$item->get($field->getReference()->getTheirFieldName())] = ...` , the $this->model is an array and not a model object, so inside `getTheirFieldName` this call runs an exception: `Model::assertInstanceOf($this->model)` .

https://github.com/atk4/ui/blob/32e2742bd6f7787ffc40a2c8c6dd98312360371a/src/Form/Control/ScopeBuilder.php#L478

Here is the content of $this->model:
image

mvorisek:
fix #1853

@mkrecek234 please test

### DIFF
--- a/src/Form/Control/Multiline.php
+++ b/src/Form/Control/Multiline.php
@@ -563,7 +563,7 @@ class Multiline extends Form\Control
     public function setLookupOptionValue(Field $field, string $value): void
     {
         $model = $field->getReference()->refModel($this->model);
-        $entity = $model->tryLoadBy($field->getReference()->getTheirFieldName(), $value);
+        $entity = $model->tryLoadBy($field->getReference()->getTheirFieldName($model), $value);
         if ($entity !== null) {
             $option = [
                 'key' => $value,
@@ -627,7 +627,7 @@ class Multiline extends Form\Control
             $model->setLimit($limit);
 
             foreach ($model as $item) {
-                $items[$item->get($field->getReference()->getTheirFieldName())] = $item->get($model->titleField);
+                $items[$item->get($field->getReference()->getTheirFieldName($model))] = $item->get($model->titleField);
             }
         }
 
--- a/src/Form/Control/ScopeBuilder.php
+++ b/src/Form/Control/ScopeBuilder.php
@@ -475,7 +475,7 @@ class ScopeBuilder extends Form\Control
             $model->setLimit($limit);
 
             foreach ($model as $item) {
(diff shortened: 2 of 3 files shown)


mkrecek234:
@mvorisek The bug above seems fixed, ScopeBuilder however creates a new bug then with a boolean field (e.g., is_deleted) in this line of code:  https://github.com/atk4/ui/blob/76755a88862922902fe83b297ed3511b6c4e3c08/src/Form/Control/ScopeBuilder.php#L682

throwing: 
image

:: = {Atk4\Ui\Form\Control\ScopeBuilder} 
$condition = {Atk4\Data\Model\Scope\Condition} 
$inputType = "text"
$inputsMap = {string[78]} ["number", "number", "number", "date", "date", +73 more]
$operator = "equals"
$operatorKey = "equals"
$operatorsMap = {string[18]} ["=", "!=", ">", ">=", "<", +13 more]
$rule = "is_deleted"
$value = false
$_COOKIE = {string[1]} ["1eqggjdoafr1hg2..."]
$_SERVER = {array[32]} 
$GLOBALS = {array[13]}

mkrecek234:
I had another similar issue where a boolean to string typecasting did not work anymore, see issue #1065 https://github.com/atk4/data/issues/1065

mvorisek:
> @mvorisek The bug above seems fixed, ScopeBuilder however creates a new bug then with a boolean field (e.g., is_deleted) in this line of code...

I tried to reproduce with `boolean` field and hit another ScopeBuilder issue. I filed https://github.com/atk4/ui/issues/1855 issue for it. Probably some issue with (no to ui) typecasting. But the exception is different than you posted here, so please post on the issue to reproduce the exception you get.

This PR fixes https://github.com/atk4/ui/issues/1853 and I am merging it.

PR #1859, opened 2022-09-11
Fix click event bubbling in Table::onRowClick() II.

### LINKED ISSUE #1858: New onRowClick implementation stops action buttons and dropdowns from working
With further changes, the onRowClick implementation renders action buttons in tables inactive, as well as action buttons in dropdowns. 

This line has to be adjusted        
https://github.com/atk4/ui/blob/ebbd3b952b44669d7793fa45e8995783ea17134c/src/Table.php#L513

to 
```
 $this->js(true)->find('a, .ui.input, .ui.checkbox, .atk4-norowclick')->on(
```
- so remove .ui.button and .ui.dropdown please.

mkrecek234:
fixes #1858

related original https://github.com/atk4/ui/pull/1655 PR

### DIFF
--- a/demos/_unit-test/grid-rowclick.php
+++ b/demos/_unit-test/grid-rowclick.php
@@ -8,6 +8,7 @@ use Atk4\Ui\Grid;
 use Atk4\Ui\JsExpression;
 use Atk4\Ui\JsFunction;
 use Atk4\Ui\JsToast;
+use Atk4\Ui\Message;
 use Atk4\Ui\Table;
 
 /** @var \Atk4\Ui\App $app */
@@ -18,6 +19,18 @@ $grid = Grid::addTo($app, ['name' => 'grid']);
 $grid->setModel($model);
 
 $grid->addDecorator($model->fieldName()->name, [Table\Column\Link::class, 'url' => 'xxx']);
+
+$grid->addActionButton('Action Button', function () {
+    return new JsToast(['message' => 'Clicked Action Button']);
+});
+
+$grid->addActionMenuItem('Action MenuItem', function () {
+    return new JsToast(['message' => 'Clicked Action MenuItem']);
+});
+
+$grid->addModalAction('Action Modal', 'Details', function ($p, $id) use ($model) {
+    Message::addTo($p, ['Clicked Action Modal: ' . $model->load($id)->name]);
+});
 
 $grid->table->onRowClick(function () {
     return new JsToast(['message' => 'Clicked on row']);
--- a/demos/_unit-test/grid-rowclick.php
+++ b/demos/_unit-test/grid-rowclick.php
@@ -40,6 +40,7 @@ $grid->addSelection();
 
 // emulate navigate for <a> for Behat
 // TODO emulate for all tests automatically in our Atk4\Ui\Behat\Context
+// this emulation is not perfect, as it works even with event.preventDefault() called
 $grid->table->js(true)->find('a')->on(
     'click',
     new JsFunction(['event'], [new JsExpression('window.location.href = \'#test\'; event.preventDefault();')])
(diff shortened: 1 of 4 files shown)


mvorisek:
I do not belive this PR fixes the problem correctly, what will happen if you will add checkbox somewhere instead of a button? Why to keep checkbox and remove dropdown?

also, how can this be the most easily reproduced on `demos/_unit-test/grid-rowclick.php` demo?

mkrecek234:
> I do not belive this PR fixes the problem correctly, what will happen if you will add checkbox somewhere instead of a button? Why to keep checkbox and remove dropdown?
> 
> also, how can this be the most easily reproduced on `demos/_unit-test/grid-rowclick.php` demo?

I tested all of that in live settings - checkbox works perfectly - so do href links. The action buttons with this code here also work without issues, whereas in the current develop they don't work. Let's please merge this. You can easily reproduce if you add action buttons to your repro code while also adding onRowClick.

mvorisek:
I am not objecting the issue, but missing test and I am really worried why `.ui.checkbox` is ok, but `.ui.dropdown` not.

mkrecek234:
> I am not objecting the issue, but missing test and I am really worried why `.ui.checkbox` is ok, but `.ui.dropdown` not.

The dropdown event logic seems to be different and I have not really grasped it - if you stop propagating  on .dropdown or .atk-menu-actions or the like, the actions dropdown menu items won't work. If you don't do it, they work, and so does clicking on the down-arrow. However, if you click to the "Actions..." text left of the arrow, it will fire also the onRowClick, Same happens if you add the td.collapsing. Looks as if the event logic for that is run a bit different than usual. Let's please merge this for now, as today we are facing  a more severe bug. We can keep the issue open though.

mvorisek:
Michael, in https://github.com/atk4/ui/pull/1655 I have put a lot of efford to cover the original issue with a test. The only positive side of your PR now is it does not break the tests, but the unanswered questions here are major and will probably lead to another issue.

What needs to be done is to reproduce the new issue with a test. Otherwise the problem can be reintroduced very soon.

The currently proposed solution seems "coded to satisfy tests and my usecase". If such hacky solution is needed, it will need a test for every kept/removed selector/UI component. I doubt so, if we cannot find a good solution quickly, we can revert https://github.com/atk4/ui/pull/1655 and recreate the original PR with a test for this issue.

mkrecek234:
This is why I suggest to adjust to the original proposed which did not include dropdown and button. I invested some time to get JS events for the menu item but have not found a solution. As this is a minor issue and we have many more critical things to address we should not try to rework the whole action button JS side now. We can safely keep support for links and checkout though, perfectly working.

mkrecek234:
Suggestion - the original issue was for Checkbox and Link, and also the original PR - both issues were solved successfully.
You provided tests for those uses cases, so PR should be complete and merged that way.

Since you added two more selectors (button and dropdown) which were untested and which actually created a problem which was not present before, we should keep this in a separate issue open and address it separately. The JS logic as said before for menu items in Table is different (and addresses to the most part non-bubbling events) and we can look at it separately.

mvorisek:
> Suggestion - the original issue was for Checkbox and Link, and also the original PR - both issues were solved successfully. You provided tests for those uses cases, so PR should be complete and merged that way.

Yes, having the tests has shown the full behaviour and if the problem is impossible to solve, with a little documentation in test & GH issue, it can be merged. But, I do not think it is impossible :)

In
https://github.com/atk4/ui/blob/e10535033fd57f2bb75034b262dfe8fa4a139062/src/Table.php#L518
we can probably exclude the unwanted events with https://stackoverflow.com/questions/5420399/determine-the-elements-which-an-event-has-bubbled-through no need to even add the handlers to excluded the problematic selectors.

I am also quite curious why the menu/dropdown button is allowed to bubble in the first place, as when any event is added
https://github.com/atk4/ui/blob/a036f67947e285f90ff7b7e0e3fc37fa0c9e7acb/src/View.php#L999-L1000
it should prevent it to bubble. Maybe no our event is added (only SUI adds it).

PR #1863, opened 2022-09-20
Fix nested accordion open

### LINKED ISSUE #1693: Nested accordions static level 2. not works in demo
Nested accordions static text
level 2. not works 

https://ui.agiletoolkit.org/demos/interactive/accordion-nested.php

mvorisek:
fix #1693
fix (partly - Accordion only) #1728


src/Form/Layout/Section/Accordion.php
```diff
@@ -30,7 +30,7 @@ protected function init(): void
             $jsError = [$form->js()->form('add prompt', $fieldName, $str)];
 
             // if a form control is part of an accordion section, it will open that section.
-            $section = $form->getClosestOwner($form->getControl($fieldName), AccordionSection::class);
+            $section = $form->getControl($fieldName)->getClosestOwner(AccordionSection::class);
```
mvorisek, line 33: @mkrecek234 the `getClosestOwner` method was broken, does this fix https://github.com/atk4/ui/issues/1728 too?

src/Accordion.php
```diff
@@ -152,22 +151,15 @@ public function getSectionIdx(AccordionSection $section)
...
-    }
-
     protected function renderView(): void
     {
         if ($this->type) {
             $this->addClass($this->type);
         }
 
-        // Only set Accordion in Top container. Otherwise Nested accordion won't work.
-        if (!$this->getClosestOwner($this, AccordionSection::class) && !$this->isDynamicSection()) {
```
mvorisek, line 170: @ibelar why `isDynamicSection` was introduced, it is still needed?

src/Form/Layout/Section/Accordion.php
mvorisek: @mkrecek234 should be fixed when not nested

PR #1864, opened 2022-09-24
Fix ID executor input UI typecast

### LINKED ISSUE #1825: CRUD edit/delete ID is not property casted
Steps to reproduce:

1. open `/demos/collection/crud3.php` demo
2. click edit or delete icon on some row
3. observe error:

image

I belive this is because of no/wrong typecasting, as Array persistence expects strictly typed ID (int instead of string)

mvorisek:
fix #1825

optimal test would be CRUD with datetime/object ID


js/src/agile-toolkit-ui.js
```diff
@@ -1,5 +1,5 @@
 import 'core-js/stable';
-import atkSemantic from 'atk-semantic-ui';
+import atkFomantic from 'atk-fomantic-ui';
```
mvorisek, line 2: not sure if this rename exactly, but unminified size has changed by -50% probably by this PR:

- before https://github.com/atk4/ui/blob/d7bdc09dceba2e8575ff0ae39e9ee12ab60e9781/public/atkjs-ui.js
- after https://github.com/atk4/ui/blob/80b1e9b23322681b8203f4faf7f09d4a182f1fb3/public/atkjs-ui.js

PR #1868, opened 2022-09-27
Allow Grid::addModalAction() button to be disabled

mkrecek234:


mkrecek234:
Sample usage:
```
$grid->addModalAction(['icon'=>'flag'], ['title' => 'Move'], function ($v, $id) use ($grid) {
      ...
            });
        },[], function ($row) { return rand(0,1) == 1;});

File: phpstan.neon.dist [hunk 1]
```diff
@@ -345,6 +345,9 @@ parameters:
         -
             path: 'src/Grid.php'
             message: '~^Method Atk4\\Ui\\Grid::addSelection\(\) should return Atk4\\Ui\\Table\\Column\\Checkbox but returns Atk4\\Ui\\Table\\Column\.$~'
+        -
+            path: 'src/Grid.php'
```
mvorisek, line 349: just addd `bool` type

src/Grid.php
```diff
@@ -542,9 +542,9 @@ public function addPopup($columnName, $popup = null, $icon = 'caret square down'
      *
      * @return View
      */
-    public function addModalAction($button, $title, \Closure $callback, $args = [])
+    public function addModalAction($button, $title, \Closure $callback, $args = [], $isDisabled = false)
```
mvorisek, line 502: I belive this is not per row, not sure what usecase it should solve.

mkrecek234: Sure, the isDisabled is either boolean or closure, if you pass function($row) { return ($row->get('age') > 12); } you can dynamically set disabled per row as needed. @mvorisek Have many use cases where this is perfect - as it is just extends an already existing feature of addActionButton to addModalAction I'd appreciate if we could merge it

phpstan.neon.dist
mkrecek234: isDisabled is either closure or bool

src/Grid.php
mvorisek: this need a phpdoc `@param` above the method like the `enabled` property phpdoc

https://dev.agiletoolkit.org/demos/tutorial/actions.php?atk_layout_maestro_wizard=3 - UA supports this natively, why do you want to extent this method? Advanced options should not be configureable by a custom parameter. One of my usecases was a different icon for example, should this be addressed by another param? No.

mkrecek234: addAction button has this feature today and all I do is to allow also addModalAction to use this helpful feature. Not everyone is using UA - a lot are using addActionButton etc. to configure Crud which is helpful. I do not see why this is a problem. Or do you want to delete also the isDisabled feature from addActionButton? That would be a pity

src/Grid.php
```diff
@@ -537,8 +537,9 @@ public function addPopup($columnName, $popup = null, $icon = 'caret square down'
      *
      * @param string|array|View $button
      * @param string            $title
-     * @param \Closure          $callback function (View $page) {...
-     * @param array             $args     extra url argument for callback
+     * @param \Closure          $callback    function (View $page) {...
+     * @param array             $args        extra url argument for callback
+     * @param \Closure | bool   $isDisabled  function ($row) { ... }
```
```suggestion
     * @param \Closure|bool   $isDisabled  function ($row) { ... }
```

src/Grid.php
mvorisek: creating UA with a seed should be the desired way to solve \*any\* related special settings, I am quite against the proposed changes

mkrecek234: Sure we could re-work this to flexibly have multiple advanced parameters - I'd say that would be an idea for future improvements. Alternative could be `ActionButtons::onHook(HOOK_GET_HTML_TAGS, function ...)` similar to 
`$table->onHook(Table\Column::HOOK_GET_HTML_TAGS, function (Table $table, Model $row) { ...`where we return array of icon, disabled, title

mkrecek234: I do not see why you are against it? There are multiple ways to code, and UA is not always the best way, especially if you want to dynamically adjust Cruds' action buttons in certain views. We should not enforce one single way of coding, but keep room for alternatives, as we all do not have all use cases in mind. 
This PR is nothing else than creating consistency between addActionButton and addModalAction with existing features. I am absolutely against removing isDisabled feature unless UA becomes much more easy to code with less lines of code. And even then there might be a lot of reasons why people do not want to use UA but call addActionButton/ addModalAction directly.

mkrecek234: @ibelar Any view on this PR and our pro / con discussion here?

mvorisek: many params is always a bad design :)

if UA params have to be supported by these methods, it should be accepted as a seed

also missing a test, at least a render demo like https://ui.agiletoolkit.org/demos/tutorial/actions.php?atk_layout_maestro_wizard=3

mvorisek:
rebase in one squashed commit onto develop, narrow \Closure phpdoc types, add Behat test and then let's review this PR

src/Grid.php
```diff
@@ -504,13 +504,14 @@ public function addPopup($columnName, $popup = null, $icon = 'caret square down'
...
      * @param string                            $title
      * @param \Closure(View, string|null): void $callback
-     * @param array                             $args     extra URL argument for callback
+     * @param array                             $args       extra URL argument for callback
+ 	* @param \Closure|bool                      $isDisabled function (Model $row) { ... }
      *
      * @return View
      */
-    public function addModalAction($button, $title, \Closure $callback, $args = [])
+    public function addModalAction($button, $title, \Closure $callback, $args = [], $isDisabled = false)
```
mvorisek, line 512: bulk method was added - https://github.com/atk4/ui/pull/1920 - adjust this PR to support both methods

mvorisek:
@mkrecek234 if you are interested to pursue this PR, let me know, now, I am closing it due inactivity

also see https://github.com/atk4/ui/pull/1920#discussion_r1209057437, the "disable action" topic is much more complex and we need to support is uniformly for everything (or nothing)

for discussion, there is https://github.com/atk4/ui/issues/1966 issue

mvorisek:
I have reworked this PR to be consistent. The idea is good.

mkrecek234:
> I have reworked this PR to be consistent. The idea is good.

Thank you, @mvorisek

PR #1869, opened 2022-09-27
Move user actions tutorial into data-action demos menu

mkrecek234:
User Actions tutorial is only available from index page, but should also get a separate menu item under the Data Action menu group.


demos/index.php
```diff
@@ -26,4 +26,4 @@
 $t->addParagraph('In this version of Agile Toolkit we introduce "User Actions"!');
 
 Button::addTo($app, ['Learn about User Actions', 'class.huge basic primary fluid' => true, 'iconRight' => 'right arrow'])
-    ->link('tutorial/actions.php');
+    ->link('data-action/actions-tutorial.php');
```
mvorisek, line 29: then tutorial demo directory has one file...

mkrecek234: So what do you suggest?

mvorisek: I have no complete overview of all demos, thus I cannot say exactly. Does the tutorial even provide some extra value over the standard "actions" demo? If not much, I do not see a value in this change.

mkrecek234: Yes it does - in chat ibelar pointed me to an example which was NOT covered in the data-action demo folder. So surely the tutorial should be placed there as it gives an intro to UA. Sure we could rework the whole index page and reshuffle the demos, which is quite some work - this PR just helps to pit examples to the chapter they belong to

mvorisek: not sure if the tutorial should be the 1st menu item or if it should be named a tutorial at all

but I would not object the changes in this PR if @ibelar is happy with them

mvorisek:
Docs improvements are always welcomed, but I feel this change is quite isolated. In menu, we do not have tutorials.

PR #1872, opened 2022-10-01
Move non-View JxXxx classes into Js namespace

mvorisek:
regex for renamespaced classes:
```
Ui\\\\?(Jquery|JsChain|JsConditionalForm|JsExpression(able)?|JsFunction|JsModal|JsReload|JsToast|JsVueService)
```


mvorisek:
I am not sure about this PR completely. Let's not merge it into v4.0.0 release, maybe later.

mkrecek234:
Hi @mvorisek For clarification: you want to move those Js classes to sub-namespace as they do not return views? How about JsModal or JsToast - they do return views, whereas JQuery etc. surely not. Not sure what the separating line for you is to decide where to place Js classes. Or move all into sub-namespace?

mvorisek:
> JsModal or JsToast

really?

mkrecek234:
JsModal triggers to show a modal and JsToast triggers to show a Toast, so in my understanding this is UI-related, even if this happens on the Js layer. Question was solely, what is the distinction for you what to move to Js - if all Js classes, fine; if only non-view related then this might get confusing.

mvorisek:
changed title from `view` to `View`, `JsModal` is js utility, but does not create `View` based modal

mkrecek234:
Fine for me 👍

mkrecek234 (approved):

PR #1874, opened 2022-10-01
Unify nowdoc CS

mvorisek:


src/JsSortable.php
```diff
@@ -45,7 +45,7 @@ protected function init(): void
         if (!$this->view) {
             $this->view = $this->getOwner();
         }
-        $this->getApp()->requireJs('https://cdn.jsdelivr.net/npm/@shopify/draggable@1.0.0-beta.12/lib/draggable.bundle.js');
+        $this->getApp()->requireJs($this->getApp()->cdn['atk'] . '/external/@shopify/draggable/lib/draggable.bundle.js');
```
mvorisek, line 48: fix https://github.com/atk4/ui/pull/1873

PR #1881, opened 2022-10-02
Always observe Modal changes

### LINKED ISSUE #1880: Modal should set `observeChanges` by default
related with https://github.com/atk4/ui/pull/334

currently in most of the Modal usage `observeChanges` is enabled manually

see also related discussion: https://github.com/atk4/ui/pull/553

mvorisek:
fix #1880

discussion https://github.com/atk4/ui/pull/334#discussion_r985299979

marking as BC-break a several settings Modal methods were removed

### DIFF
--- a/src/Modal.php
+++ b/src/Modal.php
@@ -47,7 +47,11 @@ class Modal extends View
     /** @var array */
     public $args = [];
     /** @var array */
-    public $options = [];
+    public $options = [
+        // any change in modal DOM should automatically refresh cached positions
+        // allow modal window to add scrolling when content is added after modal is created
+        'observeChanges' => true,
+    ];
 
     /** @var string Currently only "json" response type is supported. */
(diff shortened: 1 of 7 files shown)


src/Modal.php
```diff
@@ -337,21 +265,9 @@ protected function renderView(): void
...
-        if (isset($this->options['modal_option'])) {
-            $this->js(true)->modal($this->options['modal_option']);
-        } else {
-            $this->js(true)->modal();
-        }
-
-        // add setting if available.
-        if (isset($this->options['setting'])) {
-            foreach ($this->options['setting'] as $key => $value) {
-                $this->js(true)->modal('setting', $key, $value);
```
mvorisek, line 350: @ibelar any idea why the settings were not passed to the Fomantic-UI Modal constructor (initial `->js(true)->modal` call) directly?

PR #1892, opened 2022-10-07
Remove obsolete notify and spinner JS code

mvorisek:
Thanks to #1891 we can monitor which JS code is used or not.

This PR aims to either cover the uncovered JS files or remove them.


mkrecek234:
JsNotify was removed and JsToast as an identical replacement was mentioned. However, JsNotify notified the user *inside the control*, so for example an input field turned green, and inside the message was shown. Very compact, and with a clear link to the control that triggered the notification. I would suggest to get it back in, it served a purpose for daily use cases in my understanding. @mvorisek

mvorisek:
@mkrecek234 we can but it must be fully tested and the bugs must be fixed, when I was removing it, I found several bugs and not use in atk4/* ecosystem, it felt like some obsolete part to me

mkrecek234:
Can you re-introduce and list the bugs to be fixed? Unless other @atk4/trusted-maintainers  also would prefer to remove it.

mvorisek:
bugs - revert this PR locally and check the demos where replaced now with Toast

what is your usercase? do you need the Toast functionality "to place notification at specific place"? do you need to call it from JS, or just from PHP?

maybe put here some example of your usage and describe what functionalitites you are missing since this change

mkrecek234:
bugs - can we please re-introduce and file an issue as usual?
use case - if you want a notification that is triggered by a control, with JsNotify the control itself will include the notification. This was standard behaviour for File Upload as far as I know. Looks better than having a JsToast somewhere, as the message belongs to the upload control someone just used.  It is being called from PHP onUpload - I use it for onDownload and onDelete.

mvorisek:
all official usage was migrated, uploader docs: https://github.com/atk4/ui/pull/1892/files#diff-2b1ca070774a027b042a96b00c877f278c3b9bd7dd38fa166bd51175dc799d14R89

the bugs are comming from fact it was a non native Fomantic-UI component, I am against reintroducing the old broken JsNotify as long as there is some alternative

the position is a good point, but for that, any existing atk4 View based item can be used

PR #1896, opened 2022-10-08
Lazy load vue in JS vueService

mvorisek:
to reduce the main JS file size


mvorisek:
this is not easily possible as async context is required in Javascript

mvorisek:
this is problematic also for flatpickr and Sortable globals loaded after main JS

we need some solution as currently flatpickr is always loaded - no problem expect the extra download needed for every page

Sortable is added to App only when user - JS is not loaded if component is loaded using AJAX

PR #1919, opened 2022-10-26
No unneeded/nested jQuery.ready() wrap

mvorisek:
only script rendered in the initial html needs to be wrapped

no nested scripts should be wrapped no AJAX atkjs response


src/View.php
```diff
@@ -1116,20 +1116,12 @@ public function getJs(bool $forceReturn = false)
...
-
         // delegate $action rendering in hosting app if exist.
         if ($this->issetApp() && $this->getApp()->hasMethod('getViewJS')) {
             return $this->getApp()->getViewJS($actions);
         }
 
-        $ready = new JsFunction($actions);
-
-        return (new Jquery($ready))->jsRender();
+        return (new JsExpression('[]()', [new JsFunction($actions)]))->jsRender();
```
mvorisek, line 1124: this wraps by an extra closure which is not needed

PR #1920, opened 2022-10-28
Add Grid bulk action methods

### LINKED ISSUE #1727: Execute php callback on Grid->addSelection() for multiple selection
First of all, I want to thank you for this great tool that is atk4/ui !
Since some time I have a question which I believe does not find answer in the documentation (but maybe I am just blinded).
Let's suppose we have two models `User` and `Role` and `User` `hasMany` `Role`.

- It is quite straightforward to use `\Atk4\Ui\Grid` and `addCondition` to display a given `Role` and add a `User`.
- Also we can almost out of the box display the `Role` list of a `User` and add one using the same method.

My question is : how to use `$grid->addSelection()` to mass-attribute a `Role` to a selected set of `User` ? (Which enable to use grid filtering and sorting)
In the [demo](https://github.com/atk4/ui/blob/develop/demos/collection/grid.php) there is a call to some javascript
```
$grid->menu->addItem('show selection')->on('click', new \Atk4\Ui\JsExpression(
    'alert("Selected: "+[])',
    [$sel->jsChecked()]
));
```
But I does not understand how to relies on a php callback instead that would receive the selected id list and perform some processing such at foreach the list and create the appropriate role lines or others.

mhuser:
fix #1727

This PR aims to add the bulk action feature to process a selection or multiple records in Grid. Currently it is easy to perform an operation:
1.  on a single record e.g. with `addModalAction()`
2. on the whole grid e.g. with `$grid->menu->addItem()`)

It is complex to work with the results of `$grid->addSelection()` outside writing verbose code with static modals and a hidden field.


demos/collection/grid.php
```diff
@@ -72,6 +74,45 @@
 // $grid->addExecutorButton($deleteExecutor, new Button(['icon' => 'times circle outline']));
 
 $sel = $grid->addSelection();
+// Executing a modal on a bulk selection
+$callback = function ($m, $ids) use ($grid) {
```
mvorisek: types should be defined, `$m` variable should be named `$modal`

demos/collection/grid.php
```diff
@@ -72,6 +74,45 @@
...
+    if (!$ids) {
+        Message::addTo($m, [
```
mvorisek: Build action button should be disabled instead when nothing is selected.

src/Grid.php
```diff
@@ -513,6 +513,37 @@ public function addModalAction($button, $title, \Closure $callback, $args = [])
...
+     * @param string|array|MenuItem $item
+     * @param \Closure              $callback function (View $page) {...
+     * @param array                 $args     extra url argument for callback
+     *
+     * @return Jquery
+     */
+    public function addModalBulkAction($item, \Closure $callback, $args = [])
+    {
+        if (!$this->menu) {
+            throw new Exception('Unable to add Modal Bulk Action without Menu');
```
mvorisek: explicit exception is not needed

demos/collection/grid.php
```diff
@@ -72,6 +74,45 @@
...
+        $msg->text->addParagraph('IDs to be deleted:');
+        foreach ($ids as $id) {
+            $msg->text->addParagraph($id);
+        }
+        $form = Form::addTo($m);
+        $form->buttonSave->set('Delete');
+        $form->buttonSave->icon = 'trash';
+        $form->onSubmit(function (Form $form) use ($grid, $ids) {
+            // iterate trough the selected IDs and delete them
+            $grid->model->atomic(function () use ($grid, $ids) {
```
mvorisek: Must be covered with Behat test.

src/Grid.php
```diff
@@ -513,6 +513,37 @@ public function addModalAction($button, $title, \Closure $callback, $args = [])
...
+        }
+
+        $owner = $this->getOwner();
+
+        if (is_string($item)) {
+            $item = ['title' => $item];
+        }
+
+        $modal = Modal::addTo($owner, [$item[0]]);
```
mvorisek: `$item[0]` cannot be correct here, at least not always set

mvorisek: no need to assing to `$owner` first, use `$this->getOwner()` directly

src/Grid.php
```diff
@@ -513,6 +513,37 @@ public function addModalAction($button, $title, \Closure $callback, $args = [])
         return $this->getActionButtons()->addModal($button, $title, $callback, $this, $args);
     }
 
+    /**
+     * Similar to addModalAction but apply to a multiple recors selection and display in menu.
+     * When menu item is clicked, modal is displayed with the $title and $callback is executed through VirtualPage.
+     *
+     * @param string|array|MenuItem $item
+     * @param \Closure              $callback function (View $page) {...
```
mvorisek: document `\Closure(XX): XX` type better

demos/collection/grid.php
mhuser: done

demos/collection/grid.php
mhuser: nice idea. have you an example of such magic?

src/Grid.php
mhuser: done

src/Grid.php
mhuser: it appear it is not needed at all

mhuser: ok, done

src/Grid.php
mhuser: done

demos/collection/grid.php
mhuser: Maybe the master checkbox code could enable/disable bulk action button by querying a dedicated class

src/Grid.php
```diff
@@ -513,6 +513,32 @@ public function addModalAction($button, $title, \Closure $callback, $args = [])
...
     }
 
+    /**
+     * Similar to addModalAction but apply to a multiple recors selection and display in menu.
+     * When menu item is clicked, modal is displayed with the $title and $callback is executed through
+     * VirtualPage.
+     *
+     * @param string|array|MenuItem $item
+     * @param \Closure(View, array|null): void $callback
+     * @param array                            $args     extra url argument for callback
```
```suggestion
     * @param string|array|MenuItem              $item
     * @param \Closure(View, list<string>): void $callback
     * @param array                              $args     extra url argument for callback
```

`$args` are probably not needed - what is the usecase?

also I do not understand why explicit `[]` is needed as `$this->menu->addItem` 2nd arg

src/Grid.php
```diff
@@ -513,6 +513,32 @@ public function addModalAction($button, $title, \Closure $callback, $args = [])
...
+     * When menu item is clicked, modal is displayed with the $title and $callback is executed through
+     * VirtualPage.
+     *
+     * @param string|array|MenuItem            $item
+     * @param \Closure(View, array|null): void $callback
+     * @param array                            $args     extra url argument for callback
+     *
+     * @return Jquery
+     */
+    public function addModalBulkAction($item, \Closure $callback, $args = [])
```
mvorisek, line 527:
```suggestion
    public function addModalBulkAction($item, \Closure $callback, $args = []): JsExpressionable
```

src/Grid.php
```diff
@@ -513,6 +513,32 @@ public function addModalAction($button, $title, \Closure $callback, $args = [])
...
+            $item = ['title' => $item];
+        }
+
+        $modal = Modal::addTo($this->getOwner());
+
+        $modal->set(function (View $t) use ($callback) {
+            $callback($t, $t->stickyGet($this->name) ? explode(',', $t->stickyGet($this->name)) : null);
+        });
+
+        return $this->menu->addItem($item)->on('click', $modal->jsShow(array_merge([$this->name => $this->selection->jsChecked()], $args)), []);
```
mvorisek: any reason why this return JsExpressionable and non-bulk method View?

src/Grid.php
```diff
@@ -513,6 +513,32 @@ public function addModalAction($button, $title, \Closure $callback, $args = [])
         return $this->getActionButtons()->addModal($button, $title, $callback, $this, $args);
     }
 
+    /**
+     * Similar to addModalAction but apply to a multiple recors selection and display in menu.
+     * When menu item is clicked, modal is displayed with the $title and $callback is executed through
+     * VirtualPage.
+     *
+     * @param string|array|MenuItem            $item
+     * @param \Closure(View, array|null): void $callback
```
mvorisek: why to pass `null` when `$t->stickyGet($this->name)` is empty?

mvorisek: fix this, add a Behat test and I will do a final review

no typecasting is ok as non-bulk action does not have it now too, will be fixed later in #1991

src/Grid.php
```diff
@@ -513,6 +513,32 @@ public function addModalAction($button, $title, \Closure $callback, $args = [])
         return $this->getActionButtons()->addModal($button, $title, $callback, $this, $args);
     }
 
+    /**
+     * Similar to addModalAction but apply to a multiple recors selection and display in menu.
```
```suggestion
     * Similar to addModalAction but apply to a multiple records selection and display in menu.
```

demos/collection/grid.php
mhuser: Wait on #1921 in the meantime, the action will return a void array, then iterating through it will do nothing.

mhuser:
Maybe there is an interest to `addBulkAction` too, that would fire without `Modal`.

demos/collection/grid.php
mhuser: Done

src/Grid.php
mhuser: `[]` removed; not sure for `$args` but `addModalAction` has this and I used it as template.

src/Grid.php
mhuser: Should return View.

mhuser: Done

src/Grid.php
mhuser: Replaced null by empty array and added behat test

mvorisek:
Quality PR, thank you!

mhuser:
> Quality PR, thank you!

Great !

PR #1921, opened 2022-10-28
Add master checkbox to Grid

### LINKED ISSUE #1877: Add checkbox to table header to select/unselect all
When row selection is enabled, there should be "master" checkbox in the table header to select/unselect all rows.

Feature request from https://github.com/atk4/ui/issues/1727#issuecomment-1264612174.

It should probably still select the rows in the current page only.

related demo: `demos/collection/grid.php`

maybe add support for `shift` key to add support for range selection

mhuser:
fix #1877

Add a master checkbox into grid header that enables to toggle selection of whole visible records.


src/Table/Column/Checkbox.php
```diff
@@ -47,7 +47,7 @@ public function getHeaderCellHtml(Field $field = null, $value = null): string
         }
         $this->table->js(true)->find('.' . $this->class)->checkbox();
 
-        return parent::getHeaderCellHtml($field);
+        return $this->getTag('head', [['div', ['class' => $this->class], [['input/', ['type' => 'checkbox']]]]], []);
```
mvorisek: class should be `'ui checkbox ' . $this->class`

mhuser: done

mhuser:
I have a first working master checkbox but the code is ugly and it does not account for the indeterminate state if child checkboxes are changed.
I found this [fomentic example](https://fomantic-ui.com/modules/checkbox.html#grouped-checkboxes)
Would it be cleaner to use a separated javascript file?

mvorisek:
needs a Behat test

master checkbox must reflect the child states - unchecked/"partially checked"/"all checked"

if Fomantic-UI/jQuery cannot handle this in a few lines, the JS should be put into `js/`, but until you have fully working solution, feel free to keep the code uncompiled

mhuser:
image image

Ok, it works as of now with the indeterminate state if the child checkboxes are not all in the same state.

I had some issues to have the javascript statements to appear at the proper position. I first wrote them in `Grid->addSelection()` and they were called kind of before the `checkbox()` creation for fomantic-ui.

Still I feel the manner I use the javascript is not optimal because it does not look like other atk4/ui sources.

tests-behat/grid.feature
```diff
@@ -36,3 +36,31 @@ Feature: Grid
...
+  Scenario: master checkbox
+    Given I am on "_unit-test/grid-rowclick.php"
+    When I click using selector "xpath(//div[@id='grid']//tr[2]//div.ui.child.checkbox)"
+    Then No toast should be displayed
+    When I click using selector "xpath(//div[@id='grid']//tr[1]//div.ui.master.checkbox)"
+    Then No toast should be displayed
+    When I click using selector "xpath(//div[@id='grid']//tr[1]//div.ui.master.checkbox)"
+    Then No toast should be displayed
+
+  Scenario: dynamic scroll
```
mvorisek: seems like a wrong merge

mhuser: Seems I have created a git mess

mhuser: Fixed

tests-behat/grid.feature
```diff
@@ -36,3 +36,12 @@ Feature: Grid
     When I click using selector "xpath(//div[@id='grid']//tr[2]//a)"
     Then No toast should be displayed
     Then PATCH MINK the url should match "~_unit-test/grid-rowclick.php#test~"
+
+  Scenario: master checkbox
+    Given I am on "_unit-test/grid-rowclick.php"
+    When I click using selector "xpath(//div[@id='grid']//tr[2]//div.ui.child.checkbox)"
+    Then No toast should be displayed
```
mhuser: I should add a toast that display the checkbox selection to test if the selection is OK

mhuser: Done.

mhuser:
I did not find any figures of the checkbox column inside the [`Column/Checkbox`](https://agile-ui.readthedocs.io/en/latest/tablecolumn.html?highlight=checkbox#checkbox) or the [`Grid`](https://agile-ui.readthedocs.io/en/latest/grid.html?highlight=checkbox#selection) documentation.

There seem to be no documentation updates to do because the core functionalities of `Grid` and `Column/Checkbox` are not affected.

mhuser:
> needs a Behat test
> 
> master checkbox must reflect the child states - unchecked/"partially checked"/"all checked"
> 
> if Fomantic-UI/jQuery cannot handle this in a few lines, the JS should be put into `js/`, but until you have fully working solution, feel free to keep the code uncompiled

@mvorisek The behat test is ready. It toasts the selection with a grid menu and iterates trough the various selection states by clicking on checkboxes.

The javascript is still ugly yet and fail behat latest chrome slow. I see I have to enhance it, but I do not know how.

tests-behat/grid.feature
```diff
@@ -36,3 +36,17 @@ Feature: Grid
...
+    Then Toast display should contain text "Selected:"
+    When I click using selector "xpath(//div[@id='grid']//tr[1]//div.ui.child.checkbox)"
+    Then I press menu button "Show Selection" using selector ".ui.menu.atk-grid-menu"
+    Then Toast display should contain text "Selected: 1"
+    When I click using selector "xpath(//div[@id='grid']//tr//div.ui.master.checkbox)"
+    Then I press menu button "Show Selection" using selector ".ui.menu.atk-grid-menu"
+    Then Toast display should contain text "Selected: 1,2,3,4,5"
+    When I click using selector "xpath(//div[@id='grid']//tr//div.ui.master.checkbox)"
+    Then I press menu button "Show Selection" using selector ".ui.menu.atk-grid-menu"
+    Then Toast display should contain text "Selected:"
```
mvorisek: fix EOL and test something that will text empty selection reliably

mhuser: OK, I now understand that "should contain text" do exactly what is say. It would be true for any string that contains the text and more characters and thus does not test for empty. I add "#" as EOL such as empty appears like "Selected: #".

mvorisek: image

by EOL I meant: each text file should have exactly one LF at the end

mhuser: Right! Done.

src/Table/Column/Checkbox.php
```diff
@@ -37,6 +37,7 @@ protected function init(): void
         if (!$this->class) {
             $this->class = 'cb_' . $this->shortName;
         }
+        $this->getApp()->requireJs($this->getApp()->cdn['atk'] . '/../js/src/helpers/grid-checkbox.helper.js');
```
mhuser: This line should be removed.
The javascript should be compiled (how?) in such a way it is included in the standard atk4 javascripts.
 Should be called after `$('#atk_layout_maestro_grid_view_table').find('.cb_table_column_checkbox').checkbox();`

image

 How to make this atk4/ui friendly ?

mhuser: Ok I should invoke the power of
```Bash
npm install
npm run dev
npm run build
```

mhuser:
A strange `Diff Compiled Files` Unit / Behat (latest, Chrome Slow) (pull_request) error is left.

mvorisek:
> A strange `Diff Compiled Files` Unit / Behat (latest, Chrome Slow) (pull_request) error is left.

make sure your all text files has linux/LF NL and no trailing whitespaces

mhuser:
> make sure your all text files has linux/LF NL and no trailing whitespaces

You were right: CRLF have appeared but only within /* */ comments, very strange

mhuser:
Which npm version is behat using?
Using 8.19.3 on Linux does not give me same results. But I do not know yet is the issue is npm or git.

mhuser:
> Which npm version is behat using? Using 8.19.3 on Linux does not give me same results. But I do not know yet is the issue is npm or git.

I have cloned on a linux network share using a windows git client 😅 I think that most CRLF-related difficulties come from there.

mhuser:
After deleting most files and git reset --hard from the Linux client, all tests now pass! 🎉

mhuser:
The issue was between chair and keyboard

mvorisek:
In Crud, when row is deleted, it is removed\* from DOM. Row removal must recompute the master checkbox.

Also we have dynamic scroll feature which can add table rows after the table is rendered.

\* in some other code the row is only hidden

mhuser:
> In Crud, when row is deleted, it is removed* from DOM. Row removal must recompute the master checkbox.
> 
> Also we have dynamic scroll feature which can add table rows after the table is rendered.
> 
> * in some other code the row is only hidden

OK, I think I have to somehow bind to the row changes in javascript, at least for the removal action.

mvorisek:
DOM node changes can be detected using `MutationObserver`, make sure you do not watch more events than needed to keep it efficient

mhuser:
I checked demos for a test case.
https://ui.agiletoolkit.org/demos/collection/grid.php
https://ui.agiletoolkit.org/demos/collection/crud.php
Unfortunately, none of them seems to have a working delete action.
`grid.php` has the delete button commented, if restored, the row does not fade away and the button also appears below the grid.
`crud.php` deletion just triggers a read-only error without hiding the row.

-> This is expected, as least crud.php, web demos are read only as they are public thanks to https://github.com/atk4/ui/blob/5.0.0/demos/init-db.php#L25

mhuser:
I spoke too soon. `crud.php` deletion is working well in master branch! It only has problem on the online demo.

-> see reply in your comment above

mvorisek:
@abbadon1334 can you please look as the Deploy? The RO mode is handled by demos itself - transaction is rolled back before the app exists, the database (sqlite file) itself should be writeable.

mvorisek:
image

the checkbox column should not have unneeded width

src/Table/Column/Checkbox.php
```diff
@@ -47,12 +47,15 @@ public function getHeaderCellHtml(Field $field = null, $value = null): string
                 ->addMoreInfo('field', $field);
         }
         $this->table->js(true)->find('.' . $this->class)->checkbox();
+        $this->table->js(true, new JsExpression('atk.gridCheckboxHelper.masterCheckbox();'));
 
-        return parent::getHeaderCellHtml($field);
+        return $this->getTag('head', [['div', ['class' => 'ui master checkbox ' . $this->class], [['input/', ['type' => 'checkbox']]]]], []);
```
mvorisek: image

the checkbox column should not have unneeded width

mhuser: This is true but is not a feature added by the master-checkbox 🤔 need to tweak width somehow

mhuser: Probably a css targeting the checkbox column

js/src/helpers/grid-checkbox.helper.js
```diff
@@ -0,0 +1,57 @@
...
+            // uncheck all children
+            onUnchecked: function () {
+                const $childCheckbox = $(this).closest('.table').find('.child.checkbox');
+                $childCheckbox.checkbox('uncheck');
+            },
+        });
+    },
+
+    childCheckbox: function () {
+        $('.table .child.checkbox').checkbox({
```
mvorisek: About 1/2 of the CI Behat jobs fail with:

```
tests-behat/grid.feature:79
```

image

I did not experienced such Behat failures on other CI jobs. We run Behat testing very frequently, so it seems related with this PR.

Google Chrome Performance tools in developer tools might help you to understand the issue.

mhuser: The IPP=1000 page load is slow, probably because it checks all checkboxes.
image

mhuser: Without the `$grid->addSelection();`, setting the IPP to 1000 is much faster.
image

mhuser: I will try to set `fireOnInit: false`. As of now, it is not possible to have checkboxes preset on init.

mhuser: Looks much better.
image
How fortunate this behat ipp test has been added!

mhuser:
I have now only LF line endings, but it seems that Unit / Behat (latest, Chrome Slow) keeps some CRLF in comments, which lead diff to fail 🤔

mvorisek:
> I have now only LF line endings, but it seems that Unit / Behat (latest, Chrome Slow) keeps some CRLF in comments, which lead diff to fail 🤔

disable auto crlf in your git config

and `js/package.json` and `js/package-lock.json` should have no changes, delete the lock file and restore the original deps, then run `npm install`

src/Table/Column/Checkbox.php
mhuser: image
Fixed

mhuser: used existing collapsing class and fitted checkbox

mhuser:
> disable auto crlf in your git config
> 
> and `js/package.json` and `js/package-lock.json` should have no changes, delete the lock file and restore the original deps, then run `npm install`

Thanks, I will try it !
Can you please retrigger the failing Behat (8.0, Chrome) test?
Seems it had some execution issue and not a PR-related error.

src/Table/Column/Checkbox.php
mhuser: Done.

js/src/helpers/grid-checkbox.helper.js
mhuser: Done.

mvorisek: > I will try to set `fireOnInit: false`. As of now, it is not possible to have checkboxes preset on init.

I guess the init invoked something in the master checkbox which recomputed something using the whole data, ie. O(n^2) complexity. Or some slowdown due FUI JS? In general, 1k rows is no heavy workload, so it should be fast even with 10k rows :)

mhuser: Yes, each checkbox controls all the other checkboxes to see if the master checkbox has to be changed, which is very inefficient. The default is unchecked for all, and cannot be changed yet so this check has no sense.

mhuser:
> DOM node changes can be detected using `MutationObserver`, make sure you do not watch more events than needed to keep it efficient

I was not able to improve my js skills enough to address this yet. Maybe we could merge this PR as is (when the quality will be satisfactory) and open a new issue for that cases where a row is removed/hidden from DOM or the grid is dynamically scrolled?

A preliminary test on my production server already granted me positive feedback from the users that appreciate the possibility to check all the displayed rows in a single click 😄  When the array has multiple pages, only the visible page is selected.

mhuser:
In combination with the modal bulk action #1920, this will enable to mass process of selected elements of a grid conveniently.
This needs to be merged first in such a way the other PR could implement Disable/Enable of the bulk action depending on the master checkbox state.

tests-behat/grid.feature
```diff
@@ -37,6 +37,20 @@ Feature: Grid
...
+    Then Toast display should contain text "Selected: #"
+    When I click using selector "//tr[1]//div.ui.child.checkbox"
+    Then I press button "Show Selection"
+    Then Toast display should contain text "Selected: 1#"
+    When I click using selector "//tr//div.ui.master.checkbox"
+    Then I press button "Show Selection"
+    Then Toast display should contain text "Selected: 1,2,3,4,5#"
+    When I click using selector "//tr//div.ui.master.checkbox"
+    Then I press button "Show Selection"
+    Then Toast display should contain text "Selected: #"
```
mvorisek, line 63: TODO https://github.com/atk4/ui/pull/1920#discussion_r1209057437

tests-behat/grid.feature
```diff
@@ -37,6 +37,20 @@ Feature: Grid
...
+
```
mvorisek, line 71: #1920 adds support for non-UA (atk4/data User Action) bulk support

before we finish this PR, another PR should be made to add bulk support for "regular UA for multiple records" - https://github.com/atk4/data/blob/4.0.0/src/Model/UserAction.php#L32

src/Table/Column/Checkbox.php
mvorisek: please submit this as separate PR

demos/_unit-test/grid-rowclick.php
```diff
@@ -17,7 +17,7 @@
 require_once __DIR__ . '/../init-app.php';
```
mvorisek, line 17: please keeps conversations in threads

> I checked demos for a test case.
> https://ui.agiletoolkit.org/demos/collection/grid.php
> https://ui.agiletoolkit.org/demos/collection/crud.php
> Unfortunately, none of them seems to have a working delete action.
> `grid.php` has the delete button commented, if restored, the row does not fade away and the button also appears below the grid.
> `crud.php` deletion just triggers a read-only error without hiding the row.

This is expected, as least `crud.php`, web demos are read only as they are public thanks to https://github.com/atk4/ui/blob/5.0.0/demos/init-db.php#L25

src/Table/Column/Checkbox.php
mvorisek: > used existing collapsing class

good finding of correct change, but I am not sure if we should do it, the issue is non-`collapsing` table itself

> and fitted checkbox

👍, merged in https://github.com/atk4/ui/pull/2154

src/Table/Column/Checkbox.php
```diff
@@ -50,13 +50,16 @@ public function getHeaderCellHtml(Field $field = null, $value = null): string
...
 
-        return parent::getHeaderCellHtml($field);
+        return $this->getTag('head', [['div', ['class' => 'ui master fitted checkbox ' . $this->class], [['input/', ['type' => 'checkbox']]]]], ['class' => ['collapsing']]);
     }
 
     #[\Override]
     public function getDataCellTemplate(Field $field = null): string
     {
-        return $this->getApp()->getTag('div', ['class' => 'ui checkbox ' . $this->class], [['input/', ['type' => 'checkbox']]]);
+        $this->table->js(true, new JsExpression('atk.gridCheckboxHelper.childCheckbox();'));
```
mvorisek, line 61: this must be moved into `getHeader...` otherwise it will be added multiple times because of #2146

mvorisek:
TODO for myself - review after #2225

tests-behat/grid.feature
mvorisek: moving into https://github.com/atk4/ui/issues/1966#issuecomment-2495666872 feature request

tests-behat/grid.feature
mvorisek: I have removed `APPLIES_TO_MULTIPLE_RECORDS` for now in favor of `APPLIES_TO_ALL_RECORDS` until implemented fully.

js/src/helpers/grid-checkbox.helper.js
mvorisek: There is still something wrong with the performance. When I set `ipp` to 500 which lists about 250 countries, checking the master checkbox tooks about 500 ms. That is way to slow than I would expect. I am checking what can be done.

mvorisek:
@mhuser, thank for this great contribution and sorry for longer review! ❤

mhuser:
@mvorisek 
You are welcome. This is great! 😃

PR #1925, opened 2022-10-29
Fix modal error handling when opening

### LINKED ISSUE #1723: Card demo is broken
steps to reproduce:

1. open `/demos/data-action/factory-view.php`
2. click
  image
3. the following is shown:
  image

mvorisek:
the error handling topic is quite complex, `closeOnLoadingError` modal data variable is newly set when the modal is to be opened for the first time from API, once the load is complete, the variable is removed.

Any inner modal must NOT close any parent modal! (checked in JS code, JS exception in console will be thrown, but the modal will never be closable by an user then)


tests-behat/modal-error.feature
```diff
@@ -0,0 +1,28 @@
...
+    Then I hide js modal
+    When I press button "Test Modal load JS error"
+    Then Modal is open with text 'Fomantic-UI "modal.onShow" setting cannot be customized outside atk'
+    Then I hide js modal
+
+  Scenario: ModalExecutor with PHP error exception is displayed
+    When I press button "Test ModalExecutor load PHP error"
+    Then Modal is open with text "Atk4\Data\Exception [code: 404]: Record with specified ID was not found"
+    Then I hide js modal
+    # TODO Behat is failing
```
mvorisek, line 25: there is some mistake, a few other Behat tests are failing too

PR #1926, opened 2022-10-30
Fix DropdownCascade /w custom row render callback

samsouthardjr:
This is a very small bug fix -- the getNewValues method should not use the hard-coded 'name' field if a renderRowFunction exists


mvorisek:
LGTM, thank you!

PR #1939, opened 2022-11-24
Fix Table\Column\Labels  for null fields

mkrecek234:
Steps to reproduce:
- Adjust demo: https://dev.agiletoolkit.org/demos/collection/tablecolumns.php
- Set "interests" field for one of the records to null
- The interests field is configured to show content in table as Table\Column\Labels which will render an error, since explode(...) does not work with null strings.


mvorisek:
Thank you!

PR #1948, opened 2022-12-12
Lint LESS files using Stylelint

mvorisek:


public/css/agileui.less
```diff
@@ -113,14 +113,14 @@
             justify-content: center;
         }
     }
-    &:not(active) {
+    &:not(.active) {
```
mvorisek, line 116: was a bug I belive

js/.stylelintrc.js
```diff
@@ -0,0 +1,27 @@
...
+        'color-function-notation': 'legacy',
+        indentation: 4,
+        linebreaks: 'unix',
+        'max-line-length': null,
+        'no-descending-specificity': null,
+        'rule-empty-line-before': null,
+        'string-quotes': 'single',
+
+        // TODO
+        'at-rule-no-unknown': null, // 1 error: Unexpected unknown at-rule "@atkFooterHeight" (a stylelint bug?)
```
mvorisek, line 18: TODO before merge

public/css/agileui.less
```diff
@@ -216,7 +211,7 @@ footer.atk-footer {
     width: 40%;
     z-index: @rightPanelIndex;
     &.atk-visible {
-        box-shadow: 0px 0 5px fade(@black, 20%);
+        box-shadow: 0 0 5px fade(@black, 20%);
```
mvorisek: this css value is not passing styleling, introduced in https://github.com/atk4/ui/commit/03f0ab2027e8bcd6c2a2b27d95c058e7bf082566

@ibelar any idea how to fix it?

ibelar: did you try like this:         
box-shadow: 0px 0px 5px fade(@black, 20%);

mvorisek: LESS can render it, but what should it do?

docs https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow

mvorisek: https://lesscss.org/less-preview/#eyJjb2RlIjoiQGJsYWNrOiAjMDAwO1xuXG5saWIge1xuICAgIGJveC1zaGFkb3c6IDAgMCA1cHggZmFkZShAYmxhY2ssIDIwJSk7XG59IiwiYWN0aXZlVmVyc2lvbiI6IjQuMS4zIn0= `fade()` seems to be a LESS function but Stylelint does not like it...

and should be safe to change to `rgba` https://lesscss.org/less-preview/#eyJjb2RlIjoiQGJsYWNrOiAjMDUwO1xuXG5saWIge1xuICAgIGJveC1zaGFkb3c6IDAgMCA1cHggcmdiYShAYmxhY2ssIDIwJSk7XG4gICAgYm94LXNoYWRvdzI6IDAgMCA1cHggZmFkZShAYmxhY2ssIDIwJSk7XG59IiwiYWN0aXZlVmVyc2lvbiI6IjQuMS4zIn0=

PR #1959, opened 2022-12-29
Add multi column decorator support in CardTable

DarkSide666:
previously value was typecasted to UI string wrongly when building the data, but the data for later formatting are expected to in native datatype (like `bool`)

TODO render country name in flag tooltip and add test asserting the format and asserting no reload

### DIFF
--- a/demos/interactive/cardtable.php
+++ b/demos/interactive/cardtable.php
@@ -12,4 +12,11 @@ require_once __DIR__ . '/../init-app.php';
 
 Header::addTo($app, ['Card displays read-only data of a single record']);
 
-CardTable::addTo($app)->setModel((new Stat($app->db))->loadAny());
+$entity = (new Stat($app->db))->loadAny();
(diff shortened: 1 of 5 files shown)


mvorisek:
@DarkSide666 please rebase this PR

demos/interactive/cardtable.php
```diff
@@ -12,4 +12,11 @@
...
 Header::addTo($app, ['Card displays read-only data of a single record']);
 
-CardTable::addTo($app)->setModel((new Stat($app->db))->loadAny());
+$entity = (new Stat($app->db))->loadAny();
+$entity->project_code .= ' <b>no reload</b>';
+
+CardTable::addTo($app)->setModel($entity);
+
+// CardTable uses internally atk4_local_object type which uses weak references,
+// force GC to test the data are kept referenced correctly
```
mvorisek, line 21: fow what purpose?

PR #1968, opened 2023-01-11
Drop AbstractView::initDefaultApp and App::init methods

mvorisek:
Creating many implicit apps can imply negative performance and/or unexpected (other than configured, like implicit UI persistence comma) behaviour, both hard to debug.

If you call `View::invokeInit()` manually (not recommended, but needed for edge cases), then you need to set the app manually before the call.

This PR also asserts app is set no later than in init, this requirement may be drop in the future, but currently it is needed to load a template which is defined in almost every View.

Also drop `ExecutorFactory::create()` method in favor of `ExecutorFactory::createExecutor()`.

### DIFF
--- a/src/Table/Column/ActionMenu.php
+++ b/src/Table/Column/ActionMenu.php
@@ -68,6 +68,7 @@ class ActionMenu extends Table\Column
             $item = Factory::factory([View::class], ['name' => false, 'ui' => 'item', 'content' => $item]);
         }
 
+        $item->setApp($this->getApp());
         $this->items[] = $item;
 
         $item->addClass('{$_' . $name . '_disabled} i_' . $name);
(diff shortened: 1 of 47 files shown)


src/Table/Column/ActionMenu.php
```diff
@@ -68,6 +68,7 @@ public function addActionMenuItem($item, $action = null, string $confirmMsg = ''
             $item = Factory::factory([View::class], ['name' => false, 'ui' => 'item', 'content' => $item]);
         }
 
+        $item->setApp($this->getApp());
```
mvorisek, line 71: in theory, app can be already set if an object with an app preset is passed

mkrecek234:
@mvorisek Using invokeInit heavily, how can I set app manually? My central app class is set using $app = new App and then $app->invokeInit() is called, as around 50 models require App Scope right away. What is the recommended solution for this BC-break?

mvorisek:
there should be no or minimal BC break if you use App in Models

PR #1973, opened 2023-01-21
Fix Tabs/Menu/Paginator wrapping when width is too small

### LINKED ISSUE #1780: Paginator & menu of tables/grids/cruds & Tabs not breaking properly on small width columns
Steps to reproduce:
- e.g., add three columns in a view
- add a crud with menu bar and paginator in the first column
- fill with sufficient data to have many pages, so paginator is wider

Error:
- menu bar does not break responsively into next line (or better: fold into burger menu)
- Paginator does not responsively break properly into two lines - exceeds width into next column

image

image

Expected result:
- Menu folds down into burger menu (or at least render in multiple lines)
- Paginator folds into menu with drop-down to select page, or at least breaks into multiple lines, does not exceed column width of outer view, even if many pages have to be displayed.

mkrecek234:
fix #1780


mvorisek:
There is an https://github.com/fomantic/Fomantic-UI/issues/2345 issue. How did you find the solution, any docs?

Also I wonder why this PR is lowering the coverage.

mkrecek234:
Please check https://fomantic-ui.com/collections/menu.html#stackable

mvorisek:
When I was developing another component I now understand what `stackable` does.

Fomantic-UI source code:
```
& when (@variationButtonStackable) {
    /* --------------
       Stackable
    --------------- */

    /* Tablet Or Below */
    @media only screen and (max-width: @largestMobileScreen) {
        .ui.stackable.buttons {
            flex-direction: column;
            width: 100%;
...
```

the title of this PR is correct (ie. for/based on mobile devices with small viewport width), but I would say we should NOT merge it, as we do not want to stack based on viewport width, but instead on owning element width

the general usecase is limitted owning element width, like several tables (with paginator/menu) next to each other

please read https://github.com/fomantic/Fomantic-UI/issues/2345 - the solution is to use `wrapping` class, but it is currently supported on FUI buttons, but not menu

if you would like to get this patched before fixed in FUI officially, you need to patch CSS to add `wrapping` class support

at least to fix #1780, `stackable` might be otherwise ok

mkrecek234:
@mvorisek From a UI perspective, menus should stack, not wrap from my understanding if you have small mobile screens vertically- you don‘t want more than one element per line in small viewport widths. On vertical oriented mobiles all menus are stacked in common web applications, otherwise it is not easy to be used. Wrapping would afaik line-break according to available size which can be on top for desktop screens. Nevertheless, for mobile it ahould stack.

@mvorisek @lubber-de What do you think?

src/Tabs.php
```diff
@@ -9,7 +9,7 @@
 class Tabs extends View
 {
     public $defaultTemplate = 'tabs.html';
-    public $ui = 'tabular menu';
+    public $ui = 'tabular menu stackable';
```
mvorisek: I support this change in Tabs.

image

there is an UX issue (the tab menu is hard it understand vs. other components/text) which should be reported, but I do not object to merge it until fixed

src/Menu.php
```diff
@@ -9,7 +9,7 @@
 
 class Menu extends View
 {
-    public $ui = 'menu';
+    public $ui = 'menu stackable';
```
mvorisek: such behaviour is not typical, here is an example of mobile GH

image

`stackable` might make UX between mobile and desktop too much different

src/Tabs.php
mkrecek234: Why not for menu? Please try a grid with larger horizontal top menu: The rows stack perfectly, but the menu does not in current code. With stackable all fine, so I would merge both changes

src/Menu.php
mkrecek234: Indeed best would be to hide remaining menu items with three dot menu, or make it horizontally scrollable as on Amazon.de. Currently it destroys the whole UI as the menu is very wide, but the rows are stacked, so stacking at least would be consistent to rest of grid

src/Tabs.php
mvorisek: see https://github.com/atk4/ui/pull/1973#discussion_r1084107040

my primary comment on this PR was in reaction on "fix 1780" description, which is definitely not fixed by this PR

stackable in Menu might be ok in some cases, but UX can be visually very different and can take much more space which might be not wanted

test `demos/collection/grid.php` demo - when viewport width is below stackable treshold, the table is already very badly reformat, so I would actually prefer to `unstackable` there

template/paginator.html
```diff
@@ -1,4 +1,4 @@
-<div class="{PaginatorType}ui pagination menu{/}">
+<div class="{PaginatorType}{/}">
```
mvorisek, line 1: this was a bug, class is added by Paginator as well

js/src/vue-components/multiline/multiline.component.js
```diff
@@ -46,10 +46,11 @@ export default {
     },
     data: function () {
         const tableDefault = {
-            basic: false,
+            compact: 'very',
+            basic: 'very',
+            unstackable: true,
```
mvorisek, line 51: when width is > stackable width (desktop), the format can be hard too read already, so this needs to be fixed properly in another PR

mvorisek:
@mkrecek234 please test this PR, my primary goal is to develop atk4/ui in a way unlimited components/nesting is possible, so the goal is not to focus on stackable (small viewports only)

"wrapping buttons" together with "menu" seems to be a good workaround for https://github.com/fomantic/Fomantic-UI/issues/2345 and I am completely ok to merge this PR even if the visual style might be not perfect/officially supported by Fomantic-UI ATM when wrapped

mvorisek:
⚠️the coverage decrease seems to be related with this PR

mkrecek234:
FUI 2.9.2 seems to be solving this both for paginator and also for menus if we add `wrapping` class. https://github.com/fomantic/Fomantic-UI/issues/2345#issuecomment-1409042396 Thank you @lubber-de

mvorisek:
Please investigate the coverage drop.

mvorisek:
I have rerun the changes in separate/fresh PR and I can confirm the coverage drop:

image

mvorisek:
@mkrecek234 it is your job to answer the CI questions in order to get this PR merged. The coverage is decreased, please explain.

Issue #1976, opened 2023-01-23
Share/deduplicate executor for multiple entities


mvorisek:
Here is an analysis of CardDeck (related with #1848 as table rows should be rendered via Views and not with cloned template) performance with many Cards:
- performance of atk4/data is top notch (entity is iterated/loaded much faster then one Card is added/rendered)
- 1 Card /wo any executor takes ~0.6 ms to add/render
- there is no single significant bottleneck, but large part of the time is spent in templating, as each Card consists of several Views, each with ~20 template tag tree children
- adding a Card executor is slow
  - `CardDeck::setMdel()` calls `Card::addClickAction`
  - creating/rendering executor trigger button is ok, it takes ~0.1 ms / single (one UA) button
  - which calls `Button::on` with atk4/data UserAction
  - an executor is created here:

![image](https://user-images.githubusercontent.com/2228672/213949761-ebedba62-38fe-4027-a454-423ee9a329d9.png)

there are two problems:

1. `ExecutorFactory::createExecutor()` creating executor itself is quite slow (~0.4 ms / single (one UA) executor), it cannot be significantly optimized, the only solution is to create an executor for a model once and share/deduplicate it for all entities (table rows, Cards, ...)

2. `ModalExecutor::executeModelAction()` executor should not be initialized/rendered until needed/triggered (#1928, ~1.0 ms / single (one UA) executor)

(3. also each executor (currently/with point 1 and 2) takes quite a lot of memory (~50 KB / single (one UA) executor) and significantly increase the total rendered html/js size (~4 KB / single (one UA) executor))

the good new is, at least in `Card::addClickAction()`, executor is created from model (not entity) atk4/data UA, ID is passed as 1st `$default` arg - so point 1 should be doable /wo heavy refactoring

given the need to share/deduplicate UA executors in atk4/ui at least, we should aim to support model (not entity) atk4/data UA and maybe even drop entity (not model) atk4/data UA completely in the future

Issue #101, opened 2017-03-31
Fix no-id view rendering


romaninsh:
It should be possible to create and render view without ID's. For instance when you add button with icon inside a Action column, it should render without ID (which otherwise would conflict with other buttons).

mvorisek:
we should check if ID was accessed or not and render it based on it

regex to find templates with IDs defined in a non standard way: `(?<!data-)id="(?!\{\$_id\}")`

mvorisek:
Should, but IDs are unique, so this is only a feature request.

PR #1979, opened 2023-01-24
No hardcoded template attributes

### LINKED ISSUE #1028: Some templates are missing common properties like id

mvorisek:
fix #1028

needed for #101

time to create/clone basic (`template/element.html`) template and to render it has been reduced by ~25% (related with #1976)

see `template/element.html` changes to understand how to migrate your templates

if you do not use custom templates, no action should be needed

in `View` class a few hacks are needed as `{$attributes}` tag cannot be put into all templates currently, I belive the hacks are ok otherwise a massive `HtmlTemplate` rewrite would be needed to support these cases

---

regex to find templates /wo `attributes` tag: `(?<!\s|\S)^((?!attributes)(\s|\S))*$(?!\s|\S)`


template/image.pug
```diff
@@ -1,2 +1,2 @@
-<img id="{$_id}" class="{_ui}ui{/} {$class} {_class}icon{/}" src="{$Content}" style="{$style}" {$attributes}>
+<img class="{_ui}ui{/} {$class} {_class}icon{/}" src="{$Content}" {$attributes}>
```
mvorisek, line 1: `icon` is probably historical typo here

PR #1980, opened 2023-01-29
Share user action executor for multiple entities

### LINKED ISSUE #1976: Share/deduplicate executor for multiple entities

mvorisek:
fix #1976

time to render 2 (add, delete) actions for 1000 entities is reduced from ~2000 ms to ~500 ms, ie performance is improved ~4x!

no BC break, everything works as before, only the performance on both server & client side is massively improved, also much less data needs to be transfered between the server and client

### DIFF
--- a/src/Behat/Context.php
+++ b/src/Behat/Context.php
@@ -62,10 +62,7 @@ class Context extends RawMinkContext implements BehatContext
         if (!str_starts_with($event->getStep()->getText(), 'Toast display should contain text ')
             && $event->getStep()->getText() !== 'No toast should be displayed'
         ) {
-            $hadToast = $this->getSession()->evaluateScript('jQuery(\'.toast-box > .ui.toast\').toast(\'close\').length > 0;');
-            if ($hadToast) {
-                usleep(600_000); // TODO
-            }
+            $this->getSession()->executeScript('jQuery(\'.toast-box > .ui.toast\').toast(\'destroy\')');
         }
     }
 
--- a/src/Behat/Context.php
+++ b/src/Behat/Context.php
@@ -62,7 +62,10 @@ class Context extends RawMinkContext implements BehatContext
         if (!str_starts_with($event->getStep()->getText(), 'Toast display should contain text ')
             && $event->getStep()->getText() !== 'No toast should be displayed'
         ) {
-            $this->getSession()->executeScript('jQuery(\'.toast-box > .ui.toast\').toast(\'close\');');
+            $hadToast = $this->getSession()->evaluateScript('jQuery(\'.toast-box > .ui.toast\').toast(\'close\').length > 0;');
+            if ($hadToast) {
+                usleep(600_000); // TODO
+            }
         }
     }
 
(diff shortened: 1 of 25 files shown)


src/UserAction/SharedExecutor.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * @return array<int, JsExpressionable>
+     */
+    public function jsExecute(array $urlArgs): array
+    {
+        // TODO executor::jsExecute() should be called only once, registered as a custom jQuery event and then
+        // call the event from JS with arguments to improve performance, ie. render (possibly large) JS only once
+        $res = $this->getExecutor()->jsExecute($urlArgs); // @phpstan-ignore-line
```
mvorisek, line 38: TODO

src/UserAction/JsCallbackExecutor.php
```diff
@@ -51,16 +52,21 @@ public function setAction(Model\UserAction $action)
...
     }
 
-    /**
-     * Execute model user action.
-     */
-    public function executeModelAction(array $args = [])
+    public function jsExecute(array $urlArgs = []): JsExpression
+    {
+        $res = parent::jsExecute();
+        $res->_chain[0][1][0]['urlOptions'] = array_merge($res->_chain[0][1][0]['urlOptions'], $urlArgs); // @phpstan-ignore-line TODO hack to accept/pass ID to parent/JsCallback::jsExecute() like JsExecutorInterface does, later, it should be passed using GET method
```
mvorisek, line 58: TODO

src/Behat/Context.php
```diff
@@ -62,7 +62,10 @@ public function closeAllToasts(BeforeStepScope $event): void
         if (!str_starts_with($event->getStep()->getText(), 'Toast display should contain text ')
             && $event->getStep()->getText() !== 'No toast should be displayed'
         ) {
-            $this->getSession()->executeScript('jQuery(\'.toast-box > .ui.toast\').toast(\'close\');');
+            $hadToast = $this->getSession()->evaluateScript('jQuery(\'.toast-box > .ui.toast\').toast(\'close\').length > 0;');
+            if ($hadToast) {
+                usleep(600_000); // TODO
```
mvorisek, line 67: TODO

PR #1989, opened 2023-01-31
Accept only string|null as template value strictly

mvorisek:
part of #1987

template cannot judge what persistence/type to use for typecasting


mkrecek234:
@mvorisek Why do we need this BC-break? It is very common to run `set` and just pass any model field to fill in template variables. The BC-break leads to an error because each and every set variable has to be first converted to string.

Use Case:
You have a HTML invoice template and want to set the integer order number.

Before:
`$template->set('invoice_no', $model->get('invoice_no'))`

Now:
`$template->set('invoice_no', (string) $model->get('invoice_no'))`

What is the benefit justifying this BC-break?

mvorisek:
Please read the PR description and let's continue the discussion in a private chat, as the real impact should be minimal with the best atk4/ui usage.

mvorisek:
image

mkrecek234:
Why is PHP-standard typecasting not allowed here where model field is an integer?
```
\Atk4\Ui\Header::addTo($app, [$model->get('stock')]); 
```
Simple code, simple PHP typecasting. Do we really want to enforce everyone to code like this?
```
\Atk4\Ui\Header::addTo($app, [(string) $model->get('stock')]); 
```

PR #1997, opened 2023-02-16
Multiple JS statements must be passed in JsBlock

mvorisek:
Previously mixed `JsExpression` and `array` types were used making correct OOP design impossible.

This was also dangerous as `array<JsExpressionable>` is encodeable to JSON with a completely different result.


src/Form/Control/Calendar.php
```diff
@@ -70,9 +71,6 @@ protected function renderView(): void
 
     public function onChange($expr, $default = []): void
     {
-        if (is_string($expr)) {
-            $expr = new JsExpression($expr);
-        }
         if (!is_array($expr)) {
```
mvorisek, line 74: TODO JsBlock should not wrap

src/CardDeck.php
```diff
@@ -218,16 +218,14 @@ protected function initActionExecutor(Model\UserAction $action): ExecutorInterfa
...
+     * @param string|JsExpressionable|Model|null $return
      */
-    protected function jsExecute($return, Model\UserAction $action)
+    protected function jsExecute($return, Model\UserAction $action): JsBlock
     {
         if (is_string($return)) {
             return $this->jsCreateNotifier($action, $return);
-        } elseif (is_array($return) || $return instanceof JsExpressionable) {
-            return $return;
+        } elseif ($return instanceof JsExpressionable) {
```
mvorisek, line 227: TODO

src/Panel/Right.php
```diff
@@ -152,10 +152,12 @@ public function addConfirmation(string $msg, string $title = 'Closing panel!', s
     /**
      * Callback to execute when panel open if dynamic content is set.
      * Differ the callback execution to the FlyoutContent.
+     *
+     * @param \Closure(object): void $fx
```
mvorisek, line 156: TODO `object` seems to pass phpstan, but recheck if better type can be used before merge

demos/_includes/DemoLookup.php
```diff
@@ -56,7 +57,7 @@ protected function initQuickNewRecord(): void
                 $chain->dropdown('set value', $row['value'])->dropdown('set text', $row['title']);
                 $ret[] = $chain;
 
-                return $ret;
+                return new JsBlock($ret);
```
mvorisek, line 60: TODO create JsBlock first and make JsBlock::addStatement() public and use it as much as possible

PR #2010, opened 2023-02-28
 Unify CardDeck toast with Crud

mkrecek234:
CardDeck's JsToast colour is white upon completion of UserAction, whereas Crud's JsToast colour is green.
This PR unifies it to green.

### DIFF
--- a/src/CardDeck.php
+++ b/src/CardDeck.php
@@ -59,7 +59,7 @@ class CardDeck extends View
     public $search = [VueComponent\ItemSearch::class];
 
     /** @var array Default notifier to perform when model action is successful * */
-    public $notifyDefault = [JsToast::class, 'settings' => ['class' => 'success', 'displayTime' => 5000]];
+    public $notifyDefault = [JsToast::class];
 
     /** Model single scope action to include in table action column. Will include all single scope actions if empty. */
     public array $singleScopeActions = [];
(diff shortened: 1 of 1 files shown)


src/CardDeck.php
```diff
@@ -59,7 +59,7 @@ class CardDeck extends View
     public $search = [VueComponent\ItemSearch::class];
 
     /** @var array Default notifier to perform when model action is successful * */
-    public $notifyDefault = [JsToast::class, 'settings' => ['displayTime' => 5000]];
+    public $notifyDefault = [JsToast::class, 'settings' => ['class' => 'success', 'displayTime' => 5000]];
```
mvorisek: This change will introduce inconsistency in the code.

`class.*success` regex shows this is not how CRUD handles it

JsToast default is set in https://github.com/atk4/ui/blob/3c592b8f10fe67c61f179c5c8723b07f8ab754b9/src/Js/JsToast.php#L37 and it defaults to `'success'`.

Please adjust the PR.

mkrecek234:
@mvorisek Please re-check.

src/CardDeck.php
mkrecek234: Please re-check PR.

mvorisek:
Thank you. The https://github.com/atk4/ui/blob/3c592b8f10fe67c61f179c5c8723b07f8ab754b9/src/Js/JsToast.php#L37 is problematic as `setDefault`/seeding is called after and JS expressions has no init in general.

PR #2013, opened 2023-03-01
Fix JsBlock refactor

mkrecek234:
Accordion validation error hook has to return JsBlock not array


mvorisek:
I missed that... :)

PR #2017, opened 2023-03-02
Fix Grid IPP selector

### LINKED ISSUE #2016: Grid IPP change is broken
Check https://dev.agiletoolkit.org/demos/collection/grid.php and try to change the Items per Page to 25.

Error:
`Atk4\Ui\Exception: Not sure what to do with argument`

mkrecek234:
fix #2016

### DIFF
--- a/src/ItemsPerPageSelector.php
+++ b/src/ItemsPerPageSelector.php
@@ -60,7 +60,7 @@ class ItemsPerPageSelector extends View
         $this->cb->set(function () use ($fx) {
             $ipp = isset($_GET['ipp']) ? (int) $_GET['ipp'] : null;
             // $this->pageLength->set(preg_replace("/\[ipp\]/", $ipp, $this->label));
-            $this->set($ipp); // @phpstan-ignore-line TODO https://github.com/atk4/ui/issues/2016
+            $this->set((string) $ipp);
             $reload = $fx($ipp);
             if ($reload) {
                 $this->getApp()->terminateJson($reload);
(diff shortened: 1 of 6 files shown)


mvorisek:
needs a Behat test and explanation why set should be done using string, also do not forget, IPP must be definitely typecasted with UI persistence :)

mkrecek234:
@mvorisek Even if you raised now the bar to enforce behat tests, I cannot provide this, as it drives efforts to fix newly introduced bugs significantly, even though the fix is simply like in this case. This bug is really annoying, as it destroys the IPP functionality simply. So I would welcome either a merge by you, or any contribution by you to make it ready-to-merge also based on your lifted requirements.

mvorisek:
Michael, adding a Behat test is not so complicated, but it is needed to prevent such bug from happening again.

Also please understand the fix is not correct as explained in my post above.

src/ItemsPerPageSelector.php
```diff
@@ -60,7 +60,7 @@ public function onPageLengthSelect(\Closure $fx): void
         $this->cb->set(function () use ($fx) {
             $ipp = isset($_GET['ipp']) ? (int) $_GET['ipp'] : null;
             // $this->pageLength->set(preg_replace("/\[ipp\]/", $ipp, $this->label));
-            $this->set($ipp);
+            $this->set((string) $ipp);
             $reload = $fx($ipp);
```
mvorisek, line 64: In `init` method a few lines above there is the same cast. The `set` is there for setting `View::$content` prop probably, the best way would be probably to remove every `set` call in order to set it for the render in `renderView`.

This and whole menu items must be typecasted using UI persistence in order for format numbers > 1k / page.

mvorisek:
Thank you @mkrecek234, no one can break it now :)

PR #2019, opened 2023-03-06
Fix JsBlock refactor

mkrecek234:
A missed refactoring for JsBlock

### DIFF
--- a/src/Table/Column/FilterPopup.php
+++ b/src/Table/Column/FilterPopup.php
@@ -9,6 +9,7 @@ use Atk4\Data\Model;
 use Atk4\Ui\Button;
 use Atk4\Ui\Form;
 use Atk4\Ui\Js\Jquery;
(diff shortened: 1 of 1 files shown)


mvorisek (changes requested):
Shouldn't this be detected by https://github.com/atk4/ui/blob/3c592b8f10fe67c61f179c5c8723b07f8ab754b9/src/View.php#L21?

Before merging this, I would like to have an explanation why this is not detected by phpstan or if we can help phpstan to detect it in general.

mvorisek:
repro: https://phpstan.org/r/7cfa9ded-de04-4bb7-828a-4be9b048baf9

phpstan issue: https://github.com/phpstan/phpstan/issues/9006

mvorisek (approved):
with fixed phpstan the original problem is correctly detected and this PR is correct, thank you

Issue #2021, opened 2023-03-10
TreeView not marking saved value 


mkrecek234:
Steps to reproduce:
- You have a form with a TreeView linked to a field, say article_group_id
- You edit an entity in the form, select a valid entry of the TreeView
- You click "Save"
- The record is saved propery, including the article_group_id
- You open/reload the edit form completely, showing the same record.

Issue:
- The TreeView does not mark the currently selected entry/article_group_id 

Expected behaviour:
- The correct selected value inside the tree is marked with a black bullet.

This worked properly prior the latest develop commits

mvorisek:
TreeView needs a Behat test to prevent breaking it in the future again.

mkrecek234:
Strange but working again without a change on Atk4 side as it seems, closing this issue

mvorisek:
Thank you for the feedback.

mkrecek234:
I identified the issue which persists. Once again, it is a result of all `id` typecasting which is being done incorrectly, as an integer is passed to the vue component as a typecasted value with thousand separator. @mvorisek 
I believe we should think about getting rid of this, as $model->idField as integers normally should never be typecasted with thousandSeparators. ThousandSeparators is a UI topic ONLY and should ideally only affect displayed values to the customers, whereas any data layer below is not affected by that. @mvorisek

mkrecek234:
Also if I do not find the time yet to provide your requested behat tests, here is the essence of the issue after further analysis:
- The input field contains the typecasted id of the selection, e.g., `id= 1.750` if you have a thousandSeparator `.`
- Even if you ensure that all `treeItems` `ids` are ALSO typecasted, the Vue TreeItemSelector is not capable if marking the selected item correctly, as it cannot treat `ids` like `1.750` as they are no longer integers, even though they match the input field.

Consequence: TreeItemSelector requires a proper integer `id`field, so current typecasting logic implemented will break this View if you have a thousandSeparator set. The only working workaround is #2022 , otherwise we have to change the logic how `id` fields are typecasted (or not) on hidden system fields in HTML.

mvorisek:
About your 2nd point - why? please point to vue code, as ID can be a string like UUID or any other string.

mvorisek:
minimal repro needed

PR #2022, opened 2023-03-16
Fix typecasting for TreeView

mkrecek234:
Solves issue #2021

### DIFF
--- a/src/Form/Control/TreeItemSelector.php
+++ b/src/Form/Control/TreeItemSelector.php
@@ -132,7 +132,7 @@ class TreeItemSelector extends Form\Control
     public function getValue()
     {
         if (is_scalar($this->entityField->get())) {
-            return $this->getApp()->uiPersistence->typecastLoadField($this->entityField->getField(), $this->entityField->get());
+            return $this->entityField->get();
         }
 
         return $this->getApp()->uiPersistence->typecastSaveField($this->entityField->getField(), $this->entityField->get());
--- a/src/Form/Control/TreeItemSelector.php
+++ b/src/Form/Control/TreeItemSelector.php
@@ -131,7 +131,7 @@ class TreeItemSelector extends Form\Control
      */
     public function getValue()
     {
-        return $this->getApp()->uiPersistence->typecastSaveField($this->entityField->getField(), $this->entityField->get());
+        return $this->getApp()->uiPersistence->typecastLoadField($this->entityField->getField(), $this->entityField->get());
     }
 
     protected function renderView(): void
--- a/src/Form/Control/TreeItemSelector.php
+++ b/src/Form/Control/TreeItemSelector.php
@@ -134,7 +134,7 @@ class TreeItemSelector extends Form\Control
         if (is_scalar($this->entityField->get())) {
             return $this->getApp()->uiPersistence->typecastLoadField($this->entityField->getField(), $this->entityField->get());
         }
-       
+
         return $this->getApp()->uiPersistence->typecastSaveField($this->entityField->getField(), $this->entityField->get());
     }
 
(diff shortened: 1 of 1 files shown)


mkrecek234:
Fix above works for single value treeItem correctly, however for multiple-value item the typecastLoadField does not support json yet. @mvorisek Can you check please? The sample with multiple-value is not working anymore: https://github.com/atk4/ui/blob/c6b9ed7364e835e4b0d835af7f763f24ae6cf63c/demos/form-control/tree-item-selector.php#L44

mvorisek (changes requested):
This cannot be right. You cannot mix load/save typecasting direction.

And as posted originally, I would like to have this tested by Behat. Not only to prevent it into the future, but as not tested, it was not fixed yet nor with #1991.

mkrecek234:
> This cannot be right. You cannot mix load/save typecasting direction.
> 
> And as posted originally, I would like to have this tested by Behat. Not only to prevent it into the future, but as not tested, it was not fixed yet nor with #1991.

Has to be typecastLoadField and not save which was wrong in the original code. However, the `json`array data type is not supported, may be you can have a look at it. For integer it is properly working with the initial fix

mvorisek:
"load" direction is almost surely wrong, as "load" direction implies a value to be loaded from an input, which is probably not the case here

mkrecek234:
Valid point - the key solution has to be that scalar values shall not be typecasted (see latest commit), as the Vue component cannot handle ids with thousandSeparator. Still typecasting for non-scalars.

Working and tested now in live setup and demos, sorry behat test has to be provided by another contributor

mvorisek:
Still the fix is wrong, the ID can be internally stored as any type - see #1991 prooving this case - ID wrapped in an object.

mkrecek234:
Would appreciate to see your improved code proposal - the current implementation in develop branch is definitely wrong, and also not working in a defined setup

mvorisek:
The main question is why is the typecast an issue. Any fix of `getValue` method will introduce inconsistency with `Form\Control\Input::getValue()` method. I am currently out of the time to look into this as the proper fix incl. test might take MD easily.

mkrecek234:
> The main question is why is the typecast an issue. Any fix of `getValue` method will introduce inconsistency with `Form\Control\Input::getValue()` method. I am currently out of the time to look into this as the proper fix incl. test might take MD easily.

The Vue TreeItemSelector does not support an Input field value which has thousandSeparator, so ID = 1.750 will not work, even though Atk4 saves and loads it properly. The input field thus has to have a non-typecasted ID inside, otherwise TreeViewItemSelector will not mark the correct item as active.

mkrecek234:
The main reason for the above issue is simply that `treeItems` `id` values are not typecasted, whereas the currently selected  value(s) are typecasted. The Vue component simply then cannot match the correct items.
So either we typecast the `treeItems` node `id` values or we do not typecast the `input` current value. In the light of the discussion, as those values are all "hidden" system value, typecasting of integers does not add any value here in my opinion.

mvorisek:
Please understand the solution you provided cannot work with #1991, ie. the fix is still not correct, there is no reason to handle scalars differently as long as there are not handled differently on reverse typecasting - they are not and never will be.

The correct way to fix this problem is to:
a) provide minimalistic test cases - basic, and if not needed, for edge cases too
b) verify the visible data/integers/datatimes (objects) are typecasted correctly
c) verify all cases/save etc. are fixed properly
d) verify the TreeView coverage has increased (open CodeCov and check uncovered lines, regular scenarios must be covered)

mvorisek (changes requested):
Please understand the solution you provided cannot work with #1991, ie. the fix is still not correct, there is no reason to handle scalars differently as long as there are not handled differently on reverse typecasting - they are not and never will be.

The correct way to fix this problem is to:
a) provide minimalistic test cases - basic, and if not needed, for edge cases too
b) verify the visible data/integers/datatimes (objects) are typecasted correctly
c) verify all cases/save etc. are fixed properly
d) verify the TreeView coverage has increased (open CodeCov and check uncovered lines, regular scenarios must be covered)

mvorisek:
Let's start with a failing test.

PR #2023, opened 2023-03-20
Add Model support for TreeItemSelector control source data

mkrecek234:
The current implementation of TreeItemSelector requires a static array of nodes to display a tree. However, TreeItemSelector should also be usable with tree nodes in an `Atk4/Model`.

The included implementation offers an easy way to use any table which has tree items and references to a parent tree item id to build a dynamic tree.

It includes also an extended demo to show how it is being used. 
Given the usage of model data, this will also be a prerequisite to understand the typecasting issue stated in #2022


src/Form/Control/TreeItemSelector.php
```diff
@@ -112,6 +120,46 @@ public function setTreeItems(array $treeItems)
         return $this;
     }
 
+    /**
+     * Recursive function to return sub-node for a given parent.
+     *
+     * @param $model : Model
```
mvorisek: This is invalid phpdoc, was it created using some IDE?

If yes, fix your IDE asap, otherwise it may be merged by a mistake. Also make sure no phpdoc is added if not needed.

src/Form/Control/TreeItemSelector.php
```diff
@@ -112,6 +120,46 @@ public function setTreeItems(array $treeItems)
...
+     * @param $nodes: array
+     * @param $parentId : int|string|null
+     *
+     * @return array|null
+     */
+    protected function addNodes(Model $model, array $nodes, $parentId = null)
```
mvorisek: ID can be non-scalar in theory.

src/Form/Control/TreeItemSelector.php
```diff
@@ -112,6 +120,46 @@ public function setTreeItems(array $treeItems)
...
+        }
+
+        return null;
+    }
+
+    public function setModel(Model $model): void
+    {
+        parent::setModel($model);
+
+        $nodes_array = (clone $this->model)->export();
```
mvorisek: Why clone? Why to materialize/export here?

src/Form/Control/TreeItemSelector.php
```diff
@@ -112,6 +120,46 @@ public function setTreeItems(array $treeItems)
...
+        $this->treeItems = $this->addNodes($model, $nodes_array);
+        unset($nodes_array);
```
mvorisek: unset is completely useless

src/Form/Control/TreeItemSelector.php
```diff
@@ -112,6 +120,46 @@ public function setTreeItems(array $treeItems)
...
+                $newNode = [];
+                $newNode['name'] = $node[$model->titleField];
+                $newNode['id'] = $node[$model->idField];
+                $newNode['parent_id'] = $node[$this->parentIdField];
+                $newNode['nodes'] = $this->addNodes($model, $nodes, $newNode['id']);
+                $result[] = $newNode;
+            }
+        }
+
+        if (count($result) > 0) {
```
mvorisek: Why this if?

src/Form/Control/TreeItemSelector.php
mkrecek234: Fixed

src/Form/Control/TreeItemSelector.php
mkrecek234: Materialize is faster, as it does not involved a new DB query for every node. So up to you what is better - fix to use model only below.

src/Form/Control/TreeItemSelector.php
mkrecek234: Fixed

src/Form/Control/TreeItemSelector.php
```diff
@@ -112,6 +120,36 @@ public function setTreeItems(array $treeItems)
...
+
+        $nodeModel = (clone $model)->addCondition($this->parentIdField, $parentId);
+
+        foreach ($nodeModel as $node) {
+            if ($node->get($this->parentIdField) === $parentId) {
+                $newNode = [];
+                $newNode['name'] = $node->getTitle();
+                $newNode['id'] = $node->getId();
+                $newNode['parent_id'] = $node->get($this->parentIdField);
+                $newNode['nodes'] = $this->addNodes(clone $model, $node->getId());
```
mvorisek: why to clone here?

src/Form/Control/TreeItemSelector.php
```diff
@@ -112,6 +120,36 @@ public function setTreeItems(array $treeItems)
...
+     * @param mixed $parentId
+     */
+    protected function addNodes(Model $model, $parentId = null): array
+    {
+        $result = [];
+
+        $nodeModel = (clone $model)->addCondition($this->parentIdField, $parentId);
+
+        foreach ($nodeModel as $node) {
+            if ($node->get($this->parentIdField) === $parentId) {
```
mvorisek, line 119: why this cond?

src/Form/Control/TreeItemSelector.php
```diff
@@ -39,6 +40,13 @@ class TreeItemSelector extends Form\Control
      */
     public $loaderCssName = 'atk-tree-loader';
 
+    /**
+     * The field name which includes the parent node's id.
```
mvorisek: One line desc should be put on the same line as phpdoc.

demos/form-control/tree-item-selector.php
```diff
@@ -50,6 +50,12 @@
 $control = $form->addControl('tree1', [Form\Control\TreeItemSelector::class, 'treeItems' => $items, 'allowMultiple' => false, 'caption' => 'Single selection:']);
 $control->set(502);
 
+$model = new File($app->db);
+$control = $form->addControl('tree2', [Form\Control\TreeItemSelector::class, 'allowMultiple' => false, 'caption' => 'Selection from tree in Atk4/Model:',
+    'parentIdField' => $model->fieldName()->parent_folder_id]);
```
mvorisek: no extra NL, I am supprised this CS passed the CI

src/Form/Control/TreeItemSelector.php
mkrecek234: Because we have to set a new parentId condition on an unconditioned model, otherwise will have contradictory conditions

src/Form/Control/TreeItemSelector.php
mkrecek234: Because we have to set a new parentId condition on an unconditioned model, otherwise will have contradictory conditions

mkrecek234: Correct myself - this clone is indeed not needed, as we clone earlier into `nodeModel` already. Correct in latest commit

mkrecek234:
Can you check this PR please - I'd like to get the model support for TreeItemSelector pushed into PR. Desperately needed. Thank you

mvorisek (changes requested):
There is something horibly wrong, with this PR applied, the `demos/form-control/tree-item-selector.php` demo is awfully slow.

Before this PR: <200 ms
After this PR: ~2000 ms

UPDATE:
and nothing is displayed at all:
image

src/Form/Control/TreeItemSelector.php
```diff
@@ -102,6 +106,36 @@ public function onItem(\Closure $fx): void
...
+    /**
+     * @param mixed $parentId
+     */
+    protected function addNodes(Model $model, $parentId = null): array
+    {
+        $result = [];
+
+        $nodeModel = (clone $model)->addCondition($this->parentIdField, $parentId);
+
+        foreach ($nodeModel as $node) {
```
mvorisek, line 118: This implies a DB query (Model::getIterator() which implies SELECT). Recursively.

Single query recursive data fetch can be hard to implement and will require RECURSIVE CTE.

A good enough compromise might be to fetch +1 tree level collectively at once instead of 1 leaf at each time.

mkrecek234: @mvorisek Recursive CTE is an overkill for such a simple task. Or shall I import the whole tree definition table into an array, and do the recursive operation inside the array? That would speed it up as well, and we talk about reasonably sized trees typically (otherwise we would need another treeview component which dynamically loads sub-trees upon click, there are only few Vue components doing this beautifully). Materializing the whole table into an array, I did before to speed up the script, but you did not like it.

mvorisek: It it obvious both extremes - current dummy impl. and whole table materialization is not a solution. The later case can produce OOM error easily, imagine the perf. on table with millions of records. It would be good to impl. recursive iterator in atk4/data Model and use it here.

mkrecek234: I think your worries are not realistic, as models for this TreeItemSelector are never tables of million entries. And if they are TreeItem is the wrong control as written before: you would need a Tree Control which dynamically loads sub-trees on clicks. As said, Recursive CTE is an overkill for this simple control, unless you see yourself in a position to implement this soon. Even with recursive CTE, the control would not load properly in time with your million-entries example. Otherwise I suggest we go for the faster export variant. Working well in real life with a few hundreds entries.

mvorisek: Michael, let's focus on the solution and please reread my posts above. I did not written CTE is needed, I have written the current demo is awfully slow, which is a fact and I hope we both understand the PR cannot be merged as is.

I can also tell that table with millions of records is a realistic usecase, not for display, but imagine any tree features per some item like product. No product will have a lot of features, but the whole table can be huge. And as long as you cannot resolve the "minimal spanning tree" in one query, you cannot fetch the data using a single query.

PR #2025, opened 2023-03-29
Feature/card improvements

DarkSide666:
Multiple improvements for Card and CardDeck - see description of commits.

Also adds additional demo how we can create CardDeck with custom Cards.
image

### DIFF
--- a/src/CardDeck.php
+++ b/src/CardDeck.php
@@ -22,7 +22,7 @@ class CardDeck extends View
 
     public $defaultTemplate = 'card-deck.html';
 
-    /** @var class-string<View> Card type inside this deck. */
+    /** @var class-string<View>|array Card type inside this deck. */
     public $card = Card::class;
 
     /** @var bool Whether card should use table display or not. */
@@ -152,7 +152,7 @@ class CardDeck extends View
         if ($count) {
             foreach ($this->model as $m) {
                 /** @var Card */
-                $c = $this->cardHolder->add(Factory::factory([$this->card], ['useLabel' => $this->useLabel, 'useTable' => $this->useTable]));
+                $c = $this->cardHolder->add(Factory::factory((array) $this->card, ['useLabel' => $this->useLabel, 'useTable' => $this->useTable]));
                 $c->setModel($m, $fields);
                 if ($extra) {
                     $c->addExtraFields($m, $extra, $this->extraGlue);
--- a/src/CardDeck.php
+++ b/src/CardDeck.php
@@ -85,10 +85,10 @@ class CardDeck extends View
     ];
 
     /** @var array A collection of menu button added in Menu. */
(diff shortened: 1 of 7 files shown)


src/CardDeck.php
```diff
@@ -147,7 +152,7 @@ public function setModel(Model $model, array $fields = null, array $extra = null
         if ($count) {
             foreach ($this->model as $m) {
                 /** @var Card */
-                $c = $this->cardHolder->add(Factory::factory([$this->card], ['useLabel' => $this->useLabel, 'useTable' => $this->useTable]));
+                $c = $this->cardHolder->add(Factory::factory((array) $this->card, ['useLabel' => $this->useLabel, 'useTable' => $this->useTable]));
```
mvorisek, line 155: never use `(array)` cast, change the def. value to be an array seed instead

src/HtmlTemplate.php
```diff
@@ -157,7 +157,7 @@ protected function _setOrAppend($tag, string $value = null, bool $encodeHtml = t
         // in this case we don't throw exception if tags don't exist
         if (is_array($tag) && $value === null) {
             foreach ($tag as $k => $v) {
-                $this->_setOrAppend($k, $v, $encodeHtml, $append, false);
+                $this->_setOrAppend($k, (string) $v, $encodeHtml, $append, false);
```
mvorisek, line 160: never cast to a string, string type is required

DarkSide666: But if you use `->set($model)`, then it will try to set model `id` which is integer into template too.

src/CardDeck.php
DarkSide666: Ok, will work on that a bit more probably tomorrow. Looks like there are more seed related things to fix in this (and Card) classes for consistency with other views and also to ease extending.

src/HtmlTemplate.php
mvorisek: see l153/typecasting above, this is already fully handled when set thru model, we cannot do it with array, as php type itself is not enought to determine DBAL/typecasting type, since #1987 we format `int` using UI persistence config, so `(string)` will not produce desired output

src/CardSection.php
```diff
@@ -64,15 +64,15 @@ public function addFields(Model $model, array $fields, bool $useLabel = false, b
     /**
      * Add fields label and value to section.
      */
-    private function addSectionFields(Model $model, array $fields, bool $useLabel = false): void
+    protected function addSectionFields(Model $model, array $fields, bool $useLabel = false): void
```
mvorisek, line 67: Private methods are often added as the API might not be well defined. If private, we do not have to care, but if protected, we need to support the API in a long term. Please confirm you understand this and also there are no simillar private methods in other components.

PR #2027, opened 2023-03-30
Fix UploadImage::clearThumbnail default image reset

DarkSide666:
`defaultSrc` property should also be used in `clearThumbnail` method if no thumbnail is passed as method argument.

### DIFF
--- a/demos/form-control/upload.php
+++ b/demos/form-control/upload.php
@@ -21,7 +21,7 @@ $control = $form->addControl('file', [Form\Control\Upload::class, ['accept' => [
 // $control->set('a_generated_token');
 
 $img->onDelete(function (string $fileId) use ($img) {
-    $img->clearThumbnail('./images/default.png');
+    $img->clearThumbnail();
 
     return new JsToast([
         'title' => 'Delete successfully',
--- a/src/Form/Control/UploadImage.php
+++ b/src/Form/Control/UploadImage.php
@@ -17,7 +17,7 @@ class UploadImage extends Upload
      *
      * @var string
      */
-    public $thumnailRegion = 'AfterAfterInput';
+    public $thumbnailRegion = 'AfterAfterInput';
 
     /** @var string|null The default thumbnail source. */
     public $defaultSrc;
@@ -30,16 +30,23 @@ class UploadImage extends Upload
             $this->accept = ['.jpg', '.jpeg', '.png'];
         }
 
-        if (!$this->thumbnail) {
-            $this->thumbnail = (new View(['element' => 'img', 'class' => ['right', 'floated', 'image'], 'ui' => true]))
-                ->setAttr(['width' => 36, 'height' => 36]);
+        $this->add($this->getThumbnail(), $this->thumnailRegion);
+    }
+
+    public function getThumbnail(): View
+    {
+        if ($this->thumbnail) {
+            return $this->thumbnail;
         }
+
+        $this->thumbnail = (new View(['element' => 'img', 'class' => ['right', 'floated', 'image'], 'ui' => true]))
+            ->setAttr(['width' => 36, 'height' => 36]);
 
         if ($this->defaultSrc) {
             $this->thumbnail->setAttr(['src' => $this->defaultSrc]);
         }
 
-        $this->add($this->thumbnail, $this->thumnailRegion);
+        return $this->thumbnail;
     }
 
     /**
--- a/src/Form/Control/UploadImage.php
+++ b/src/Form/Control/UploadImage.php
@@ -30,7 +30,7 @@ class UploadImage extends Upload
             $this->accept = ['.jpg', '.jpeg', '.png'];
         }
 
-        $this->add($this->getThumbnail(), $this->thumnailRegion);
+        $this->add($this->getThumbnail(), $this->thumbnailRegion);
     }
 
     public function getThumbnail(): View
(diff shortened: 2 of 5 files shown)


mvorisek:
Am I right the `defaultSrc` is/was working, but not when/with `clearThumbnail` method only?

Also please name the PRs better, but still <70 chars, to be self explainatory in the changelog. No RTM label needed, if PR is not RTM, mark is as draft.

src/Form/Control/UploadImage.php
```diff
@@ -58,11 +56,11 @@ public function setThumbnailSrc($src): void
     /**
      * Clear the thumbnail src.
      * You can also supply a default thumbnail src.
-     *
-     * @param string $defaultThumbnail
      */
-    public function clearThumbnail($defaultThumbnail = null): void
+    public function clearThumbnail(string $defaultThumbnail = null): void
     {
+        $defaultThumbnail ??= $this->defaultSrc;
```
mvorisek: I would prefer here `if ($defaultThumbnail === null) {}` to stress this codepath is used if the arg is empty, not only more readable, but coverage can be also measured better.

demos/form-control/upload.php
```diff
@@ -21,7 +21,7 @@
 // $control->set('a_generated_token');
 
 $img->onDelete(function (string $fileId) use ($img) {
-    $img->clearThumbnail('./images/default.png');
+    $img->clearThumbnail();
```
mvorisek, line 24: PR LGTM, but see coverage drop, please add Behat test

DarkSide666: -0.02% drop is not that much :)
I couldn't find a way how to identify clicking on image button to clear upload field contents.

mvorisek: You should be able to use some existing Behat upload test and click on the clear button using xpath selector.

DarkSide666: OK tests added and now it's `+0.29%`. Can it be merged?

src/Form/Control/UploadImage.php
```diff
@@ -58,11 +63,13 @@ public function setThumbnailSrc($src): void
     /**
      * Clear the thumbnail src.
      * You can also supply a default thumbnail src.
-     *
-     * @param string $defaultThumbnail
      */
-    public function clearThumbnail($defaultThumbnail = null): void
+    public function clearThumbnail(string $defaultThumbnail = null): void
```
mvorisek: the `$defaultThumbnail` param should be removed in favor ot the `$this->defaultSrc` only, otherwise user should use `setThumbnailSrc` method

src/Form/Control/UploadImage.php
```diff
@@ -30,24 +30,29 @@ protected function init(): void
...
 
-        if (!$this->thumbnail) {
-            $this->thumbnail = (new View(['element' => 'img', 'class' => ['right', 'floated', 'image'], 'ui' => true]))
-                ->setAttr(['width' => 36, 'height' => 36]);
+        $this->add($this->getThumbnail(), $this->thumbnailRegion);
+    }
+
+    public function getThumbnail(): View
+    {
+        if ($this->thumbnail) {
```
mvorisek: put initialization in `if ($this->thumbnail === null) {` to return only via one code/coverage flow path at the end

DarkSide666:
Comments addressed.

CodeCov has slight delay, but now it shows correct `+0.30%`

tests-behat/upload.feature
```diff
@@ -7,3 +7,13 @@ Feature: Upload
 
     When I select file input "file" with "Žlutý kůň" as "$kůň"
     Then Toast display should contain text "(name: $kůň, md5: b047fb155be776f5bbae061c7b08cdf0)"
+
+    When I click using selector "#atk_layout_maestro_form_form_layout_file_button"
```
mvorisek: Two last things:
- `Context::elementAttributeShouldContainText` method seems to be introduced in quite illogical place in the file
- the ID selectors in Behat tests can change (for ex. if placed inside another layout, generator is changed, ..), please see other tests and use selectors based on `input name=file/img` element + inner button/img tag

all other changes are perfect and the coverage increase is perfect as well

DarkSide666: Really last ones?? :D

* Moved method to misc, not sure if that's better place to be
* Changed selectors, waiting on tests

mvorisek: :D

the selectors still have the problem, the ID can be anything, base it on `<input name="img`, you can use xpath like `div[./input[@name='']]/...` to select owning div wit given `input` tag (see other Behat tests for examples)

DarkSide666: OMG I finally did that with selectors like `xpath(//div.action[.//input[@name='file']]//*[contains(concat(' ', normalize-space(@class), ' '), ' button ')][1])` :)

src/Form/Control/UploadImage.php
```diff
@@ -30,24 +30,27 @@ protected function init(): void
...
+    public function getThumbnail(): View
+    {
+        if ($this->thumbnail === null) {
             $this->thumbnail = (new View(['element' => 'img', 'class' => ['right', 'floated', 'image'], 'ui' => true]))
                 ->setAttr(['width' => 36, 'height' => 36]);
-        }
 
-        if ($this->defaultSrc) {
-            $this->thumbnail->setAttr(['src' => $this->defaultSrc]);
+            if ($this->defaultSrc) {
```
mvorisek:
```suggestion
            if ($this->defaultSrc !== null) {
```

PR #2028, opened 2023-04-04
Add App::terminate() stream support

DarkSide666:


tests/AppTest.php
```diff
@@ -68,4 +70,22 @@ public function testUnexpectedOutputLateError(): void
             ob_end_clean();
         }
     }
+
+    /* throws headers already sent exception so not sure how to write this test for stream output
+    public function testStreamResponse(): void
```
mvorisek: this test should test sending large (10GB+) data with known pattern to test the response is not materialized somewhere

src/App.php
```diff
@@ -1128,7 +1132,11 @@ protected function outputResponse(string $data): void
             return;
         }
 
-        $this->response->getBody()->write($data);
+        $stream = $this->response->getBody();
+        if ($data !== '' && $stream->isWritable()) {
```
mvorisek: the 2nd cond should not be needed as can silently loose data if `$data !== ''`

DarkSide666: I think it is needed because otherwise it can hit this exception https://github.com/Nyholm/psr7/blob/master/src/Stream.php#L244 in case you use Stream, but also provide output as string in some way.

tests/AppTest.php
DarkSide666: To be honest I have no idea how to write that. Will have to ask ChatGPT :)

src/App.php
mvorisek: yes - but you cannot ignore data - AFAIS the exception was not present and cannot pop out in our use, if yes, like with stream and late error, you need to add a test for it and handle it correctly

DarkSide666: ok will try

src/App.php
```diff
@@ -394,7 +395,7 @@ public function setResponseHeader(string $name, string $value): self
      * directly, instead call it form Callback, JsCallback or similar
      * other classes.
      *
-     * @param string|array $output Array type is supported only for JSON response
+     * @param string|array|StreamInterface $output Array type is supported only for JSON response
```
mvorisek, line 398:
```suggestion
     * @param string|StreamInterface|array $output Array type is supported only for JSON response
```

demos/_unit-test/stream.php
```diff
@@ -10,15 +10,23 @@
...
+$size_mb = (int) ($_GET['size_mb'] ?? 64);
+
 $button = Button::addTo($app, ['Download', 'class.atk-test' => true]);
-$button->on('click', function () use ($app) {
+$button->on('click', function () use ($app, $size_mb) {
     // Generate big data and write it to a temporary file
     $pattern = str_repeat('0123456789ABCDEF', 65536); // 1Mb
-    $chunks = 64; // 1 chunk = 1Mb, 1024 chunks = 1Gb etc
+    $total_size = strlen($pattern) * $size_mb;
     $tempFile = tempnam(sys_get_temp_dir(), 'test.txt');
```
mvorisek: I would like to see here a dynamic stream emitter via callback /wo intermediate temporary file.

On the DemosTest side, there should be a reader which will read the stream in parts, compare it, and discard immediatelly.

DarkSide666: Ok then I guess this is up to you. I've no idea how to do that and have to move on :)

mvorisek: implemented

tests/AppTest.php
mvorisek: implemented - DemosTest tested with 60GB, DemosHttpTest seems to buffer the output on disk, not exactly sure why, probably due some php cli server/guzzle buffering - I will look shortly tomorrow, otherwise the PR is complete

mvorisek: it seems PHP cannot flush data faster than about 150 MB/s, I have opened a php-src issue for it

mvorisek: implemented and tested with 6GB

the main bottleneck seems to be in Guzzle HTTP client - https://github.com/guzzle/guzzle/issues/3115

until improved, let's test the streaming support with 50 MB only

PR #2031, opened 2023-04-27
Add cross-request RW demos Behat testing

mvorisek:
Currently all changes are tested to be made and returned within the same DB transaction which is great to enforce the atk4/ui to made as less server requests possible and always work with fresh data for each Behat scenario, but this imply we cannot test advanced features - like CardDeck or more than one change per test scenario.

### DIFF
--- a/demos/init-db.php
+++ b/demos/init-db.php
@@ -26,6 +26,16 @@ try {
 
 trait ModelPreventModificationTrait
 {
+    protected function allowDbModifications(): bool
+    {
+        static $rw = null;
+        if ($rw === null) {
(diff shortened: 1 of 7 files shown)


demos/init-db.php
```diff
@@ -26,6 +26,16 @@
 
 trait ModelPreventModificationTrait
 {
+    protected function allowDbModifications(): bool
+    {
+        static $rw = null;
+        if ($rw === null) {
+            $rw = file_exists(__DIR__ . '/db-behat-rw.txt');
```
mvorisek, line 33: Pro tip: to test cross-request RW demos localy, create empty file manually

Issue #2034, opened 2023-05-01
`JsSearch::useAjax` is not honored


mvorisek:
In `/demos/collection/grid.php?no-ajax=1` demos I would expect no AJAX search, currently the table is updated when typing search input.

mvorisek:
https://github.com/atk4/ui/blob/5b9c371dc1/js/src/plugins/js-search.plugin.js#L191-L196 is not covered as the coverage is lost as the browser reloads the page

UPDATE: fixed in #2037

PR #2037, opened 2023-05-06
Fix Behat coverage collection when page navigates

mvorisek:
related with #2034


js/src/vue-components/inline-edit.component.js
```diff
@@ -87,7 +87,7 @@ export default {
                 method: 'POST',
                 onComplete: function (r, e) {
                     if (r.hasValidationError) {
-                        that.clearError();
+                        that.hasError = true;
```
mvorisek, line 90: bug from previous PR: https://github.com/atk4/ui/pull/2036

PR #2043, opened 2023-05-14
Prevent setting text as HTML in `Text`

### LINKED ISSUE #2040: HTML text can be passed easily to `Text`
Example:

```
$t = Text::addTo($parentView)->set('Name <b>popup');
var_dump($t->render());
```

image

`Text` class should be probably removed completely in favor of `View` only.

mvorisek:
fix #2040

### DIFF
--- a/src/Text.php
+++ b/src/Text.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Atk4\Ui;
 
+use Atk4\Ui\HtmlTemplate\Value as HtmlValue;
+
 /**
  * Simple text block view.
  */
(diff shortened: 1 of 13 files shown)


src/Text.php
```diff
@@ -34,7 +56,9 @@ public function getHtml(): string
      */
     public function addParagraph($text)
     {
-        $this->content .= $this->getApp()->getTag('p', [], $text);
+        $this->content[] = (new HtmlValue())->dangerouslySetHtml('<p>');
+        $this->content[] = (new HtmlValue())->set($text);
+        $this->content[] = (new HtmlValue())->dangerouslySetHtml('</p>');
```
DarkSide666: here you could simply use `$this->dangerouslyAddHtml('</p>');`, but that's additional function call, so probably better not.

DarkSide666 (approved):
LGTM

PR #2049, opened 2023-05-21
Improve form JS init selector

mvorisek:
related with https://github.com/atk4/ui/pull/1275 and https://github.com/fomantic/Fomantic-UI/issues/2792

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -167,11 +167,14 @@ class Form extends View
                 ->addMoreInfo('layout', $this->layout);
         }
 
+        // allow to submit by pressing an enter key when child control is focused
+        $jsSubmit = $this->js(false, null, $this->formElement)->form('submit');
+        $this->on('submit', new JsExpression('if (event.target === this) { []; }', [$jsSubmit]));
+
         // Add save button in layout
         if ($this->buttonSave) {
             $this->buttonSave = $this->layout->addButton($this->buttonSave);
             $this->buttonSave->setAttr('tabindex', 0);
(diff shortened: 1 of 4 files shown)


src/Form.php
```diff
@@ -167,14 +168,11 @@ protected function initLayout(): void
...
-        // allow to submit by pressing an enter key when child control is focused
-        $jsSubmit = $this->js(false, null, $this->formElement)->form('submit');
-        $this->on('submit', new JsExpression('if (event.target === this) { []; }', [$jsSubmit]));
-
-        // Add save button in layout
+        // add save button in layout
         if ($this->buttonSave) {
             $this->buttonSave = $this->layout->addButton($this->buttonSave);
             $this->buttonSave->setAttr('tabindex', 0);
+            $jsSubmit = $this->js()->form('submit');
```
mvorisek, line 160: Invoke `div.ui.form` `submit`, thanks to l138, the submit will be sent to the native form tag.

src/Form.php
```diff
@@ -135,9 +135,8 @@ protected function init(): void
         parent::init();
 
         $this->formElement = View::addTo($this, ['element' => 'form', 'shortName' => 'form'], ['FormElementOnly']);
+        $this->on('submit', new JsExpression('if (event.target === this) { []; }', [$this->formElement->js()->trigger('submit')]));
```
mvorisek, line 138: This is the main point of this PR, dispatch native event without instantiating Fomantic-UI form behaviour for the native form tag.

PR #2055, opened 2023-05-26
Fix DockerFile read-only SQLite error in demos

abbadon1334:
I have been running the command directly on the deploy server. However, I believe it is correct to merge this PR and test the usual deployment flow via webhook.

As pointed out by @mvorisek, the root cause of the issue lies in incorrect permissions. To resolve this, it is essential to set the correct owner for the folder where the SQLite database is located. I hope this adjustment will effectively address the problem.


mvorisek:
@abbadon1334 did you test is locally?

abbadon1334:
i tested directly on deploy server and it works, but to be sure i think the best is to trigger the webhook

mvorisek:
The read-only error is gone. Thank you!

Please fix also cert renew of https://dev.atk4.org/ to be done via cron/automatically.

PR #2073, opened 2023-06-17
Remove View::content property and View::set() method

mvorisek:
continue #2075 and #2239

`View::set` method was too abstract and clean override using LSP was not possible.

If `View::$content` property is needed (and optionally set the data using constructor and/or `View::set` method), migrate to `ViewWithContent` class which behaves like the original `View`.

There were two major usecases for `View::$content` property / `View::set()` method:
- set text data for `Content` tag
- set value for form control

### DIFF
--- a/demos/_includes/Demo.php
+++ b/demos/_includes/Demo.php
@@ -9,6 +9,7 @@ use Atk4\Ui\Exception;
 use Atk4\Ui\Js\JsExpression;
 use Atk4\Ui\Js\JsFunction;
 use Atk4\Ui\View;
+use Atk4\Ui\ViewWithContent;
 
 class Demo extends Columns
 {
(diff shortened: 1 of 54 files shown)


bedengler:
I am using content frequently. 
What would it be replaced with? 
How can we set the text of a view then? I'm using content, because it can contain html, whereas if we use the \Atk4\Ui\Text element, html will be converted into text.

mvorisek:
In all atk4/ui classes which use content property all works as before.

Bare `View` class however newly will not contain the content property.

So the question is, if you are using `View` class directly (with content property) or extend `View` class in your code (again with content property).

mkrecek234:
I am using View class often for simple views or Fomantic UI elements or fast prototyping of controls saving time to explicitly extend that class.

bedengler:
> In all atk4/ui classes which use content property all works as before.
> 
> Bare `View` class however newly will not contain the content property.
> 
> So the question is, if you are using `View` class directly (with content property) or extend `View` class in your code (again with content property).

I am using view directly. What about creating a new extended class named "\Atk4\Ui\Content" which can have the content property?
Btw: I don't get the reason, why it makes sense to remove the content property? What does it make better?

mvorisek:
> I am using view directly. What about creating a new extended class named "\Atk4\Ui\Content" which can have the content property?

That is how it is done - the class is named `ViewWithContent`.

PR #2076, opened 2023-06-19
Fix table column tooltip typecast

mkrecek234:


mvorisek:
The tooltip should be definitely typecasted. The coverage says it is covered, but I do not think the actual shown string is asserted, so this change should be covered with a Behat test (and preferably with non-string field).

mvorisek:
@mkrecek234 do you want to pursue this PR?

PR #2077, opened 2023-06-26
Fix typos

szepeviktor:
From https://github.com/atk4/data/issues/1104

@mvorisek


szepeviktor:
_Smoke (latest, CodingStyle)_ started failing 2 days ago. -> #2078

szepeviktor:
May I contribute a new CI workflow that regenerates (and commits) generated files?

mvorisek:
> May I contribute a new CI workflow that regenerates (and commits) generated files?

In past we have been using that, but it turned out more commits are harded to work with, so we have CI to assert all files are up-to-date, but you have to commit them manually from your side - you changed JS files, you need to rebuild them and commit.

szepeviktor:
> you changed JS files, you need to rebuild them and commit

Could you please do that for me this time?

mvorisek:
In `js/` run `npm install` and `npm build`.

szepeviktor:
Done.

szepeviktor:
It would be much better (no stomach pain for viktor) to set `engines` in package.json.

```json
{
    "engines": {
        "node": "^18.12",
        "npm": "^9.2.0",
        "yarn": "please-use-npm"
    }
}
```

And npm install should be [npm ci](https://docs.npmjs.com/cli/v9/commands/npm-ci) to leave the lock file unchanged.

mvorisek:
> It would be much better (no stomach pain for viktor) to set `engines` in package.json.
> 
> ```json
> {
>     "engines": {
>         "node": "^18.12",
>         "npm": "^9.2.0",
>         "yarn": "please-use-npm"
>     }
> }
> ```
> 
> And npm install should be [npm ci](https://docs.npmjs.com/cli/v9/commands/npm-ci) to leave the lock file unchanged.

we use `npm ci` in CI, with `npm install`, did the lock file change?

`node` / `npm` - why?

`yarn` - we should be compatible with `yarn`

szepeviktor:
> we use `npm ci` in CI

Here "ci" does not mean "continuous integration" but clean install.
The goal is to have generated files updated without package versions change.

demos/interactive/popup.php
```diff
@@ -172,7 +172,7 @@ public function linkCart(View $cart, JsExpressionable $jsAction = null): void
 //
 // This can happen when your popup content is non-trivial. So we are moving Popup into the app and linking up
 // the triggers. Now, since it's outside, we can't use a single jsAction to reload menu item (along with label)
-// and the contens. We could use 2 requests for reloading, but that's not good.
+// and the contents. We could use 2 requests for reloading, but that's not good.
```
mvorisek: content?

szepeviktor:
> with `npm install`, did the lock file change?

I was running only `npm ci`.

demos/interactive/popup.php
szepeviktor: Done.

mvorisek:
How viable is to add https://github.com/crate-ci/typos into our CI for atk4/{core, data, ui}? Can we run it without any false positives out of the box?

szepeviktor:
> `yarn` - we should be compatible with `yarn`

Here is how to do that.
```bash
yarn import # import npm's package-lock.json
yarn install --frozen-lockfile
```

szepeviktor:
> How viable is to add https://github.com/crate-ci/typos into our CI

That is a default workflow in my work.
https://github.com/szepeviktor/debian-server-tools/blob/master/webserver/Continuous-integration-Continuous-delivery-spelling.yml

I was planning to improve your CI workflows.
for example https://github.com/szepeviktor/debian-server-tools/blob/master/webserver/Continuous-integration-Continuous-delivery.yml

PR #2092, opened 2023-08-16
Fix Lookup initial change detection

### LINKED ISSUE #2020: Lookup default value is breaking canLeave
I have a fairly complex View including tabs, and inside the tabs I have a form which itself consists of Accordions and many default fields. The form is configured to have `$form->canLeave = false`. However, since recently, the onLeave feature does not work properly. Sometimes, the browser warns always, even though nothing was changed in the form. Sometimes, the browser does not warn at all.

Unfortunately, I cannot reproduce this at all currently.

mkrecek234:
fix #2020

### DIFF
--- a/src/Form/Control/Lookup.php
+++ b/src/Form/Control/Lookup.php
@@ -389,7 +389,7 @@ class Lookup extends Input
             $this->model = $this->model->loadBy($idField, $this->entityField->get());
 
             $row = $this->renderRow($this->model);
-            $chain->dropdown('set value', $row['value'])->dropdown('set text', $row['title']);
+            $chain->dropdown('set value', $row['value'], true)->dropdown('set text', $row['title'], true);
         }
 
         $this->js(true, $chain);


mvorisek:
canLeave feature needs Behat tests

mkrecek234:
Would be nice to have behat tests for canLeave (but then for ALL form controls which is an overkill for this PR to merge)  - this should not prevent merging this. If you read FUI documentation, it is clear that lookup was wrongly initialized before which is fixed now. Would be welcomed if you could contribute it if you consider it compulsory here

src/Form/Control/Lookup.php
mkrecek234, line 1: No change needed for another dropdown "set value" call though in https://github.com/atk4/ui/blob/4255c21b3c661dbed2090a7542bc7bc6ee86e4a1/src/Form/Control/Lookup.php#L276 - here a new record in the dropdown will be added and pre-selected. This indeed needs to fire the change event in the form.

mkrecek234:
@atk4/trusted-maintainers Can someone help providing the tests that @mvorisek requests for this PR to merge? This is a simple bug introduced which is fixed by the above PR. Very annoying as forms today will always prevent to be left when a lookup control is present. Thank you contributors!

mvorisek:
The change looks correct:

docs: https://github.com/fomantic/Fomantic-UI/blob/2.9.3/src/definitions/modules/dropdown.js#L2687 (and https://fomantic-ui.com/modules/dropdown.html#behavior)

@mkrecek234 did you manage to get Behat working locally as discussed on Discord?

mkrecek234:
Unfortunately no time, @mvorisek - but can confirm that the change is 100% correct, fixing the issue correctly and according to Fomantic UI specs.

mvorisek:
I have merged it, it does not worsen coverage but be aware it can break anytime as no tests.

PR #2095, opened 2023-08-23
Preserve full page path in App::url() method

abbadon1334:


mvorisek:
> if Atk4 is used with pretty urls routing, App::url in case of legitimate path like `https://domain/admin/users/` returns `https://domain/admin/users/index`

are you sure, isn't `.../index.php` returned?

> there was something in 4.0, removed in actual development.

please point to an exact commit/PR

and/or is something wrong with https://github.com/atk4/ui/pull/2065?

src/App.php
```diff
@@ -112,6 +112,15 @@ class App
...
     private $portals = [];
 
+    /**
+     * @var string used in method App::url to build the url
+     *
+     *  Used only in method App::url
+     *  If filename part is missing during building of url, this page will be used
+     *  Hint: if you use a routing system, you need to set this and urlBuildingExt to empty string
+     */
+    protected $urlBuildingPage = 'index';
```
mvorisek, line 122: Is `index.php` needed by some major webserver? If not, what about dropping it completely?

abbadon1334:
> and/or is something wrong with #2065?

At the moment i see only the dropping of App::$page in this commit  as a problem in case of routing

#### NOT using php routing

you are right about returning /index.php is the default behaviour of nearly every webserver but there is a directive configuration for every webserver:

- Apache [https://httpd.apache.org/docs/2.4/mod/directive-dict.html#Description](https://httpd.apache.org/docs/2.4/mod/directive-dict.html#Description)

- Nginx [http://nginx.org/en/docs/http/ngx_http_index_module.html](http://nginx.org/en/docs/http/ngx_http_index_module.html)

A standard Atk4 App using App::$urlBuildingPage and App::$urlBuildingExt can be set the correct return of App::url even for non standard configuration

#### using php routing

If you put in front of Atk for example https://github.com/nikic/FastRoute will manage the application routing.

Excepted results:
`App::url('/') = '/'`
`App::url('/admin/users/') = '/admin/users/'`

Actual results:
`App::url('/') = '/index'`
`App::url('/admin/users/') = '/admin/users/index'`

### Conclusion
I think the best option is not dropping it , but leave full control of default Index returning from building url, remove the needs of extend the App only to override one line in `App::url`

mvorisek:
> > and/or is something wrong with #2065?
> 
> At the moment i see only the dropping of App::$page in this commit as a problem in case of routing
> #### NOT using php routing

so `App::$page` was possible to be set as the newly introduced `urlBuildingPage`?

> you are right about returning /index.php is the default behaviour of nearly every webserver but there is a directive configuration for every webserver:
> 
>     * Apache https://httpd.apache.org/docs/2.4/mod/directive-dict.html#Description
> 
>     * Nginx http://nginx.org/en/docs/http/ngx_http_index_module.html

AFAK the only issue is with php internal webserver as `/virtual_route` is not supported if not routed via single script.

But I agree, badly configured Apache/nginx should still be supported.

-> as both major webserver name it "index", the property should be named with "index" too

-> what is your opinion on empty string to be the default?

> A standard Atk4 App using App::$urlBuildingPage and App::$urlBuildingExt can be set the correct return of App::url even for non standard configuration

> I think the best option is not dropping it , but leave full control of default Index returning from building url, remove the needs of extend the App only to override one line in `App::url`

-> needs phpunit test

abbadon1334:
> so `App::$page` was possible to be set as the newly introduced `urlBuildingPage`?

LGTM

abbadon1334:
> AFAK the only issue is with php internal webserver as `/virtual_route` is not supported if not routed via single script.
> 
> But I agree, badly configured Apache/nginx should still be supported.
> 
> -> as both major webserver name it "index", the property should be named with "index" too

i think default must be page=`index` ext=`.php` to give excepted result (index.php) for previous version + badly configured web server as you pointed.
 
> -> what is your opinion on empty string to be the default?

i think the correct default value is `index`

abbadon1334:
> -> needs phpunit test

i think is needed, i proceed adding functional test with App::url no behat or webserver only checking excepted output. What do you think?

abbadon1334:
@mvorisek, I may have gone a bit overboard with the test cases, but during the process, I discovered some unexpected behavior that didn't give me enough confidence to close the PR. As a result, I've refactored the entire `App::url` method. I found that it was becoming too complex, so I've broken it down into multiple private sub-methods for better readability and maintainability.

Additionally, I believe I've resolved an issue where, under certain conditions, the URL returned was missing the relative path.

i see the demos are broken due to recent changes in Control::set method

tests/AppTest.php
```diff
@@ -68,4 +70,141 @@ public function testUnexpectedOutputLateError(): void
...
+        $stickyGetArguments = [
+            '__atk_json' => false,
+            '__atk_tab' => false,
+        ];
+
+        foreach ($_GET as $key => $value) {
+            $stickyGetArguments[$key] = $value;
+        }
+
+        $app = new App([
```
mvorisek, line 182: AppTest has `CreateAppTrait` trait, use it :)

mvorisek:
> i see the demos are broken due to recent changes in Control::set method

I belive that is because of a new release of phpstan :) - please fix it first in a separate PR

> Additionally, I believe I've resolved an issue where, under certain conditions, the URL returned was missing the relative path.

then please split this PR to contain minimal fix + minimal "directory index" feature + minimal tests

once that is done, submit another refactoring PR incl. `$_SERVER['REQUEST_URI']` change, the change might look small, but it can be huge, a few months ago I was fixing one URL /w realpath & symlink resolving issue and it took me half day, so these changes needs to be landed separately to be able to bisect/revert them if needed

tests/AppTest.php
```diff
@@ -68,4 +70,141 @@ public function testUnexpectedOutputLateError(): void
...
+     * @param string|array<0|string, string|int|false> $page                URL as string or array with page name as first element and other GET arguments
+     * @param bool                                     $useRequestUrl       Simply return $_SERVER['REQUEST_URI'] if needed
+     * @param array<string, string>                    $extraRequestUrlArgs Additional URL arguments, deleting sticky can delete them
+     */
+    public function testUrlBuilding($page, bool $useRequestUrl, array $extraRequestUrlArgs, string $requestUrl, string $exceptedStd, string $exceptedCustom, string $exceptedRouting): void
+    {
+        $factory = new Psr17Factory();
+        $request = $factory->createRequest('GET', 'http://127.0.0.1' . $requestUrl);
+
+        $_SERVER = [
```
mvorisek: let's move this functional test to the refactoring PR too, setting global `$_SERVER` might have side effect on other tests in theory, better to discuss/fix later

tests/AppTest.php
abbadon1334: i think is better than createApp + setDefaults after __construct call, no?

tests/AppTest.php
abbadon1334: i know but even with injecting App::$request gives error in case of ($useRequestUrl = true) on $_SERVER['REQUEST_URI'] because it was not initialized

abbadon1334:
> > i see the demos are broken due to recent changes in Control::set method
> 
> I belive that is because of a new release of phpstan :) - please fix it first in a separate PR

Sure

> > Additionally, I believe I've resolved an issue where, under certain conditions, the URL returned was missing the relative path.
> 
> then please split this PR to contain minimal fix + minimal "directory index" feature + minimal tests
> 
> once that is done, submit another refactoring PR incl. `$_SERVER['REQUEST_URI']` change, the change might look small, but it can be huge, a few months ago I was fixing one URL /w realpath & symlink resolving issue and it took me half day, so these changes needs to be landed separately to be able to bisect/revert them if needed

for sure, trying to test the results of the previous App::url gives inconsistent response, but for sake of testing i can checkout the develop branch and add only unit tests and we can discuss the inconsistency in every cases

mvorisek:
the phpstan issues are probably because of https://github.com/phpstan/phpstan-src/pull/2540, see https://github.com/phpstan/phpstan/releases/tag/1.10.31

tests/AppTest.php
abbadon1334: Ok done

src/App.php
```diff
@@ -691,47 +691,83 @@ public function stickyForget(string $name): void
      * @param bool                                     $useRequestUrl       Simply return $_SERVER['REQUEST_URI'] if needed
```
mvorisek: what is the `$useRequestUrl` usecase?

it replaces `$page`, so the design does not seems to be right, I am inclined to remove it completely, is that possible and what usecases will it break?

tests/AppTest.php
```diff
@@ -67,4 +68,123 @@ public function testUnexpectedOutputLateError(): void
...
+    {
+        return [
+            // simple cases
+            [[], false, [], '/', '/index.php', '/default.php8', '/'],
+            [[], false, [], '/test/', '/test/index.php', '/test/default.php8', '/test/'],
+            [[], true, [], '/', '/index.php', '/default.php8', '/'],
+            [[], true, [], '/test/', '/test/index.php', '/test/default.php8', '/test/'],
+
+            // simple cases with query args in request
+            // [[], false, [], '/?test=atk4', '/index.php?test=atk4', '/default.php8?test=atk4', '/?test=atk4'],
```
mvorisek: @abbadon1334 why are these tests commented, shouldn't this PR fix all the URL issues? Or is #2101 needed for some reason?

tests/AppTest.php
```diff
@@ -67,4 +68,123 @@ public function testUnexpectedOutputLateError(): void
...
+            [[], true, ['extra_args' => 'atk4'], '/?test=atk4', '/index.php?extra_args=atk4&test=atk4', '/default.php8?extra_args=atk4&test=atk4', '/?extra_args=atk4&test=atk4'],
+            [[], true, ['extra_args' => 'atk4'], '/test/?test=atk4', '/test/index.php?extra_args=atk4&test=atk4', '/test/default.php8?extra_args=atk4&test=atk4', '/test/?extra_args=atk4&test=atk4'],
+
+            // simple cases with page as string
+            ['test', false, [], '/', 'test.php', 'test.php8', 'test'],
+            ['test/test/a', false, [], '/', 'test/test/a.php', 'test/test/a.php8', 'test/test/a'],
+            ['test/index', false, [], '/', 'test/index.php', 'test/index.php8', 'test/index'],
+            ['test/index', false, [], '/test/', 'test/index.php', 'test/index.php8', 'test/index'],
+            ['test', true, [], '/', '/index.php', '/default.php8', '/'],
+            ['test', true, [], '/request-url', '/request-url.php', '/request-url.php8', '/request-url'],
```
mvorisek, line 99: brokes testing on Windows

src/App.php
abbadon1334: i think you are right, was there but apparently is not used anywhere in codebase.

tests/AppTest.php
abbadon1334: No commented out cannot be resolved due to the inconsistency of App::url, In #2101 these tests are uncommented and App::url is refactored to give results consistency.

tests/AppTest.php
abbadon1334: I will check later on Windows the directory separator is probably messed up, but I think only in tests due to "emulation" of the webserver

tests/AppTest.php
mvorisek: in #2101 we only switch from $_XXX globals to request API, this PR already does some App::url changes, can the changes needed for consistent URLs/all tests passing be done in this PR (using $_XXX globals)?

src/App.php
```diff
@@ -661,42 +661,78 @@ public function url($page = [], bool $useRequestUrl = false, array $extraRequest
...
+        $lastChar = substr($pagePath, -1);
+        if ($lastChar === '/') {
+            return $prefix . $pagePath . $this->urlBuildingIndexPage . $this->urlBuildingExt;
+        }
+
+        if ($pagePath === '') {
+            return '/' . $pagePath . $this->urlBuildingIndexPage . $this->urlBuildingExt;
+        }
+
+        $pagePathPart = trim(dirname($pagePath), '.');
```
mvorisek: this is super wrong, we cannot just trim unlimited `.`

also `dirname('/xxx')` on Windows results `\`

tests/AppTest.php
mvorisek: to be fixed in https://github.com/atk4/ui/pull/2095#discussion_r1326492320 thread

src/App.php
mvorisek: addressed in https://github.com/atk4/ui/pull/2107

tests/AppTest.php
mvorisek: answer: #2101 is really not needed

src/App.php
```diff
@@ -657,37 +657,49 @@ public function stickyForget(string $name): void
...
+
+        $request = $this->getRequest();
+
+        if ($pagePath === null) {
+            $pagePath = $request->getUri()->getPath();
+        }
+
+        $prefix = '';
+        if (substr($pagePath, 0, 2) === '..') {
+            $prefix = '..';
```
mvorisek: this is untested and limited to one level only, if "dots" are supported, then "." must be supported as well

@abbadon1334 did you coded this by yourself or used a GPT to code it?

it took me 2 hours to clean your code and the tests...

abbadon1334: The code started as clear but added some conditional to solve some issues in demos, e.g. the `../` and the ./ which are the same as empty string due to browser behavior.
sorry @mvorisek, but due to the mess and tiredness, i run at the end rector for renaming variables and method names, it seems more clear and i left it, next time i commit the actual version and do a rector just to go back if there is any problem

tests/CallbackTest.php
```diff
@@ -89,8 +89,8 @@ public function testViewUrlCallback(): void
         $this->simulateCallbackTriggering($cbApp);
         $this->simulateCallbackTriggering($cb);
 
-        $expectedUrlCbApp = '?' . Callback::URL_QUERY_TRIGGER_PREFIX . 'aa=callback&' . Callback::URL_QUERY_TARGET . '=aa';
-        $expectedUrlCb = '?' . /* Callback::URL_QUERY_TRIGGER_PREFIX . 'aa=1&' . */ Callback::URL_QUERY_TRIGGER_PREFIX . 'bb=callback&' . Callback::URL_QUERY_TARGET . '=bb';
+        $expectedUrlCbApp = '/index.php?' . Callback::URL_QUERY_TRIGGER_PREFIX . 'aa=callback&' . Callback::URL_QUERY_TARGET . '=aa';
```
mvorisek, line 92: This change is not correct as the $_SERVER vars are not correctly set by `simulateCallbackTriggering` method.

Path on line https://github.com/atk4/ui/blob/f1277b7926/src/App.php#L673 is `''` instead of `'/'`, but in properly configured `$_SERVER`, the path must always start with `/`.

src/App.php
mvorisek: most non-BC-breaking changes were fixed/landed in #2108

tests/CallbackTest.php
mvorisek: TODO added in https://github.com/atk4/ui/commit/b3852843ae5387064ac03a3cf70a761ad70ece64#diff-ad46c7350003ccf7f086dccbecc6fcb7706a1558b14eb5e437a856b8c619e71fR673

src/App.php
mvorisek: @abbadon1334 please review the 4 last commits

The end changed LoC is minimal.

I have updated the PR title - is this what you wanted or something else/more?

abbadon1334: Switching title to "preserve" make sense even for tests without full path.
LGTM and thanks to go ahead on my work and do it better

PR #2099, opened 2023-09-04
Add favicon.ico to demos

abbadon1334:


mvorisek:
In many cases the `demos/` are not HTTP root path (`/`). You probably tried to fix missing favicon reported by a browser, in official hosted demos you should probably add the favicon in the Docker deploy script. If ok with you, let's close this PR.

abbadon1334:
> In many cases the `demos/` are not HTTP root path (`/`). You probably tried to fix missing favicon reported by a browser, in official hosted demos you should probably add the favicon in the Docker deploy script. If ok with you, let's close this PR.

you are right we already added it to deploy, i noticed it running the demos and i open the PR.

PR #2101, opened 2023-09-05
Replace $_GET/$_POST/$_FILES with App::getRequestXxx()

### LINKED ISSUE #2009: Direct access of $_GET/$_POST/$_FILES must be replaced with App::getRequestXxx()
TODO from https://github.com/atk4/ui/pull/1706#issuecomment-997432216 and https://github.com/atk4/ui/pull/1706#discussion_r1120561649

abbadon1334:
fix #2009


src/App.php
```diff
@@ -1193,4 +1224,82 @@ private function getRenderedPortals(): array
 
         return $portals;
     }
+
+    /**
+     * Return true if $_GET[$key] exists.
+     */
+    public function issetRequestGetParam(string $key): bool
+    {
+        return isset($this->request->getQueryParams()[$key]);
```
mvorisek, line 1233: will return false when null

src/App.php
```diff
@@ -1193,4 +1224,82 @@ private function getRenderedPortals(): array
 
         return $portals;
     }
+
+    /**
+     * Return true if $_GET[$key] exists.
+     */
+    public function issetRequestGetParam(string $key): bool
```
mvorisek, line 1231: `hasXxx` name LBTM

src/App.php
```diff
@@ -1193,4 +1224,82 @@ private function getRenderedPortals(): array
...
+    }
+
+    /**
+     * Return $_GET param by key or null if not exists.
+     *
+     * @return mixed
+     */
+    public function getRequestGetParam(string $key)
+    {
+        return $this->getRequestGetParams()[$key] ?? null;
```
mvorisek, line 1251: with https://github.com/atk4/ui/pull/2101/files#r1316302660 returning `true` for `null`, this should throw if the key is not present

src/App.php
```diff
@@ -1193,4 +1224,82 @@ private function getRenderedPortals(): array
...
+     */
+    public function getRequestGetParams(): array
+    {
+        return $this->request->getQueryParams();
+    }
+
+    /**
+     * Return $_GET param by key or null if not exists.
+     *
+     * @return mixed
```
mvorisek, line 1247: AFAK the GET can be string, null

and array of strings, but we do not want to support it probably using this method, as longer alternative exists :)

composer.json
```diff
@@ -53,7 +53,7 @@
     "require": {
         "php": ">=7.4 <8.3",
         "atk4/data": "dev-develop",
-        "nyholm/psr7": "^1.4",
+        "nyholm/psr7": "^1.6.0",
```
mvorisek, line 56:
```suggestion
        "nyholm/psr7": "^1.6",
```

looks fine - https://github.com/Nyholm/psr7/blob/1.6.0/CHANGELOG.md

make sure you change it also in the `require-release`!

demos/_unit-test/post.php
```diff
@@ -10,6 +10,14 @@
 /** @var \Atk4\Ui\App $app */
 require_once __DIR__ . '/../init-app.php';
 
+// Creation of request from globals depenmds on getallheaders
```
mvorisek: no hacks in individual files, this must be fixed in central place

if the comment about `getallheaders` is true, then https://github.com/Nyholm/psr7-server/blob/1.0.2/src/ServerRequestCreator.php#L52 needs to be replaced

src/App.php
```diff
@@ -1151,4 +1182,101 @@ private function outputResponseJson($data): void
         $this->setResponseHeader('Content-Type', 'application/json');
         $this->outputResponse($data);
     }
+
+    /**
+     * Return true if $_GET[$key] exists.
+     */
+    public function hasRequestGetParam(string $key): bool
```
mvorisek, line 1189: It seems you completely missed docs, they cannot be checked using CI, so for every change you commit, you must keep them in sync!

demos/_unit-test/post.php
abbadon1334: i move it to demos/init-app.php

i don't think the internal creation of ServerRequest must be changed, we need it only in unit tests running via CLI

src/App.php
abbadon1334: you are right, unit test failing makes me do some too fast cut/paste.
I update them

tests/DemosTest.php
```diff
@@ -502,7 +503,12 @@ public function provideDemoAssertJsonResponsePostCases(): iterable
      */
     public function testDemoAssertJsonResponsePost(string $path, array $postData): void
     {
-        $response = $this->getResponseFromRequest($path, ['form_params' => $postData]);
+        $response = $this->getResponseFromRequest($path, [
+            'headers' => [
+                'Content-Type' => 'application/x-www-form-urlencoded',
```
mvorisek: if `$postData` is an array (and in our case always is), the content-type should not be needed to be set explicitly - can this be fixed in the guzzle lib directly?

abbadon1334: From ServerRequestCreator

_POST must be loaded only if `Content-Type` equals to : `application/x-www-form-urlencoded` OR `multipart/form-data`

```php
    public function fromGlobals(): ServerRequestInterface
    {
        $server = $_SERVER;
        if (false === isset($server['REQUEST_METHOD'])) {
            $server['REQUEST_METHOD'] = 'GET';
        }

        $headers = \function_exists('getallheaders') ? getallheaders() : static::getHeadersFromServer($_SERVER);

        $post = null;
        if ('POST' === $this->getMethodFromEnv($server)) {
            foreach ($headers as $headerName => $headerValue) {
                if (true === \is_int($headerName) || 'content-type' !== \strtolower($headerName)) {
                    continue;
                }
                if (\in_array(
                    \strtolower(\trim(\explode(';', $headerValue, 2)[0])),
                    ['application/x-www-form-urlencoded', 'multipart/form-data']
                )) {
                    $post = $_POST;

                    break;
                }
            }
        }

        return $this->fromArrays($server, $headers, $_COOKIE, $_GET, $post, $_FILES, \fopen('php://input', 'r') ?: null);
    }
```

In Browser, this is done internally.
In Guzzle probably leave to the user all options, even bad formatted requests.
I think We can add Content-Type if POST refactoring it in our code: https://github.com/atk4/ui/blob/d64706e9f8cf3eab4b1b64cadea8eb1b6881774e/tests/DemosTest.php#L199

src/App.php
```diff
@@ -617,7 +618,7 @@ protected function createRequestPathFromLocalPath(string $localPath): string
             } else {
                 $request = \Symfony\Component\HttpFoundation\Request::createFromGlobals();
```
mvorisek, line 619: TODO for myself - this relies on $_GET, rework it in https://github.com/atk4/ui/pull/2095 and test by unsetting `$_GET`

src/App.php
mvorisek: test cases:

```
var_dump($app->hasRequestGetParam('x')); // test URL: ?x, passed, empty query value (without '=') is still non-null/string
var_dump($app->getRequestGetParam('x')); // test URL: ?x[]=1&x[]=2, TypeError: Return value of Atk4\Ui\App::getRequestGetParam() must be of the type string, array returned
```

src/App.php
```diff
@@ -26,6 +26,7 @@
 use Psr\Http\Message\ResponseInterface;
```
mvorisek, line 26: @abbadon1334 please explain this coverage change:

image

- it seems some error is newly triggered - how to repro it?
- it should NOT be triggered by these refactoring changes
- a Behat test to actually trigger it will be good, as a separate PR please

src/App.php
```diff
@@ -300,8 +301,8 @@ public function caughtException(\Throwable $exception): void
         // remove header
         $this->layout->template->tryDel('Header');
 
-        if (($this->isJsUrlRequest() || strtolower($_SERVER['HTTP_X_REQUESTED_WITH'] ?? '') === 'xmlhttprequest')
-                && !isset($_GET['__atk_tab'])) {
+        if (($this->isJsUrlRequest() || $this->request->getHeaderLine('HTTP_X_REQUESTED_WITH') === 'xmlhttprequest')
```
mvorisek, line 304: image

the removed `strtolower` caused 500 HTTP code on AJAX failure to be returned instead of 200

I will fix this, headers are case sensitive, so no `strtolower` is wanted but instead compare to `XMLHttpRequest` value

demos/_unit-test/post.php
mvorisek: discussed in https://github.com/atk4/ui/pull/2101#discussion_r1321152445

tests/DemosTest.php
mvorisek: Yes, the reason is https://github.com/Nyholm/psr7-server/blob/1.0.2/src/ServerRequestCreator.php#L62 - next time, please post a link instead of copy pasting the code here... Anyway, see https://github.com/atk4/ui/commit/6143477f79da6de711472bae89ffaeae48751ced, when all request headers are reconstructed into $_SERVER, no additional code is needed

src/App.php
```diff
@@ -320,6 +321,98 @@ public function getRequest(): ServerRequestInterface
...
+     */
+    public function tryGetRequestGetParam(string $key): ?string
+    {
+        return $this->request->getQueryParams()[$key] ?? null;
+    }
+
+    /**
+     * Get the value of a specific GET parameter from the HTTP request.
+     */
+    public function getRequestGetParam(string $key): string
```
mvorisek: @abbadon1334 do you prefer `getRequestGetParam` over `getRequestQueryParam`?

or what about `getRequestQuery($k)`/`getRequestPost($k)`?

demos/init-app.php
```diff
@@ -29,6 +29,12 @@
     'catchExceptions' => (bool) ($_GET['APP_CATCH_EXCEPTIONS'] ?? true),
     'alwaysRun' => (bool) ($_GET['APP_ALWAYS_RUN'] ?? true),
 ]);
+unset($_SERVER);
+unset($_GET);
+unset($_POST);
+unset($_FILES);
+unset($_COOKIE);
```
mvorisek: @abbadon1334 one last thing to address, this unset is breaking one Behat test

src/App.php
abbadon1334: laravel: `$request->query('name');`
slim: `$request->getQueryParam('page');`
Symfony: `$request->query->get('name');`

i think **Query** is the correct way of naming it, but i should leave **Param**.
So i think the best option is: `getRequestQueryParam($k)` and `getRequestPostParam($k)`

WDYT?

demos/init-app.php
abbadon1334: Am i wrong, or this is needed only for PHPUnit for continuous/fast calls without isolation of requests?

mvorisek: The `unset($_XXX)` there is for non-phpunit actually, it is there to make sure every atk4/ui component use the request details from App/request object. See the Behat failure, it can be easily reproduced localy /w any browser, if these unsets are present, the filtering does not work.

src/App.php
mvorisek: I agree. Renamed. `->getApp()->getRequestQueryParam()` seems however quite long, as I expect it to be used quite frequently, we can introduce shortcut method in View.

demos/init-app.php
mvorisek: please make the tests passing - but please do NOT force push the branch nor merge in develop - I will then do final review and merge

mvorisek:
Thank you @abbadon1334, quality PR!

PR #2106, opened 2023-09-10
Drop unused App portals code

mvorisek:
App portals objects are not stored since https://github.com/atk4/ui/pull/1771


src/App.php
```diff
@@ -1186,21 +1151,4 @@ private function outputResponseJson($data): void
...
         $this->outputResponse($data);
     }
-
-    /**
-     * Generated HTML and JS for portal view registered to app.
-     */
-    private function getRenderedPortals(): array
-    {
-        // prevent looping (calling App::terminateJson() recursively) if JsReload is used in Modal
-        unset($_GET['__atk_reload']);
```
mvorisek, line 1196: This PR does not replace this unset in any way, but it should be no longer needed, as tested since it was introduced in #1374.

PR #2128, opened 2023-11-10
Handle non-string tooltip description properly

mkrecek234:
Tooltips should allow for non-string fields, string-conversion missing, fixed by this PR

### DIFF
--- a/src/Table/Column/Tooltip.php
+++ b/src/Table/Column/Tooltip.php
@@ -72,7 +72,7 @@ class Tooltip extends Table\Column
 
         return [
             '_' . $field->shortName . '_data_visible_class' => '',
-            '_' . $field->shortName . '_data_tooltip' => (string) $tooltip,
+            '_' . $field->shortName . '_data_tooltip' => $this->getApp()->uiPersistence->typecastSaveField($row->getModel()->getField($this->tooltipField), $tooltip),
             '_' . $field->shortName . '_icon' => $this->icon,
         ];
     }
--- a/src/Table/Column/Tooltip.php
+++ b/src/Table/Column/Tooltip.php
@@ -72,7 +72,7 @@ class Tooltip extends Table\Column
 
         return [
             '_' . $field->shortName . '_data_visible_class' => '',
-            '_' . $field->shortName . '_data_tooltip' => (String) $tooltip,
+            '_' . $field->shortName . '_data_tooltip' => (string) $tooltip,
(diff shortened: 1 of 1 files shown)


mvorisek:
Michael, the `(string)` cast is forbidden to be used in the whole project. Can you please finally accept it?

Not even because we need to use UI typecast, but also because for ex. `(string) float type` can return incomplete precision.

src/Table/Column/Tooltip.php
```diff
@@ -72,7 +72,7 @@ public function getHtmlTags(Model $row, ?Field $field): array
 
         return [
             '_' . $field->shortName . '_data_visible_class' => '',
-            '_' . $field->shortName . '_data_tooltip' => $tooltip,
+            '_' . $field->shortName . '_data_tooltip' => $this->getApp()->uiPersistence->typecastSaveField($row->getModel()->getField($this->tooltipField), $tooltip),
```
mvorisek, line 75: Provide a proof you changes are consistent:

1. typecasting in all other places is done after/before? the if statement.
2. what is the 2nd field param?
3. how this change can be tested

mkrecek234: @mvorisek If you have an improvement to the submitted code, your submission would be highly welcomed. I cannot find another comparable getHtmlTags code as all others use getTag. At all other places, the if statement is based on non-typecasted values. It is obvious that the non-typecasted value will create issues, as this was missed to be corrected when type-strictness was increased. So happy if we can work together to solve this bug.

mkrecek234: Small tip for repro - you just add a decorator to any grid:
```
         $grid->addDecorator('any_field', new  \Atk4\Ui\Table\Column\Tooltip(
            [
                'tooltipField' => 'any_float_type_field',
                'icon' => 'bookmark red',
            ]));
```

mvorisek, line 75: how is this PR different to https://github.com/atk4/ui/pull/2076?

mkrecek234: You're right - we don't move along with bug-fixing, so issues are identified now again and again. Sorry. Would be nice to have this primitive bug fixed asap, so duplication in issue reporting and PR submission won't happen again.

mvorisek: closing in favor of the original PR

PR #2142, opened 2024-01-22
Always reload full executor container

### LINKED ISSUE #1928: Modal executor should (re)load whole modal incl. buttons
currently only modal inner content/view is r(e)loaded causing uncessary html/js generated like:

image

image

no modal content should be generated before displayed (when inner content/view is loaded via API) and once AJAX for (re)load is fired, whole content should be rendered/downloaded

this will improve performance when main page has a lot of modal executors and also allow greater executor buttons customizations per (re)loaded content/records

filter popup generates a lot code as well, maybe move it to AJAX completely as well

mvorisek:
fix #1928

also improve performance as analysed in https://github.com/atk4/ui/issues/1976#issue-1552358090

the only known negative is no loader before modal/panel is shown, this is expected and in the future, it should be solved by a loader added to the triggering element (like button), the triggering element should be probably also disabled until the action is complete


DarkSide666 (approved):
I'm a bit left behind of late progress of ATK and UI in particular so hard to say if all is good here, but on first glance looks OK.

Only thing I would like to add here (or later in some future PR) is as you already said - _in the future, it should be solved by a loader added to the triggering element (like button), the triggering element should be probably also disabled until the action is complete_
This is quite essential to have especially in cases if modal is slow to load and it should be restricted to call it multiple times in parallel.

PR #2149, opened 2024-01-30
Minor table columns improvements

### LINKED ISSUE #1811: Review all `array_merge_recursive` usages
`array_merge_recursive` can merge the data not as expected, review all usages in this repo. All other major repos has no use, thus fine.

related with https://github.com/atk4/data/issues/1046

mvorisek:
also fix #1811


src/Table/Column.php
```diff
@@ -272,16 +271,15 @@ public function addClass($class, $position = 'body')
     }
 
     /**
-     * Adds a new attribute to the cells of this column. The optional second argument may be "head",
-     * "body" or "foot". If position is not defined, then attribute will be applied on all cells.
```
mvorisek, line 276: since introduction the comment is wrong - https://github.com/atk4/ui/blob/35052dddd9/src/Column/Generic.php#L21

PR #2157, opened 2024-02-03
Fix deprecated Node 16 in CI

mvorisek:
needs https://github.com/codecov/codecov-action/issues/1262


mvorisek:
replaced by #2227

PR #2158, opened 2024-02-03
Require iterable types in phpstan

mvorisek:
finish once https://github.com/phpstan/phpstan/issues/8438 is implemented


mvorisek:
replaced by https://github.com/atk4/ui/pull/2221

PR #2175, opened 2024-02-23
Fix ob_end_flush() in JsSse init

Vladzimir:
https://gist.github.com/connerbw/a2fa7712efe135d5854f4d32d67ca09f#file-eventemitter-php-L114

### DIFF
--- a/src/JsSse.php
+++ b/src/JsSse.php
@@ -193,6 +193,5 @@ class JsSse extends JsCallback
         while (ob_get_level() > 0) {
             ob_end_flush();
         }
-        flush();
     }
 }
--- a/src/JsSse.php
+++ b/src/JsSse.php
@@ -190,8 +190,7 @@ class JsSse extends JsCallback
         }
 
         // prevent buffering
-        $levels = ob_get_level();
-        for ($i = 0; $i < $levels; $i++) {
+        while (ob_get_level() > 0) {
             ob_end_flush();
         }
         flush();
--- a/src/JsSse.php
+++ b/src/JsSse.php
@@ -190,8 +190,10 @@ class JsSse extends JsCallback
         }
 
         // prevent buffering
-        if (ob_get_level()) {
+        $levels = ob_get_level();
+        for ($i = 0; $i < $levels; $i++) {
             ob_end_flush();
         }
+        flush();
     }
 }


src/JsSse.php
```diff
@@ -190,8 +190,10 @@ protected function initSse(): void
         }
 
         // prevent buffering
-        if (ob_get_level()) {
+        $levels = ob_get_level();
+        for ($i = 0; $i < $levels; $i++) {
             ob_end_flush();
         }
+        flush();
```
mvorisek: please explain why `flush()` is helpful

also note there is `self::flush()`

mvorisek, line 197:
```suggestion
        while (ob_get_level() > 0) {
            ob_end_flush();
        }
        flush();
```

src/JsSse.php
```diff
@@ -190,8 +190,9 @@ protected function initSse(): void
         }
 
         // prevent buffering
-        if (ob_get_level()) {
+        while (ob_get_level() > 0) {
             ob_end_flush();
```
mvorisek, line 194: please make all tests passing

Vladzimir: > please make all tests passing

Sorry, but I can't do this

src/JsSse.php
Vladzimir: > please explain why `flush()` is helpful

I don't know. The code is taken from the example I gave. If you think it is not needed, delete it.

src/JsSse.php
mvorisek: Obviously, we cannot merge anything that would break the whole CI :)

Might I know your usecase?

Vladzimir: I use the framework https://github.com/f3-factory/fatfree-core And SSE stops working in it.
Or are you talking about something else?

mvorisek: The fix seems legit. Upper layer/framework can open more than 1 output buffer. I even wonder why the atk4 tests passed when they rely on ob. In anycase, now the tests needs to be fixed to pass.

Vladzimir: Unfortunately, I'm not good at testing.

mkrecek234:
Hi there, I have the issue that output buffering is still active with any of the old or the new suggested code - I am running PHP-FPM on Apache, and it looks as if none of the code will stop the buffering. All scripts only transmit upon finish, and not in between. Any idea what to do in that situation with FPM?

mvorisek:
@mkrecek234 open a separate issue for that and link https://github.com/php/php-src/pull/12785#discussion_r1412054712 there. This PR should be solely for the `ob_end_flush()` unroll as it is not easy to pass testing on itself.

mvorisek:
we should fix it

src/JsSse.php
mvorisek: This was the cause of CI failure, as it prevented HTTP code to be set later.

Does this PR now fixes your issue with output buffering together with https://github.com/atk4/ui/pull/2175#discussion_r1501397309 framework you use?

src/JsSse.php
mvorisek: see https://github.com/atk4/ui/pull/2175#discussion_r1535566245 for the cause

PR #2187, opened 2024-03-25
Fix SSE flush for Apache PHP FPM

### LINKED ISSUE #2178: JsSse in PHP-FPM setup requires special configuration
Situation:

- You are using JsSse to send server-side updates to clients
- You. are having an environment with Apache and PHP-FPM, buffering pages
- The default in PHP-FPM under Apache is output buffering which destroys the possibility to use Sse, as the request will only be sent out to client if complete (or buffer size of typically 4096 is reached)

Solution:
- Configure PHP-FPM in Apache module to allow flush by this configuration in php-fpm.conf for example:

```
    <FilesMatch ".+\.ph(?:ar|p|tml)$">
        SetHandler "proxy:unix:/run/php/php8.2-fpm.sock|fcgi://localhost"
    </FilesMatch>

    <Proxy "fcgi://localhost">
        ProxySet flushpackets=on
        ProxySet flushwait=20
        ProxySet max=10
        ProxySet timeout=100
    </Proxy>
```
Suggestions to limit flushing packets only on selected scripts or directories are welcome.  This way you can keep output buffering for scripts that are non-Sse.

Also see:
https://github.com/php/php-src/pull/12785#discussion_r1412054712

mvorisek:
fix #2178

### DIFF
--- a/demos/interactive/console.php
+++ b/demos/interactive/console.php
@@ -127,10 +127,12 @@ $tab = $tabs->addTab('Use after form submit', static function (VirtualPage $vp) 
     $form->addControl('bar');
 
     $console = Console::addTo($vp, ['event' => false]);
-    $console->set(static function (Console $console) {
-        $model = $_SESSION['atk4_ui_console_demo'];
+    $console->set(static function (Console $console) use ($form) {
+        $entity = $form->model;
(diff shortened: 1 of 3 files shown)


mkrecek234:
@mvorisek Can confirm flushing works now properly with this fix, so no longer required to adjust the PHP-FPM configuration

PR #2192, opened 2024-03-26
Use View::$entity property for entity access

mvorisek:
In 99% cases the impact is on `Form::$model` property usage. Seach `form->model` in your codebase and replace it with `form->entity`.

The motivation is to simplify model/entity orientation for the developer. When Model is guaranteed to be entity only, is should never be named model.

### DIFF
--- a/docs/form-control.md
+++ b/docs/form-control.md
@@ -377,22 +377,22 @@ If you want to customize how a record is displayed and/or add an icon, Dropdown 
 This function is called with each model record and needs to return an array:
 
 ```
-$dropdown->renderRowFunction = function (Model $record) {
+$dropdown->renderRowFunction = function (Model $entity) {
     return [
-        'title' => $record->getTitle() . ' (' . $record->get('subtitle') . ')',
+        'title' => $entity->getTitle() . ' (' . $entity->get('subtitle') . ')',
     ];
-}
+};
 ```
 
 You can also use this function to add an Icon to a record:
 
 ```
-$dropdown->renderRowFunction = function (Model $record) {
+$dropdown->renderRowFunction = function (Model $entity) {
     return [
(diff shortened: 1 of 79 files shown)


docs/form-control.md
```diff
@@ -422,12 +422,12 @@ class MyDropdown extends \Atk4\Ui\Dropdown
...
 
 ```
-function (Model $record) {
+function (Model $entity) {
     return [
-        'title' => $record->getTitle,
-        'icon' => $record->value > 100 ? 'money' : 'coins',
-        'someOtherField' => $record->get('SomeOtherField'),
-        'someOtherField2' => $record->get('SomeOtherField2'),
+        'title' => $entity->getTitle,
```
DarkSide666: getTitle()

docs/grid.md
```diff
@@ -152,7 +152,7 @@ to populate a content:
 ```
 $grid->addModalAction('Details', 'Additional Details', function (View $p, $id) use ($grid) {
     // $id of the record which was clicked
-    // $grid->model = $grid->model->load($id);
+    // $model = $grid->model->load($id);
```
DarkSide666, line 155: no need for $model variable at all

DarkSide666 (commented):
Minor comments, but didn't review all toughly.

I guess i understand the purpose, but it will make life a bit more complicated for programmer to remember that form views have entity, but grid views have model.
I do not have a lot of up to date code, but time ago using $form->model was quite a common case. Especially in form submit callback.

docs/form-control.md
mvorisek: good catch!

docs/grid.md
mvorisek: docs are not always to be analysed :)

PR #2200, opened 2024-04-07
Allow extending Form/Control/Upload class by using different JS plugin

harish2704:
This PR allow us to create custom upload widgets by extending Upload or UploadImage class. 

( Context: I tried to create CroppedUploadImage widget class which will allow user to crop the image before upload )

### DIFF
--- a/src/Form/Control/Upload.php
+++ b/src/Form/Control/Upload.php
@@ -19,6 +19,8 @@ class Upload extends Input
 
     public string $inputType = 'hidden';
 
+    public string $jsPluginName = 'atkFileUpload';
+
     /**
      * The uploaded file ID.
      * This ID is return on form submit.
@@ -173,7 +175,7 @@ class Upload extends Input
 
                 if (count($postFiles) > 0 && reset($postFiles)['error'] === 0) {
                     $this->addJsAction(
-                        $this->js()->atkFileUpload('updateField', [$this->fileId, $this->getInputValue()])
+                        $this->js()->{$this->jsPluginName}('updateField', [$this->fileId, $this->getInputValue()])
                     );
                 }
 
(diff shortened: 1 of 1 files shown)


mvorisek (changes requested):
In general, it is not good practice to support different jQuery method name as it will address an usecase only when the accepted args are the same.

As you already can/registered jQuery method in your application, instead of registering a new one, you should override the atk4 jQuery `atkFileUpload` method to call your extended function possibly if some other conditions are met.

harish2704:
@mvorisek : Thank for your quick response. 

Regarding your comments, 
>In general, it is not good practice to support different jQuery method name as it will address an usecase only when the accepted args are the same.

Yes, I do agree that it is not a perfect choice or practice, but in this way at-least we get some extensibility through which at-least some of the usecases can be fulfilled . 
The current situation worse than this . 

> As you already can/registered jQuery method in your application, instead of registering a new one, you should override the atk4 jQuery atkFileUpload method to call your extended function possibly if some other conditions are met. 

IMHO, it is more hackish method than the suggested one. 
First of all, arguments passed to this plugin call are fixed. More hacks will be required to create "matching conditions"
Second, because of the previous issue, the default behavior of existing upload or uploadImage may break if the customization is in effect
Third, I think some JS level monkey patching will be required to override the registered JS plugin ( I am not sure about it )

In short, The suggested change at-least opens a way for others to extend functionality by writing a plugin following same public interface ( function arguments ) even though it is not a perfect solution

mvorisek:
I agree, but we need something conceptually working for every component, otherwise we will hack more and more.

If you do not want to solve this in JS now, you can override Upload::js() method to return a custom JS chain and replace (first chain call) "atkFileUpload" there.

PR #2209, opened 2024-06-01
Impl. clearable dropdown

### LINKED ISSUE #572: Lookup field with empty value doesn't show placeholder
One thing i don't like with Lookup field is that if it has placeholder and you select "..." value then it shows that ... value and not make it blank and show placeholder again.

It's small glitch, but annoying :)

### LINKED ISSUE #2026: Dropdown/lookup field shows empty value for required model fields
If we have model `hasOne` field , then in form it shows red asterisk near dropdown/lookup field in case `required => true` is set in model field.
BUT dropdown still contains empty value option in it.
image

I think empty value option should not be there if field is required, then we should force to fill it already in Form. Otherwise if it's left blank in Form and you try to save it will throw big, red exception.

----

And another related, but opposite issue.
If model field is defined like this
```
$this->addField('color', [
    'enum' => ['green', 'red', 'blue', 'orange', 'yellow', 'purple', 'gray'],
]);
```
Then in Form it shows dropdown/lookup field, but there is no empty option in there. Model field is not set as required, so empty option should be available in dropdown. Asterisk do not show too (which is correct).

mvorisek:
simillar to https://github.com/atk4/ui/pull/2063/commits/6fcc1f225c0b2f0095d8d4106b1ed92ba6b37a2e

fix #572
fix #2026


mkrecek234:
Hi @mvorisek With this PR, Lookup fields always get a clearable icon if they are either nullable or not required. I am of the opinion that 

a) clearable should only be activated by default if the field is not required and is nullable at the same time
b) it should still be possible to override this default behaviour using setDropdownOption('clearable', false) - currently not possible as it is forced on rendering.

Furthermore I saw that on a disabled/readOnly field with `multiple = true, the clearable icon was present allowing the erase the dropdown field.

Let me know if you agree, then I can work on a PR to suggest a change.

mvorisek:
Hi Michael, "almost always clearable" - what are the usecases? User should be able to revert any chenge he did on the UI side, so after selecting an option it should be always possible to clear it. The validation on send is of course always working.

"multiple" - if there is an issue, PR with testcase highly welcomed!

rickyosser:
> Hi Michael, "almost always clearable" - what are the usecases? User should be able to revert any chenge he did on the UI side, so after selecting an option it should be always possible to clear it. The validation on send is of course always working.
> 
> "multiple" - if there is an issue, PR with testcase highly welcomed!

@mvorisek it's all well as long as you actually use a model to populate the dropdown. In my case I'm not using any models at all in my current project as all DB/data related handling is done through a rest-api. I'm just using atk4 for building the GUI. In that case creating a dropdown using setSource I will not be able to add validation in the model, I could use a check in the callback but as I'm not really interested in what the user chooses, as any value except no value is acceptable. I could use radio-buttons instead but that uses more screen-area than needed to have a minimal and consistent look.

I would prefer to have setDropdownOption('clearable', false) honoured.
I will create a PR for it so you can check if it's ok to implement. (Calling @mkrecek234 for comments as well.)

mkrecek234:
@rickyosser Fully agree that we should still be able to programmatically configure the clearable setting, independent of the default, model-based logic. There are some use cases where the data structure allows nullable and non-required values, but you want the UI to force the user to choose a value. So needs to be independently configurable for sure.

PR #2211, opened 2024-06-17
Add "smallint" and "bigint" support to FilterModel

### LINKED ISSUE #2210: FilterModel bug with in 5.2 release
Simple bug, but unfortunately GitHub won't let me create a new branch anymore, so here in quick words:

- Create grid with $grid->addFilterColumn() where one of the columns is a bigint ID.
- FilterModel https://github.com/atk4/ui/blob/12b9e3d21dd3764caea10f9737ab2d396f941aea/src/Table/Column/FilterModel.php#L52 misses the bigint type assignment and thus it throws an error.

Just add 
```
            Types::INTEGER => FilterModel\TypeNumber::class,
```
to `$class` array and all works fine again.

mkrecek234:
fix #2210 and wrapped numeric type filtering in general

related with https://github.com/atk4/data/pull/1223 and #2208


mvorisek:
Well, it is not strictly a bug, but I will say the impact is not minimal as all IDs are bigint by default since v5.2.

Please investigate on what simplest demo the ID can be filtered and tested. We need a test before we merge it.

mvorisek:
@mkrecek234 please test and review

PR #2219, opened 2024-08-18
Add reveal eye to password form control

rickyosser:
This is an addition to the Password form-control which adds an eye to toggle password visibility.

The JS-script part maybe should be more ATK4 friendly but that is above my current knowledge, I will learn with help.


rickyosser:
Hi,
I understand the codingstyle errors and stuff but need help to decode the StaticAnalysis errors.
1, Overrride of the init() function, this might not be what I want to do, how would I go about to implement this as an addition/extend which will run automatically if the variable $this->eye is set to true?
2, I don't understand this at all, please enlighten me...

Thanks in advance.
Rickard

src/Form/Control/Password.php
```diff
@@ -4,7 +4,44 @@
...
 class Password extends Line
 {
     public string $inputType = 'password';
+
+    /**
+     * @var bool by setting this variable to true you get an Eye-button on the right side which can toggle the visibility of the entered password
+     */
+    public $eye;
+
+    protected function init(): void
```
mvorisek, line 17: just do what the PHPStan tells you here :) (see other init() overrides)

for the second PHPStan error - I think you do not need a variable and do not want to what the JsExpression in an array

src/Form/Control/Password.php
```diff
@@ -4,7 +4,44 @@
...
+use Atk4\Ui\Js\JsExpression;
+
 class Password extends Line
 {
     public string $inputType = 'password';
+
+    /**
+     * @var bool by setting this variable to true you get an Eye-button on the right side which can toggle the visibility of the entered password
+     */
+    public $eye;
```
mvorisek:
```suggestion

    /** By setting this variable to true you get an Eye-button on the right side which can toggle the visibility of the entered password */
    public bool $eye;
```

src/Form/Control/Password.php
```diff
@@ -4,7 +4,44 @@
 
 namespace Atk4\Ui\Form\Control;
 
+use Atk4\Ui\Button;
+use Atk4\Ui\Js\JsExpression;
+
 class Password extends Line
```
mvorisek, line 9: Based on https://github.com/whatwg/html/issues/7293 it is true that native solution does not exist in all browsers currently.

IE - at least https://stackoverflow.com/questions/61449079/how-to-hide-the-eye-from-a-password-input-in-ms-edge-and-ie#61450596 is needed to supress the native behaviour of IE, or do we want to keep native eye where supported?

Firefox - https://connect.mozilla.org/t5/ideas/add-a-reveal-password-function-for-logins/idi-p/1140 - did you know there is native (but I agree not much intuitive) reveal - can we turn on/off the native one?

Chrome - it seems there is no native solution (except when using 3rd party extension)

rickyosser: I think I meant a more native ATK4 solution with more Js functions instead of a big javascript blob... :)

src/Form/Control/Password.php
rickyosser: Thanks, I fixed this.

src/Form/Control/Password.php
rickyosser: Fixed...

src/Form/Control/Password.php
```diff
@@ -4,7 +4,43 @@
...
+        if ($this->eye) {
+            $button = new Button([
+                'class.grey' => true,
+                'iconRight' => 'eye slash',
+            ]);
+
+            $this->action = $button;
+
+            $button->on('click',
+                        new JsExpression("
```
mvorisek, line 31:
```suggestion
            $button->on('click', new JsExpression(<<<'EOF'
                ...
```

prefered coding style

src/Form/Control/Password.php
```diff
@@ -4,7 +4,45 @@
...
+                'class.grey' => true,
+                'iconRight' => 'eye slash',
+            ]);
+
+            $this->action = $button;
+
+            $button->on(
+                'click',
+                new JsExpression(<<<EOF
+		    var input_element = document.querySelector(`[id$='form_layout_{$this->shortName}_input']`);
```
mvorisek: using `shortName` to refering anything is wrong - see how it is done in other Form Controls

src/Form/Control/Password.php
```diff
@@ -4,7 +4,45 @@
...
 
+use Atk4\Ui\Button;
+use Atk4\Ui\Js\JsExpression;
+
 class Password extends Line
 {
     public string $inputType = 'password';
+
+    /** By setting this variable to true you get an Eye-button on the right side which can toggle the visibility of the entered password */
+    public bool $eye;
```
mvorisek: TODO for myself - should we use `false` or `null` by default

or `true` as default?

rickyosser: I'm happy to put false as the default, this is from my point of view an option. 
But if there is a vote and true is selected as default, that's an easy change.

src/Form/Control/Password.php
rickyosser: I'm very unexperienced in ATK4 source-code writing, how would I go about to find the name of the input field I want to use in a Form Control? The only cases where it's used is:
Radio
Checkbox
Textarea

src/Form/Control/Password.php
```diff
@@ -4,7 +4,46 @@
...
```
mvorisek, line 15:
```suggestion
    public bool $revealEye;
```

src/Form/Control/Password.php
```diff
@@ -4,7 +4,46 @@
...
+                'class.grey' => true,
+                'iconRight' => 'eye slash',
+            ]);
+
+            $this->action = $button;
+
+            $button->on(
+                'click',
+                new JsExpression(
+                    <<<EOF
```
mvorisek:
```suggestion
                    <<<'EOF'
```

and pass all strings using JsExpression templation to prevent JS injections

why not `getElementById`?

also rename `input_element` to `inputElem`

add space after `if`

use `===` for cmp

and use 4 spaces indentation in the JS if bodies

Then the 1st round of review should be done and we can focus on the impl.

Also, a Behat (functional) test for testing the reveal functonality is needed.

rickyosser: It's all good and you will have to guide me... :)

Anyway, yes,
1, getElementById works for the input field, how can I get the ID of the icon connected to the same field, it has a different ID.

I think I need more information on a lot of things, like Behat for exmple.

mvorisek: > and pass all strings using JsExpression templation to prevent JS injections

see https://github.com/atk4/ui/blob/5.2.0/src/Js/JsExpression.php#L35, ie. `new JsExpression('let str = 'x' + [] + 'y';', ['arg'])`

> Also, a Behat (functional) test for testing the reveal functonality is needed.

see https://github.com/atk4/ui/tree/develop/tests-behat directory, create demo with password file and `password.feature` test which will check:
1. password is hidden
2. click reveal
3. password is shown
4. click reveal again
5. password must be hidden again

you can either setup a testing environment on your PC or use the CI to make it passing using trial and error

src/Form/Control/Password.php
mvorisek: see https://github.com/atk4/ui/blob/5.2.0/src/Form/Control/Input.php#L111 and https://github.com/atk4/ui/blob/5.2.0/src/Form/Control/Dropdown.php#L158

so the correct way to get the ID is `$this->name . '_input'`

it is guaranteed to be unique

src/Form/Control/Password.php
rickyosser: Thanks for the info, I'm going to try and see what I can do about Behat, I've never used it before so....

Anyway, do you or anyone else have an idea on how to get the "ID" of the icon of the button?
In the code it looks like:
```
<i id="_b7947b56bcedc631d38d468ce0f74cda__voip_password_button_icon" class="  eye slash icon" style=""></i>
```

src/Form/Control/Password.php
mvorisek: the 2nd template value must not use `shortName` too, as the ID can be based on shortened identifier, so the assumption with `shortName` is not guaranteed to be always true

src/Form/Control/Password.php
mvorisek: ID - button has `->name` too, the same for inner icon probably

rickyosser: @mvorisek  I'm aware of it, that's why I asked about it before your latest message.

 I checked the structure of the object but name and shortName are empty.
I really need help here to find the id of the button without using the "shortName" which is a part of the id.

Thanks in advance!


``` 
object(Atk4\Ui\Button)#2346 (27) {
  ["defaultTemplate"]=>
  string(11) "button.html"
  ["ui"]=>
  string(6) "button"
  ["icon"]=>
  NULL
  ["iconRight"]=>
  string(9) "eye slash"
  ["_jsActions":protected]=>
  array(0) {
  }
  ["model"]=>
  NULL
  ["region"]=>
  NULL
  ["class"]=>
  array(1) {
    [0]=>
    string(4) "grey"
  }
  ["style"]=>
  array(0) {
  }
  ["attr"]=>
  array(0) {
  }
  ["template"]=>
  NULL
  ["content"]=>
  NULL
  ["element"]=>
  string(3) "div"
  ["executorFactory":protected]=>
  NULL
  ["stickyArgs"]=>
  array(0) {
  }
  ["defaultName"]=>
  string(3) "atk"
  ["_addLater":protected]=>
  array(0) {
  }
  ["_rendered":protected]=>
  bool(false)
  ["_app":"Atk4\Ui\AbstractView":private]=>
  NULL
  ["maxNameLength"]=>
  int(60)
  ["uniqueNameHashes"]=>
  array(0) {
  }
  ["elements"]=>
  array(0) {
  }
  ["_elementNameCounts":"Atk4\Ui\AbstractView":private]=>
  array(0) {
  }
  ["_initialized":"Atk4\Ui\AbstractView":private]=>
  bool(false)
  ["name"]=>
  NULL
  ["_owner":"Atk4\Ui\AbstractView":private]=>
  NULL
  ["shortName"]=>
  NULL
}
```

mvorisek: It is because the `$button` is not initialized, it needs to be part of the render tree first.

mvorisek: Looking into https://github.com/atk4/ui/blob/5.2.0/src/Form/Control/Input.php#L167 I belive you can replace `new Button(...)` with `Button::addTo($this, ...)`. As such call is in `$this::init()`, `addTo(` will call the init on the button and after the call the button should have the ID present.

tests-behat/revealEye.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Password with revealEye
+
+  Scenario:
+    Given I am on "form-control/input2.php"
+    When I fill in "password_re_norm" with "123"
+    Then "password_re_norm" should contain text "***"
```
mvorisek: A little unrelated - you found a bug with Behat IMO - missing step should emit non-zero exit code. It might be worth to investigate and report to Behat/Mink. Currently, only the "lowest" test pipelines can detect such issue correctly and fail the CI.

src/Form/Control/Password.php
rickyosser: I will try your find, I solved it by calling $this->renderView() for now but I think your find is more elegant if it works.

rickyosser: Ok, I tried it and yes, it renders the button directly but gets a tag not found error on "content"... So, I'm sticking with the working solution for now.

tests-behat/revealEye.feature
rickyosser: I have no clue at all on what you wrote. I'll take it at face value.
I'm not even sure the script works as I don't know if the field returns "***" to behat?

I wouldn't even know where to start with trying to report a bug in this case... :)

demos/form-control/input2.php
```diff
@@ -26,6 +26,16 @@
...
 
+$group = $form->addGroup('Password');
+$group->addControl('password_norm', [Form\Control\Password::class], ['type' => 'text'])->set('editable');
+$group->addControl('password_read', [Form\Control\Password::class, 'readOnly' => true])->set('read only');
+$group->addControl('password_disb', [Form\Control\Password::class, 'disabled' => true])->set('disabled');
+
+$group = $form->addGroup('Password with revealEye');
+$group->addControl('password_re_norm', [Form\Control\Password::class, 'revealEye' => true], ['type' => 'text'])->set('editable');
+$group->addControl('password_re_read', [Form\Control\Password::class, 'revealEye' => true, 'readOnly' => true])->set('read only');
+$group->addControl('password_re_disb', [Form\Control\Password::class, 'revealEye' => true, 'disabled' => true])->set('disabled');
```
mvorisek: I like this demo a lot!

Based on testing in FF with native reveal using context menu, "disabled" form control should have the reveal button disabled. This is consistent with normal text control which cannot be copied. - UPDATE: implemented

tests-behat/revealEye.feature
mvorisek: I do not know where to report it too ATM :)

Exit code - CLI exit code returned of the Behat CLI application

> I'm not even sure the script works as I don't know if the field returns "***" to behat?

Behat detected some grammar that you have written but not recognized.

It takes some time to understand Behat/Mink, that is normal. In linux, you can setup the testing environment quite easily, in Windows, it is a little more complicated as you need to convince all the drivers to work with your Chrome browser.

rickyosser:
Nice changes!
I wondered how to find the element id, now I have an example for future use, thank you!

src/Form/Control/Password.php
mvorisek: My preference is to make this function enabled by default with as much as possible native look & feel.

I mean - not using `Input::$action` as someone else might want to use the right icon for regular input action.

Small not backgrounded icon will then look perfectly.

And the "slash" on the icon should be inverted - it should be based on the "actionable action", not on the current state.

tests-behat/revealEye.feature
rickyosser: As a matter of fact, I run only Linux... Have done since 1992... I just never automated tests...

mvorisek: That saves you a lot of troubles - see https://github.com/atk4/ui/blob/5.2.0/.github/workflows/test-unit.yml#L370-L378 - with more or less equivalent code you should be able to run the Behat on your local computer. The run command itself is a few lines below - https://github.com/atk4/ui/blob/5.2.0/.github/workflows/test-unit.yml#L383-L384

rickyosser: Ok, I'll check it out...

src/Form/Control/Password.php
rickyosser: I do get where this is coming from, I'm not sure I know enough of css/js and atk4 to implement it though.
And yes, the icons was second pick because it's what is in Fomantic. 

I've found these on https://www.iconfinder.com/ which are free to use. they would be a better pick:
eye_closed_light_icon
eye_light_icon

mvorisek: We should stay with https://fomantic-ui.com/elements/icon.html#icon-set set only, you can open a feature request by them if the 1st icon is popular enough. For now, we should be fine with "eye slash" for unreveal.

rickyosser: Works for me! :)

I'll do a request in the near future...

rickyosser:
@mvorisek, I copied Password.php to Password2.php and added another test-group for password.
You can review both versions side-by-side. If the new version is better I'll replace the original version.

rickyosser:
> @mvorisek, I copied Password.php to Password2.php and added another test-group for password. You can review both versions side-by-side. If the new version is better I'll replace the original version.

Ok, an unintended difference with this new approach is that the whole field is clickable. So, by clicking the field it toggles it, not only the icon.

There is another approach which is to add another button to the input field, there is a variant of this in the decorated input demo.

mvorisek:
On the right side, there should be transparent button with the reveal icon. And keep the possibility for regular right action/button.

IDK if there is Fomantic-UI style for it, if yes, great, if not, lets add minimal HTML/CSS and see if we can contribute it then into Fomantic-UI repo.

rickyosser:
> On the right side, there should be transparent button with the reveal icon. And keep the possibility for regular right action/button.
> 
> IDK if there is Fomantic-UI style for it, if yes, great, if not, lets add minimal HTML/CSS and see if we can contribute it then into Fomantic-UI repo.

Ok, I rewrote parts of it to not use the action-field, it does work when the input field is referenced (as in the code I cecked in) but in the same way as with icon, the whole field just toggles. When I create the button it has an event, as long as all code is in the init() function, when I save the button to a private variable in the class then the events can't be trapped in the recursiveRender() function. Checking with the inspector in FF there is no event associated  with the button itself.

Fomantic class for buttons supports basic and tertiary, 
basic makes a button with outlines and no bg-color, 
tertiary removes the outline as well.

That's what I found out so far. If there is a way to force the event for just the button then this would work.

I think that the icon version looks nicer though.

rickyosser:
Ok, now I got it working, the question is how to move the button to the inside of the input-field, if that is necessary.

rickyosser:
The demo looks like this now:
image

mvorisek:
The last row of examples looks very promising!

Let's minimize the changes to only this version and then let's do the tests.

tests-behat/revealEye.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Password with revealEye
+
+  Scenario:
+    Given I am on "form-control/input2.php"
+    When I fill in "password_re_norm" with "123"
+    Then "password_re_norm" should contain text "123"
```
mvorisek: I think the text will be always contained, you need to check if the input `type` attribute is changed.

rickyosser: I know, this is just trial and error...
I've got to see how to do a class check for the icon.

src/Form/Control/Password.php
```diff
@@ -4,7 +4,88 @@
...
+    {
+        return array_filter($arr, static fn ($e) => !in_array($e, $remove, true));
+    }
+
+    #[\Override]
+    protected function recursiveRender(): void
+    {
+        if ($this->revealEye) {
+            if ($this->stickyGet('iconInit')) {
+                $iconInit = $this->stickyGet('iconInit');
```
mvorisek: What is the motivation to persist reveal across loads?

rickyosser: When the click event occurs 2 Jqueries are sent, one POST, which is ok, the second is a GET, this resets the class of the field to default.
To fix this I send the class field and the inputType so it is persistant after the GET.

mvorisek: reverted in https://github.com/atk4/ui/pull/2219/commits/d7cf0494bffbd06ad7ff8146604bed100bba9e74

tests-behat/revealEye.feature
mvorisek: done

src/Form/Control/Password.php
mvorisek: IE - I tested Windows 7 and 10, it seems Microsoft does everything to kill IE in favor of Edge - in both systems, when I opened IE, Edge was opened instead. Edge is based on Chrome. And in Edge and Chrome there is no native reveal icon added. So we do not need any additional code to coexist with the native reveal.

mvorisek:
Well done. One more question - you initially introduced `revealEyeIcon` property. I would support it, the question is, can two icons be put inside the right side of `input` element by FUI? Imagine someone using the `icon` for marking required fields etc.

mvorisek:
And one more question - https://jsfiddle.net/fv2bwyq5/ - In the demo the focus is kept when a div is clicked. I do not fully understand why, but the reveal/unreveal click currently removes focus from the password, but the focus should be kept as in the demo when a div is clicked.

rickyosser:
> Well done. One more question - you initially introduced `revealEyeIcon` property. I would support it, the question is, can two icons be put inside the right side of `input` element by FUI? Imagine someone using the `icon` for marking required fields etc.

Thanks, even though you've made it better... :)
Anyway, to answer your question, yes, according to  the code, I haven't tried it, but you should be able to add another icon either with BeforeInput or AfterAfterInput and the eye will always be AfterInput. I'll try and see if it works tomorrow with the old code before your changes tonight.

rickyosser:
> And one more question - https://jsfiddle.net/fv2bwyq5/ - In the demo the focus is kept when a div is clicked. I do not fully understand why, but the reveal/unreveal click currently removes focus from the password, but the focus should be kept as in the demo when a div is clicked.

I checked 2 different applications, 1 an AccessPoint looses focus on the password field when the eye is pressed, Acronis on their on-line management platform doesn't. Anwyay, it is an easy fix, I'll commit it now.

mvorisek:
> > And one more question - https://jsfiddle.net/fv2bwyq5/ - In the demo the focus is kept when a div is clicked. I do not fully understand why, but the reveal/unreveal click currently removes focus from the password, but the focus should be kept as in the demo when a div is clicked.
> 
> I checked 2 different applications, 1 an AccessPoint looses focus on the password field when the eye is pressed, Acronis on their on-line management platform doesn't. Anwyay, it is an easy fix, I'll commit it now.

Based on https://stackoverflow.com/questions/8735764/prevent-firing-focus-event-when-clicking-on-div I meant to use native `mousedown` as `click` even can no longer cancel the focus, but I did some testing and your `elem.focus()` solution seems to have better UX and is more consistent how other elements remove focus. 👍

So only https://github.com/atk4/ui/pull/2219#issuecomment-2341915511 remains and we are ready to go!

rickyosser:
> Well done. One more question - you initially introduced `revealEyeIcon` property. I would support it, the question is, can two icons be put inside the right side of `input` element by FUI? Imagine someone using the `icon` for marking required fields etc.

It was a neat idea, it doesn't work in practice! :)
Let's stay with the icon replacement. If the developer wants to add something else he can turn of revealEye or add a label/add action.

mvorisek:
I love this PR, thank you for this core contribution ❤

PR #2220, opened 2024-08-27
WIP/Extended options - Add slider form control

rickyosser:
This adds the Fomantic-slider as a form-control.
A demo will be added shortly as well as some kind of behat tests.

There are a myriad of extra options that can be added, ie, vertical and colors and stuff.
If there is anything missing feel free to add or ask for it.


demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+use Atk4\Ui\Js\JsExpression;
+use Atk4\Ui\Js\JsFunction;
+use Atk4\Ui\Js\JsReload;
+
+class Slider extends Input
+{
+    public string $inputType = 'hidden';
+
+    /** The lowest value the slider can be. */
+    public int $min = 0;
```
mvorisek: Are we limited to integers by something?

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+            [
+                Hidden::class
+            ]
+        )->set($this->start);
+
+        $onChange = [
+            'onChange' => new JsFunction(
+                ['v'],
+                [
+                    new JsExpression($this->firstInput->js()->find('input')->jsRender().".val($('div#" . $this->slider->getHtmlId() . "').slider('get thumbValue', 'first'))"),
```
mvorisek: use templating for passing ID/elem

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+    public bool $labeled = false;
+
+    /** Whether the ticks and labels should be at the bottom. */
+    public bool $bottom = false;
+
+    /** @var object */
+    private $slider;
+
+    /** @var object */
+    private $owner;
```
mvorisek: owner is stored in the parent API already

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+    /** Whether the slider should be ticked or not. */
+    public bool $ticked = false;
+
+    /** Whether the slider should be labeled or not. */
+    public bool $labeled = false;
+
+    /** Whether the ticks and labels should be at the bottom. */
+    public bool $bottom = false;
+
+    /** @var object */
```
mvorisek:
```suggestion
    /** @var View */
```

at least if general View type is needed

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+declare(strict_types=1);
+
+namespace Atk4\Ui\Form\Control;
+
+use Atk4\Ui\View;
+use Atk4\Ui\Js\JsExpression;
+use Atk4\Ui\Js\JsFunction;
+use Atk4\Ui\Js\JsReload;
+
+class Slider extends Input
```
mvorisek: enabled/disabled/readonly demo should be added into `demos/form-control/input2.php`

and there should be probably a separate demo for Slicer + Behat tests are needed

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+    public bool $smooth = false;
+
+    /** Whether labels should auto adjust on window resize. */
+    public bool $autoAdjustLabels = true;
+
+    /** The distance between labels. */
+    public int $labelDistance = 100;
+
+    /** Number of decimals to use with an unstepped slider. */
+    public int $decimalPlaces = 2;
```
mvorisek: How is this helpful vs. `step`?

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+    public $showLabelTicks = false;
+
+    /** Define smoothness when the slider is moving. */
+    public bool $smooth = false;
+
+    /** Whether labels should auto adjust on window resize. */
+    public bool $autoAdjustLabels = true;
+
+    /** The distance between labels. */
+    public int $labelDistance = 100;
```
mvorisek: When is this helpful? Does this affect CSS only?

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+     * Show ticks on a labeled slider.
+     *'always'will always show the ticks for all labels (even if not shown)
+     * true will display the ticks only if the related label is also shown.
+     *
+     * @var 'always'|bool
+     */
+    public $showLabelTicks = false;
+
+    /** Define smoothness when the slider is moving. */
+    public bool $smooth = false;
```
mvorisek: What is this?

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+
+    /** Prevents the lower thumb to crossover the thumb handle. */
+    public bool $preventCrossover = true;
+
+    /** Settings for the slider-class */
+    /** Whether the slider should be ticked or not. */
+    public bool $ticked = false;
+
+    /** Whether the slider should be labeled or not. */
+    public bool $labeled = false;
```
mvorisek: Isn't this controlled by something in `Input` or `Control` already?

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+        $this->slider->js(true)->slider(
+            $sliderSettings,
+        );
+    }
+
+    #[\Override]
+    protected function recursiveRender(): void
+    {
+        parent::recursiveRender();
+    }
```
mvorisek:
```suggestion
```

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+    private $secondInput;
+    
+    #[\Override]
+    protected function init(): void
+    {
+        parent::init();
+
+        $this->owner = $this->getOwner();
+
+        $this->slider = View::addTo($this->owner)->addClass('ui slider');
```
mvorisek: see https://github.com/atk4/ui/blob/5.2.0/src/View.php#L50

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+        if ($this->labeled) {
+            $this->slider->addClass('labeled');
+        }
+
+        if ($this->bottom) {
+            $this->slider->addClass('bottom aligned');
+        }
+
+        $sliderSettings = [];
+        $sliderSettings = $sliderSettings + ['min' => $this->min];
```
mvorisek:
```suggestion
        $sliderSettings['min'] = $this->min;
```

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+        $sliderSettings = $sliderSettings + ['decimalPlaces' => $this->decimalPlaces];
+        $sliderSettings = $sliderSettings + ['pageMultiplier' => $this->pageMultiplier];
+        $sliderSettings = $sliderSettings + ['decimalPlaces' => $this->decimalPlaces];
+        $sliderSettings = $sliderSettings + ['preventCrossover' => $this->preventCrossover];
+        
+        /*
+         * First input value, always present
+         */
+        $this->firstInput = $this->owner->addControl(
+            $this->shortName . '_first',
```
mvorisek, line 169:
```suggestion
            $this->shortName . '__atk4__slider_first__',
```

demos/form-control/Slider.php
```diff
@@ -0,0 +1,215 @@
...
+    public string $inputType = 'hidden';
+
+    /** The lowest value the slider can be. */
+    public int $min = 0;
+
+    /** The max value the slider can be. */
+    public int $max = 10;
+
+    /** @var float|null The slider step. Set to 0 to disable step. */
+    public $step = 0;
```
mvorisek: Just a question for now - how feasible is histogram below the slider and non-liner steps or even non-numeric values (XS-S-M-L...)?

rickyosser: Yes, it's possible!
According to Fomantic [https://fomantic-ui.com/modules/slider.html#custom-interpreted-labels](https://fomantic-ui.com/modules/slider.html#custom-interpreted-labels) it can be done with:

```
var labels = ["α", "β", "γ", "δ", "ε", "ζ", "η", "θ", "ι", "κ", "λ", "μ", "ν", "ξ", "ο", "π", "ρ", "σ/ς", "τ", "υ", "φ", "χ", "ψ", "ω"];
$('#interpretedlabel')
  .slider({
    interpretLabel: function(value) {
      return labels[value];
    }
  })
;
```
As a matter of fact, all the options I've implemented, except this one is directly from Fomantic. I just broke out ticked, bottom align and labeled from the class and made those options. If some of them are funky, yes, I know, they can be removed if we deem the don't add anything. I just added it for completness for now.

demos/form-control/Slider.php
rickyosser: No, I've changed this to float, but then the result is also returned in float.

demos/form-control/Slider.php
rickyosser: Done... :) I just forgot.

demos/form-control/Slider.php
rickyosser: I can't find this, do you have an example? I just need to know how to find the parent object.

demos/form-control/Slider.php
rickyosser: I'll get to this in time...

demos/form-control/Slider.php
rickyosser: decimalPlaces 	2 	Number of decimals to use with an unstepped slider
That is from the Fomantic doc, I don't know, just always use step! :)
We can always remove the option in the future.

demos/form-control/Slider.php
rickyosser: Yes, but it also defines what labels will be shown, depending on space available.

demos/form-control/Slider.php
rickyosser: It's for the movement of the thumb, if it's in steps between ticks or smooth scrolling.

demos/form-control/Slider.php
rickyosser: No, you are thinking of caption, that's the info "before" the controller, labeled is if the slider should have text at the ticks or not.

demos/form-control/Slider.php
rickyosser: I'll remove it later, now I'm using it to check variables and classes after render.

demos/form-control/Slider.php
rickyosser: Fixed... Thanks for the idea!

demos/form-control/Slider.php
mvorisek: We can easily allow `int|float` to accept integers over 2^53, the deeper question is if the values are correctly typecasted for attributes (machine data) and UI (for user). To verify this, the easiest test is to use Slider together with field with this https://github.com/atk4/ui/blob/5.2.0/demos/init-db.php#L55 DBAL type - you can use any demo model which use this type for all IDs.

demos/form-control/Slider.php
mvorisek: https://github.com/atk4/ui/blob/5.2.0/src/AbstractView.php#L19

demos/form-control/Slider.php
mvorisek: must be `View` not `object|View`, not only because we do not want to support general `object`, but also PHPStan will simplify `object|View` into `object` only

demos/form-control/Slider.php
rickyosser: @mvorisek, I think I need your help on this one. It seems that the function (interpretLabel) need to be existing before  semantic.js (Fomantic-UI) is loaded. I've hacked a version of semantic.js to try this and it indeed added the custom labels then. This is not acceptable though as I know very well, it was just done for test.
I do need help though to add the function first on the page before loading Fomantic-UI. Maybe there is another way but I've already spent to much time on this without any workable solution.

mvorisek: > It seems that the function (interpretLabel) need to be existing before semantic.js (Fomantic-UI) is loaded.

That seems weird to me. All user JS is evaluated "after page is fully loaded", so the elements are guaranteed to exists. And you can define a JS function before you setup the FUI slider by JS.

I do not understand the issue you are facing here.. If you need to define global JS code - to deduplicate long JS code - you can put the code in `/js/*.js`. If you need a small JS code per control, you should always use anonymous JS function.

rickyosser: @mvorisek I fixed it... Using JsFunction and JsExpression was the way to go but I had trouble with the array but using slice instead of addressing the item directly solved it.

demos/form-control/input2.php
```diff
@@ -53,6 +53,61 @@
 $group->addControl('radio_read', [Form\Control\Radio::class, 'readOnly' => true], ['enum' => ['one', 'two', 'three']])->set('two');
 $group->addControl('radio_disb', [Form\Control\Radio::class, 'disabled' => true], ['enum' => ['one', 'two', 'three']])->set('two');
 
+$group = $form->addGroup('Sliders');
```
mvorisek, line 61: I checkouted this PR and my feelings are the current state is too wild:
- because the UX is very different to other controls - for example the vertical space is too large and the control is too wide,
- there are too many options - for example labels above the horizontal axis/line, the atk4/ui framework aims to provide unified and robust developer experience, not 1:1 FUI port - think this way, in the future (in a decade or so), the FUI might became obsolete and we should be able to replace FUI and still support all supported options

So for now, please put this complex PR on hold and open another PR with minimal Slider impl. Just Slider class with normal/RO/disabled support and min/max value. This should be presented in `demos/form-control/input2.php` and min/max options in `demos/form-control/slider.php` and tested with Behat tests.

mkrecek234: Hi everyone, just as food for thought (courtesy of @abbadon1334) a minimal implementation of Fomantic-UI slider sample code:
```
                $slider = $this->add([View::class] )->addClass('ui labeled ticked small slider');

                $slider->js(true)->slider([
                    "min" => 0,
                    "max" => 45,
                    'showThumbTooltip' => 'true',
                    'tooltipConfig' => ['position' => 'top center',
                        'variation' => 'small visible blue'],
                    "onChange" => new \Atk4\Ui\Js\JsFunction(['v'], [
                        new \Atk4\Ui\Js\JsExpression($field->js()->find('input')->jsRender().".val(v)")]),
                ]);
```

mvorisek: > So for now, please put this complex PR on hold and open another PR with minimal Slider impl. Just Slider class with normal/RO/disabled support and min/max value. This should be presented in `demos/form-control/input2.php` and min/max options in `demos/form-control/slider.php` and tested with Behat tests.

@rickyosser would you like to work on the first minimal implementation?

rickyosser:
Hi,

yes, though I've been busy the last months with other work. I hope to
have some more time around new years, I've done some changes already
which aren't pushed yet, I'll remove the more complex options
next, except those I need for my project.

Thanks for asking,

Rickard

demos/form-control/input2.php
rickyosser: Ok, I've resumed work on this control again, I will do a less complex version but, ranged has to be in there as that is used in my application to be able to select an interval.

There are a few other options which might be interesting to add as well but that can be done after doing the simple-slider.
I will continue to maintain the more complex version as I know a few cases where this might be interesting to use. Especially with vertical sliders for building day-based rules and stuff.

rickyosser: Ok, I'm going through the options and I have some questions on what you @mvorisek want as the default (only) behaviour?

Take these examples:
Lables (Only below or above?)
Ticks or no ticks?
Custom labels (If no there is no way of setting the labels excepft for numbers).
Should the control handle setting up the classes as some need to have specific order to work?

I'll start with removing vertical slider as I'm not using it at the moment.
The reason for asking is that Slider in itself is a very potent tool to simplify and visualize an input where we have pre-defined inputs or a need to get a range within pre-defined limits. Other controls don't have that impact at all and using ie regular input controls would add a plethora of complexities to make those limits or pre-defined selections legal. Well, a simple slider replaces a dropdown in certain cases but with a more visual approach. A ranged slider could be replaced with 2 dropdowns but would not give the user a visual representation of the limits of the form. Programming those dropdowns would also be of a more complex nature than having a complex slider...

Just a couple of thoughts... :)

PR #2225, opened 2024-11-20
Add loadable typecast for JsCallback arguments

mvorisek:
Newly the passed arguments for JsCallback must define also "load typecasting".

This way we ensure the values passed to the given function passed to JsCallback are always already typecasted and with the right typecasting method for the generating source.

If values are passed as not-`JsExpressionable` thru `View::on`, they are bidirectionally typecasted implicitly.


DarkSide666 (commented):
looks good to me

PR #2237, opened 2024-12-17
Fix menu dropdown for any model

### LINKED ISSUE #2236: Ui\Dropdown with typecasting bug
Steps to reproduce:
- Add a Ui\Dropdown view and setModel for any model with a usual bigint idField
- Configure Dropdown->onChange to handle selected item
- Select an entry of the dropdown

Sample code:
```
    $agentdropdown = \Atk4\Ui\Dropdown::addTo($app, ['Select agent', 'dropdownOptions' => ['on' => 'hover']]);
    $agentdropdown->setModel($agentmodel);

    $agentdropdown->onChange(static function ($itemId) {
        return "do something with ".$itemId;
    });
```

Typecasting error `TypeError: Atk4\Ui\Dropdown::Atk4\Ui\{closure}(): Argument #2 ($value) must be of type string, int given, called in vendor/atk4/ui/src/JsCallback.php on line 90`

Solution: 
https://github.com/atk4/ui/blob/7383f1ee33078a1d3bf3e806005037134b3afb8a/src/Dropdown.php#L57

The itemId here is forced to be string, however the idField is typecasted as int. This has to be adapted.

mkrecek234:
fix #2236


demos/basic/menu.php
```diff
@@ -25,6 +25,13 @@
     return new JsToast('New selected item ID: ' . $itemId);
 });
 
+$model = new Category($app->db);
+$dropdown2 = UiDropdown::addTo($menu, ['Model with Callback', 'dropdownOptions' => ['on' => 'hover']]);
+$dropdown2->setModel($model);
+$dropdown2->onChange(static function ($itemId) {
+    return new JsToast('New selected item ID: ' . $itemId);
```
mvorisek, line 32: Isn't the ID WrappedId instance?

src/Dropdown.php
```diff
@@ -54,11 +54,11 @@ public function onChange(\Closure $fx): void
             ),
         ]);
 
-        $this->cb->set(static function (Jquery $j, string $value) use ($fx) {
+        $this->cb->set(static function (Jquery $j, $value) use ($fx) {
             return $fx($value);
         }, ['item' => new JsCallbackLoadableValue(null, function ($v) {
             return $this->getApp()->uiPersistence->typecastLoadField(
-                $this->model->getField('id'),
+                $this->model->getIdField(),
```
mvorisek, line 61: `'id'` is hardcoded in https://github.com/atk4/ui/blob/5.2.0/template/dropdown.html#L6 thus `'id'` must be hardcoded here as well - probably related with https://github.com/atk4/ui/pull/2237/files#r1889379601 - I belive this PR cannot work.

mvorisek:
I have tested this PR and so far the result is:

image

So there are these issues:
- dropdown initial text is not shown
- values are not shown - https://github.com/atk4/ui/pull/2237#discussion_r1889379601
- https://github.com/atk4/ui/pull/2237#discussion_r1899424720

PR #2239, opened 2025-01-03
Rename `Form::setModel()` method to `Form::setEntity()`

mvorisek:
continue #2192

mainly for `Form` class, but also for `Card`, `CardSection`, `CardTable`, `Form\AbstractLayout` and `InlineEdit` classes

and remove `$model`/`$entity` properties from `View`


DarkSide666 (approved):
LGTM

PR #2241, opened 2025-01-11
Strip "ID" suffix from reference form control label

### LINKED ISSUE #1759: Enhancement for hasOne dropdown captions
Steps to reproduce:
- You have a hasOne "Timezone_id" relation (which contains id and name) and a form which adds a Control for that with `$form->addControl('timezone_id')`;
- The form will include the Dropdown and caption will be "Timezone Id".

Expected behaviour:
- For hasOne fields where a Dropdown is added by default, the caption should remove the "_id" component and read "Timezone" only

image

mvorisek:
close #1759

related with https://github.com/atk4/data/pull/1239 - this improves https://github.com/atk4/ui/blob/5.2.0/src/Table/Column.php#L361 and https://github.com/atk4/ui/blob/5.2.0/src/CardTable.php#L44

### DIFF
--- a/src/Form/Layout.php
+++ b/src/Form/Layout.php
@@ -173,7 +173,7 @@ class Layout extends AbstractLayout
             if ($label === null) {
                 $label = $element->entityField->getField()->getCaption();
                 if (property_exists($element, 'model')) {
-                    $label = preg_replace('~^(?:.*\K ID$|ID )~s', '', $label);
+                    $label = preg_replace('~ ID$~i', '', $label);
                 }
             }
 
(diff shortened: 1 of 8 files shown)


src/Form/Layout.php
```diff
@@ -169,7 +169,13 @@ protected function recursiveRender(): void
             }
 
             $template = $element->renderLabel ? $labeledControl : $noLabelControl;
-            $label = $element->caption ?? $element->entityField->getField()->getCaption();
+            $label = $element->caption;
+            if ($label === null) {
+                $label = $element->entityField->getField()->getCaption();
+                if (property_exists($element, 'model')) {
+                    $label = preg_replace('~ ID$~i', '', $label);
```
mvorisek, line 176: TODO replace also "their reference field name" like in https://github.com/atk4/data/pull/1239/commits/d639bc958e5bf2d2f6827a1acbf7527101d565ee#diff-1c01b02e4e8021fe40414d7afd3f719df174d58c7076e2d436e8d994eef8f603R71

PR #2244, opened 2025-01-29
Terminate async JS handlers when DOM owner is destroyed

### LINKED ISSUE #393: Reload must unload/stop old JS
Due to nature of SSE when console is refreshed and previous process is still running, it will interfere with the new process.

console-noise

This does not happen cross-console, so I think this can be solved if we generate unique identifier when rendering a console and filter by it.

mvorisek:
fix #393

All DOM element related async callbacks must be terminated/removed before the element is removed/replaced.


js/src/Service/apiService.js
```diff
@@ -36,6 +37,11 @@ class ApiService {
         }).call(thisObject);
     }
 
+    onRequest(promise, xhr) {
+        // fix https://github.com/atk4/ui/issues/393
+        atk.elementRemoveObserver.addHandler(this, () => xhr.abort());
```
mvorisek, line 42: handler should be removed once the request is finish

mvorisek:
Hi @ibelar, might I have your review on this JS topic?

PR #2287, opened 2025-05-09
Add missing JsCallbackLoadableValue to demo callback

### LINKED ISSUE #2286: "form/form2.php" demo addAction is broken
How to reproduce:
1) open demo /ui/demos/form/form2.php
2) enter blabla in country field (first)
3) click "Check Duplicate"

Result:
```
Critical Error TypeError:
Atk4\Ui\Demos\{closure}(): Argument #2 ($name) must be of type string, Atk4\Ui\Js\Jquery given,
called in src/View.php on line 924 
```
Expected result: "This country name can be added."
Source: https://github.com/atk4/ui/blob/develop/demos/form/form2.php

No idea how to fix it, since button click interface is undocumented and differs from onSubmit interface.

mvorisek:
fix #2286

### DIFF
--- a/demos/form/form2.php
+++ b/demos/form/form2.php
@@ -11,6 +11,7 @@ use Atk4\Ui\Form;
 use Atk4\Ui\Header;
 use Atk4\Ui\Js\Jquery;
 use Atk4\Ui\Js\JsBlock;
+use Atk4\Ui\Js\JsCallbackLoadableValue;
 use Atk4\Ui\Js\JsToast;
 use Atk4\Ui\Label;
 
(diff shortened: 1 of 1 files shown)


demos/form/form2.php
```diff
@@ -39,7 +40,7 @@
         }
 
         return new JsToast('This country name can be added.');
-    }, ['args' => [$nameInput->jsInput()->val()]]);
+    }, ['args' => [new JsCallbackLoadableValue($nameInput->jsInput()->val(), static fn ($v) => $v)]]);
```
mvorisek, line 43: TODO for myself - can this be detected using phpstan or no lately on runtime?

woodholly: It works for me, but looks like black magic, not low code :)

mvorisek: Low code, but still we require explicit code to enforce strong typecasting. Here we indicate we want to keep string type.

With improve method param type PHPStan should be able to detect the issue.

Also we should throw if the type is not `JsCallbackLoadableValue` in `JsCallback`. The reason you got "argument error" is passed `JsExpression` implements `__call`. I will work on this.

PR #2288, opened 2025-05-09
Allow disabling dropdown clearable

rickyosser:
I added a small if to check if the "clearable" option is set, if it is then override the default model-based behaviour for dropdowns.

The current implementation requires you to actually use a model to populate the dropdown otherwise clearable is always set. In cases where you aren't using any models at all for DB/data related handling this will make it possible to delete the value which might not be the expected result. 

This PR will make the the Dropdown honour setDropdownOption('clearable', false).

### DIFF
--- a/src/Form/Control/Dropdown.php
+++ b/src/Form/Control/Dropdown.php
@@ -159,10 +159,12 @@ class Dropdown extends Input
     protected function jsRenderDropdown(): JsExpressionable
     {
         $dropdownOptions = $this->dropdownOptions;
-        if ($this->entityField === null || ($this->entityField->getField()->nullable || !$this->entityField->getField()->required)) {
-            $dropdownOptions['clearable'] = true;
-        }
 
+        if (!array_key_exists('clearable', $dropdownOptions)) {
+            if ($this->entityField === null || ($this->entityField->getField()->nullable || !$this->entityField->getField()->required)) {
+                $dropdownOptions['clearable'] = true;
+            }
+        }
         return $this->jsDropdown(true)->dropdown($dropdownOptions);
     }
 
--- a/src/Form/Control/Dropdown.php
+++ b/src/Form/Control/Dropdown.php
@@ -165,6 +165,7 @@ class Dropdown extends Input
                 $dropdownOptions['clearable'] = true;
             }
         }
+
         return $this->jsDropdown(true)->dropdown($dropdownOptions);
     }
 
--- a/src/Form/Control/Dropdown.php
+++ b/src/Form/Control/Dropdown.php
@@ -160,7 +160,7 @@ class Dropdown extends Input
     {
         $dropdownOptions = $this->dropdownOptions;
 
-        if (!isset($dropdownOptions['clearable']) && ($this->entityField === null || ($this->entityField->getField()->nullable || !$this->entityField->getField()->required))) {
+        if (!isset($dropdownOptions['clearable']) && ($this->entityField === null || $this->entityField->getField()->nullable || !$this->entityField->getField()->required)) {
             $dropdownOptions['clearable'] = true;
         }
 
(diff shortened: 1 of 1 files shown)


src/Form/Control/Dropdown.php
```diff
@@ -159,8 +159,11 @@ protected function jsDropdown($when = false, $action = null): JsExpressionable
     protected function jsRenderDropdown(): JsExpressionable
     {
         $dropdownOptions = $this->dropdownOptions;
-        if ($this->entityField === null || ($this->entityField->getField()->nullable || !$this->entityField->getField()->required)) {
-            $dropdownOptions['clearable'] = true;
+
+        if (!array_key_exists('clearable', $dropdownOptions)) {
+            if ($this->entityField === null || ($this->entityField->getField()->nullable || !$this->entityField->getField()->required)) {
```
mvorisek, line 164: `!isset() && (orig cond)` CS please, otherwise LGTM

src/Form/Control/Dropdown.php
```diff
@@ -159,8 +159,11 @@ protected function jsDropdown($when = false, $action = null): JsExpressionable
     protected function jsRenderDropdown(): JsExpressionable
```
mvorisek, line 159: and what about https://github.com/atk4/ui/pull/2063/commits/6fcc1f225c0b2f0095d8d4106b1ed92ba6b37a2e?

mvorisek: Also please clarify the motivation of this PR.

rickyosser: > and what about [6fcc1f2](https://github.com/atk4/ui/commit/6fcc1f225c0b2f0095d8d4106b1ed92ba6b37a2e)?

According to the Human Inteface Guidlines (in most versions) a radio-button group can not be deselected as well as only one option should be selectable. If you want an unselected radio-button group, you should really use checkboxes instead.

Check https://uxplanet.org/radio-buttons-ux-design-588e5c0a50dc for example.

rickyosser: > Also please clarify the motivation of this PR.

I added it to the initial comment.

src/Form/Control/Dropdown.php
```diff
@@ -159,7 +159,8 @@ protected function jsDropdown($when = false, $action = null): JsExpressionable
...
     {
         $dropdownOptions = $this->dropdownOptions;
-        if ($this->entityField === null || ($this->entityField->getField()->nullable || !$this->entityField->getField()->required)) {
+
+        if (!isset($dropdownOptions['clearable']) && $this->entityField === null || ($this->entityField->getField()->nullable || !$this->entityField->getField()->required)) {
```
mvorisek, line 163:
```suggestion
        if (!isset($dropdownOptions['clearable']) && ($this->entityField === null || ($this->entityField->getField()->nullable || !$this->entityField->getField()->required))) {
```

src/Form/Control/Dropdown.php
```diff
@@ -159,7 +159,8 @@ protected function jsDropdown($when = false, $action = null): JsExpressionable
     protected function jsRenderDropdown(): JsExpressionable
     {
         $dropdownOptions = $this->dropdownOptions;
-        if ($this->entityField === null || ($this->entityField->getField()->nullable || !$this->entityField->getField()->required)) {
+
+        if (!isset($dropdownOptions['clearable']) && ($this->entityField === null || ($this->entityField->getField()->nullable || !$this->entityField->getField()->required))) {
```
mvorisek, line 163:
```suggestion
        if (!isset($dropdownOptions['clearable']) && ($this->entityField === null || $this->entityField->getField()->nullable || !$this->entityField->getField()->required)) {
```

src/Form/Control/Dropdown.php
mvorisek: I am ok to allow to disable the "clearable" in opt-in fashion.

PR #2290, opened 2025-05-11
Improve comment about View\ModelTrait::$model DI in View::init()

### LINKED ISSUE #2284: View::setModel() called in View::init() - still needed in V6?
Hi,

in atk4\ui 6 the property `View::$model` was removed, and so was `View::setModel()`.

However,  part of this former logic still resides within `View::init()`:
```php

        // allow for injecting the model with a seed
        if (property_exists($this, 'model') && $this->model !== null) {
            $this->setModel($this->model); // @phpstan-ignore method.notFound
        }
```

IMHO this code is not needed in `View` any more. If any classes which extend `View` add a `$model` property, they know themselves how to handle this property and should not rely on `View` parent to call `setModel()`.

PhilippGrashoff:
fix #2284

This PR removes some logic inside `View::init()` that was maybe kept for backwards compatibility, as `View` used to have a `model` property until v6.
As `View` now does not have a `model` property any more, it also shouldn't include the logic to call `setModel()` method if a model property exists. This should be handled by classes extending `View` which add a `model` property.

### DIFF
--- a/src/View.php
+++ b/src/View.php
@@ -183,11 +183,6 @@ class View extends AbstractView
         foreach ($addLater as [$object, $region]) {
             $this->add($object, $region);
         }
-
-        // allow for injecting the model with a seed
-        if (property_exists($this, 'model') && $this->model !== null) {
-            $this->setModel($this->model); // @phpstan-ignore method.notFound
-        }
     }
 
     public function getExecutorFactory(): ExecutorFactory
--- a/src/View.php
+++ b/src/View.php
@@ -183,6 +183,11 @@ class View extends AbstractView
         foreach ($addLater as [$object, $region]) {
             $this->add($object, $region);
         }
+
+        // allow for injecting the model with a seed
+        if (property_exists($this, 'model') && $this->model !== null) {
+            $this->setModel($this->model); // @phpstan-ignore method.notFound
+        }
     }
 
     public function getExecutorFactory(): ExecutorFactory
--- a/src/View.php
+++ b/src/View.php
@@ -184,7 +184,7 @@ class View extends AbstractView
             $this->add($object, $region);
         }
 
-        // allow for injecting the model with a seed
+        // allow for injecting the model using View\ModelTrait
         if (property_exists($this, 'model') && $this->model !== null) {
             $this->setModel($this->model); // @phpstan-ignore method.notFound
         }
--- a/src/View/EntityTrait.php
+++ b/src/View/EntityTrait.php
@@ -11,6 +11,16 @@ trait EntityTrait
 {
     public ?Model $entity = null;
 
+    #[\Override]
+    protected function init(): void
+    {
+        parent::init();
+
+        if ($this->entity !== null) {
+            $this->setEntity($this->entity);
+        }
+    }
+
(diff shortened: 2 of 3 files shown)


mvorisek:
please fix CI

mvorisek:
Currently one demo relies on this: demos/_unit-test/scope-builder.php. Easy to fix.

However, what is the motivation of this change? Clean OOP?

I am very worried about migration path. I am fine with BC-break, but the main BC-breaks should be at least easily detectable using PHPStan l6 or higher. This is not the case here.

We can move this into `View\ModelTrait`, but then we would need to handle overriden init() conflicts.

mvorisek:
If DI with `model` property is minimal, we can override `setDefaults` (to not require having to resolve conflict of `init`) and throw there if the `model` (and `entity`) property was set not thru `setModel`/`setEntity`. This way, PHPStan will still not detect any issue, but runtime will.

Wdyt?

PhilippGrashoff:
Hi,

motivation for this PR - clearly clean code. However, I looked at the code in `View::init()` isolated from `ModelTrait`, which it clearly "partners" with.

- The cleanest solution would be to have a `ViewWithModel` class extending View having both the init() logic as well as ModelTrait content. But as many classes using ModelTrait already inherit `ViewWithContent`, this unfortunately is not an option.
- If the code I removed checked if ModelTrait was used, i'd be ok with it, Then, it would a) be clear that this code is meant to partner with ModelTrait and b) would only execute if ModelTrait was actually used. In the current implementation, you can extend View, add a $model property and then be surprised about the exception that setModel() is not defined :)
- The alternative would be to force users to use ModelTrait in case a $model property is added to a child class of View.
- regarding overwriting setDefaults: Can you explain what you mean? I do not 100% get it yet.

mvorisek:
> If the code I removed checked if ModelTrait was used, i'd be ok with it, Then, it would a) be clear that this code is meant to partner with ModelTrait and b) would only execute if ModelTrait was actually used. In the current implementation, you can extend View, add a $model property and then be surprised about the exception that setModel() is not defined :)

Here you landed several good points and I think the only solution is to move the code into `ModelTrait` directly to address all of them easily.

As long as the code can be put directly after parent init, it can be coded quite easily, let me to check this later.

mvorisek:
I really do not know how to fix this.

https://github.com/atk4/ui/pull/2290/commits/08bfb917082c2a2c49ac73496c90a35c8691c008 is not good solution as then traits with multiple init could not be combined.

And `TraitUtil` from core is not advised to be used in general - https://github.com/atk4/core/blob/6.0.0/src/TraitUtil.php#L31.

So maybe the updated comment seems to be fine fo "clearly clean code" motivation.

Situation when user adds `model` property is not rare, but he should use the `View\ModelTrait`, so in reality, there is no issue.

mvorisek:
> [08bfb91](https://github.com/atk4/ui/commit/08bfb917082c2a2c49ac73496c90a35c8691c008) is not good solution as then traits with multiple init could not be combined.

We would need https://github.com/phpstan/phpstan/discussions/13010 first.

PR #2293, opened 2025-05-12
Fix Form\Layout::recursiveRender() to respect template regions

### LINKED ISSUE #2283: FormLayout does not respect Regions of added Views
Hi,

see this simple example:
```php
$form = Form::addTo($app);
$someView = View::addTo($form->layout, ['SomeContent'], ['SomeRegion']);
$someButtom = Button::addTo($form->layout,['SomeButtonTitle'],['SomeOtherRegion']);
```
with a custom formlayout template like
```html
<div class="eooCbcLeft">
    {$name}
    {$details}
    {$Content}
     {$Buttons}
</div>

<div class="eooCbcRight">
    {$SomeRegion}
    {$SomeOtherRegion}
</div>
```

`Form\Layout` will not respect the regions `SomeRegion` and `SomeOtherRegion`, but add `$someView` to `{$Content} and `$someButton` to `{$Buttons}`

I made some simple changes to Form\Layout which solves both problems, but its an ugly, and IMO a proper solution should refactor `Form\Layout::recursiveRender()`

```diff

    #[\Override]
    protected function recursiveRender(): void
    {
        $labeledControl = $this->inputTemplate->cloneRegion('LabeledControl');
        $noLabelControl = $this->inputTemplate->cloneRegion('NoLabelControl');
        $labeledGroup = $this->inputTemplate->cloneRegion('LabeledGroup');
        $noLabelGroup = $this->inputTemplate->cloneRegion('NoLabelGroup');

        $this->template->del('Content');

        foreach ($this->elements as $element) {
            // buttons go under Button section
            if ($element instanceof Button) {
+                if ($element->region && $this->template->hasTag($element->region)) {
+                   $this->template->dangerouslyAppendHtml(
+                      $element->region,
+                     $element->getHtml()
+                );
+                } else {
                    $this->template->dangerouslyAppendHtml('Buttons', $element->getHtml());
+                }
                continue;
            }

            if ($element instanceof self) {
                if ($element->label && !$element->inline) {
                    $template = $labeledGroup;
                    $template->set('label', $element->label);
                } else {
                    $template = $noLabelGroup;
                }

                if ($element->width) {
                    $template->set('width', $element->width);
                }

                if ($element->inline) {
                    $template->set('class', 'inline');
                }
                $template->dangerouslySetHtml('Content', $element->getHtml());
+                if ($element->region && $this->template->hasTag($element->region)) {
+                    $this->template->dangerouslyAppendHtml(
+                        $element->region,
+                        $template->renderToHtml()
+                    );
+                } else {
                    $this->template->dangerouslyAppendHtml('Content', $template->renderToHtml());
+                }
                continue;
            }

            // anything but controls or explicitly defined controls get inserted directly
            if (!$element instanceof Control || !$element->layoutWrap) {
+                if ($element->region && $this->template->hasTag($element->region)) {
+                    $this->template->dangerouslyAppendHtml(
+                        $element->region,
+                        $element->getHtml()
+                    ); // @phpstan-ignore method.notFound

+                } else {
                    $this->template->dangerouslyAppendHtml(
                        'Content',
                        $element->getHtml()
                    ); // @phpstan-ignore method.notFound
+                }
                continue;
            }

            $template = $element->renderLabel ? $labeledControl : $noLabelControl;
            $label = $element->caption ?? $element->entityField->getField()->getCaption();

            // anything but form controls gets inserted directly
            if ($element instanceof Control\Checkbox) {
                $template = $noLabelControl;
                $element->template->set('Content', $label);
            }

            if ($this->label && $this->inline) {
                if ($element instanceof Control\Input) {
                    $element->placeholder = $label;
                }
                $label = $this->label;
                $this->label = null;
            } elseif ($this->label || $this->inline) {
                $template = $noLabelControl;
                if ($element instanceof Control\Input) {
                    $element->placeholder = $label;
                }
            }

            // controls get extra pampering
            $template->dangerouslySetHtml('Input', $element->getHtml());
            $template->trySet('label', $label);
            $template->trySet('labelFor', $element->name . '_input');
            $template->set('controlClass', $element->controlClass);

            if ($element->entityField->getField()->required) {
                //$template->append('controlClass', 'required ');
            }

            if ($element->width) {
                $template->append('controlClass', $element->width . ' wide ');
            }

            if ($element->hint && $template->hasTag('Hint')) {
                $hint = Factory::factory($this->defaultHintSeed);
                $hint->name = $element->name . '_hint';
                if (is_object($element->hint) || is_array($element->hint)) {
                    $hint->add($element->hint);
                } else {
                    $hint->set($element->hint);
                }
                $hint->setApp($this->getApp());
                $template->dangerouslySetHtml('Hint', $hint->getHtml());
            } elseif ($template->hasTag('Hint')) {
                $template->del('Hint');
            }

            if ($this->template->hasTag($element->shortName)) {
                $this->template->dangerouslySetHtml($element->shortName, $template->renderToHtml());
            } else {
                $this->template->dangerouslyAppendHtml('Content', $template->renderToHtml());
            }
        }

        // collect JS from everywhere
        foreach ($this->elements as $view) {
            foreach ($view->_jsActions as $when => $actions) { // @phpstan-ignore property.notFound
                foreach ($actions as $action) {
                    $this->_jsActions[$when][] = $action;
                }
            }
        }
    }



```

PhilippGrashoff:
Fix https://github.com/atk4/ui/issues/2283 - respect region definitions of elements instead of always putting them into 'Buttons' or 'Content' sections.

Also make code of `recursiveRender()` more readable and adaptable by splitting its logic into several methods.


PhilippGrashoff:
I also do not see any sensible solution for this other than the clearer comment. I also checked `class_uses()`, but it only returns traits used by the current class, not by childen. And having to loop through all the parents just to check if ModelTrait is used seems an overkill, especially in a method executed as often as View::init().
Lets just update the comment and link this PR so others can read this discussion.

src/Form/Layout.php
```diff
@@ -124,120 +128,144 @@ public function addSubLayout($seed = [self::class], $addDivider = true)
...
+    {
+        if ($element->region && $this->template->hasTag($element->region)) {
+            $this->template->dangerouslyAppendHtml(
+                $element->region,
+                $element->getHtml()
+            );
+        } else {
+            $this->template->dangerouslyAppendHtml($defaultRegion, $element->getHtml());
         }
     }
```
mvorisek:
```suggestion
    private function renderElementToRegion(View $element, string $defaultRegion): void
    {
        $this->template->dangerouslyAppendHtml(
            $element->region && $this->template->hasTag($element->region)
                ? $element->region
                : $defaultRegion,
            $element->getHtml()
        );
    }
```

src/Form/Layout.php
```diff
@@ -124,120 +128,144 @@ public function addSubLayout($seed = [self::class], $addDivider = true)
     #[\Override]
     protected function recursiveRender(): void
     {
-        $labeledControl = $this->inputTemplate->cloneRegion('LabeledControl');
-        $noLabelControl = $this->inputTemplate->cloneRegion('NoLabelControl');
-        $labeledGroup = $this->inputTemplate->cloneRegion('LabeledGroup');
-        $noLabelGroup = $this->inputTemplate->cloneRegion('NoLabelGroup');
-
-        $this->template->del('Content');
+        $this->prepareTemplates();
```
mvorisek: I prefer to get this reverted. The templates are defined by the main template, no need to create properties for them.

The reason is keep the code as simple as possible, I do not see any value in this. If you do, please share the motivation and usecase.

src/Form/Layout.php
PhilippGrashoff: Nicer with ternary

src/Form/Layout.php
PhilippGrashoff: The only motivation to have the templates as properties is to have them available in the sub-methods without passing them as parameters. However, when writing this code, I also didn't see any other advantage in having these ttemplates as properties. Will pass them as parameters, unless you see a nicer way.

tests/Form/LayoutTest.php
```diff
@@ -0,0 +1,64 @@
...
+    public function testRecursiveRenderRespectsRegions(): void
+    {
+        $template = new HtmlTemplate(
+            '<div id="defaultRegion">{$Content}{$Buttons}</div><div id="anotherRegion">{$input2}{$AnotherRegion}</div>'
+        );
+        $form = new Form(
+            [
+                'buttonSave' => false,
+                'layout' => [
+                    Form\Layout::class,
```
mvorisek: is the seed class needed here?

tests/Form/LayoutTest.php
```diff
@@ -0,0 +1,64 @@
...
+        ViewWithContent::addTo($form->layout, ['View1']);
+        Form\Layout::addTo($form->layout, ['label' => 'Layout1']);
+
+        // all of these should be rendered inside #anotherRegion
+        Button::addTo($form->layout, ['Button2'], ['AnotherRegion']);
+        ViewWithContent::addTo($form->layout, ['View2'], ['AnotherRegion']);
+        Form\Layout::addTo($form->layout, ['label' => 'Layout2'], ['AnotherRegion']);
+
+        $renderedHtml = $form->getHtml();
+        $firstPart = substr($renderedHtml, 0, strpos($renderedHtml, 'id="anotherRegion"'));
```
mvorisek: This is too fragile and hard to read. Why not to use something like contains string and assert the whole inner html 1:1?

src/Form/Layout.php
```diff
@@ -129,115 +129,132 @@ protected function recursiveRender(): void
         $labeledGroup = $this->inputTemplate->cloneRegion('LabeledGroup');
         $noLabelGroup = $this->inputTemplate->cloneRegion('NoLabelGroup');
 
-        $this->template->del('Content');
+        $this->template->del('Content'); // TODO Content is empty in default template anyway. Needed?
```
mvorisek: There are mayn changes in this PR.

Please update the commits to represent clear history of the changes. When moving a larger chunk of code somewhere, do it in a separate commit, so the real changes can be reviewed easily.

src/Form/Layout.php
```diff
@@ -129,115 +129,132 @@ protected function recursiveRender(): void
...
                 }
+            }
+        }
+    }
 
-                if ($element->inline) {
-                    $template->set('class', 'inline');
-                }
-                $template->dangerouslySetHtml('Content', $element->getHtml());
+    protected function renderLayout(self $layout, HtmlTemplate $labeledGroup, HtmlTemplate $noLabelGroup): void
```
mvorisek:
```suggestion
    protected function recursiveRenderLayout(self $layout, HtmlTemplate $labeledGroup, HtmlTemplate $noLabelGroup): void
```

tests/Form/LayoutTest.php
PhilippGrashoff: No, changed it.

tests/Form/LayoutTest.php
PhilippGrashoff: asserting the whole HTML now

src/Form/Layout.php
PhilippGrashoff: Hi, I do not 100% get what you expect - lets have a quick ckat about this?

PhilippGrashoff:
after asserting the whole rendered HTML, CS Fixer complains about 2 lines which only have 2 spaces. Is there a simple way to stop CS Fixer complaining about this? The 2 spaces exist in the rendered HTML which is asserted, so I cannot simply remove them to satisfy CS fixer.

src/Form/Layout.php
mvorisek: I need to review moved code for corectness and then the individual changes.

So reset your local branch git branch on develop. Then move the code into the separate methods as you did /wo any additonal changes and commit. Then apply changes you propose in 1 or more separate commit per change. Then force force push.

mvorisek:
> after asserting the whole rendered HTML, CS Fixer complains about 2 lines which only have 2 spaces. Is there a simple way to stop CS Fixer complaining about this? The 2 spaces exist in the rendered HTML which is asserted, so I cannot simply remove them to satisfy CS fixer.

Can we fix the source/generator or does the trailing spaces do anything?

PhilippGrashoff:
The spaces dont do anything, no. Fixing this would be another PR IMO. I will check if CS Fixer complains about it when using <<<EOF

src/Form/Layout.php
PhilippGrashoff: understood, will do.

PhilippGrashoff: Done

mvorisek:
I have respushed your last 4 commits as you based the changes on "your develop", not the official develop.

Then thank you, so much easier to review now!

I redone the refactoring on my side and found still one difference - https://github.com/atk4/ui/compare/df52b30d0c2bf00c509ead36613d35d9e359b6f1...238dda14975a239f11082fc7dd2bc708103c8da2 - please explain and commit as extra commit.

PhilippGrashoff:
Hi, sry for messing up with my local dev - will always create a new branch locally for any changes from now.
Regarding your last comment: Don't ask me where this is coming from - I re-added the logic that Sub-Layouts are also rendered in "their" regions if they are defined.

CS is still failing unfortunately due to the spaces in the otherwise empty lines. Of course I could remove them by processing the rendered HTML, but that seems a bit messy to me - changing some return value of a function you want to assert right before asserting.

mvorisek:
I will look into the trailing spaces later, there are present probably because some leading spaces are added somewhere.

mvorisek:
The trailing 2 spaces are comming from https://github.com/atk4/ui/blob/6.0.0/template/form/layout/generic-input.html#L13.

mvorisek:
Great contribution! ❤

PR #2302, opened 2025-10-19
Save Multiline using structured changes

mvorisek:
related with #2297 and #2032

Currently only contained references (defined using `ContainsXxx`) were supported.

The goal is to migrate saving of contained data in favor unified reference saving.

After this PR any reference, stored even in different table/persistence, can be saved.


mvorisek:
linking https://github.com/atk4/data/issues/1285 here with docs for ContainsXxx

PR #2313, opened 2026-04-30
Fix no colspan for empty table

PhilippGrashoff:
Table::tEmpty default Template is:
```
  <tr>
    <td colspan="{$columns}">No records.</td>
  </tr>
```
However, {$columns} was never set, so the <td> always had an empty colspan value. This PR fixes it.


src/Table.php
```diff
@@ -427,6 +427,7 @@ protected function renderView(): void
         // add totals rows or empty message
         if ($this->_renderedRowsCount === 0) {
             if (!$this->jsPaginator || !$this->jsPaginator->getPage()) {
+                $this->tEmpty->set('columns', (string) count($this->columns));
```
mvorisek, line 430: Is there a demo where this can be tested - before and after?

PhilippGrashoff: Sorry for the extremely delayed answer. You can use the tablefilter demo. If you add some filter so no record matches, the empty colspan attribute  is visible. E.g. https://ui.atk4.org/demos/collection/tablefilter.php
Bildschirmfoto vom 2026-06-21 14-50-55

PR #2315, opened 2026-06-22
Fixes to IPP in Grid.php and Grid-demo.

rickyosser:
Added 50 to items list both in the Grid-demo as well as in Grid. 
Fixed so Grid ipp will show selected value if set, otherwise fallback to 50 as it is default.

If $ipp is 0 don't set $this->ipp to the first item as that always sets the viewed selection
to the first item in the list. If we actually track the ipp in the app we want it to set and
show the correct value, as it was before reload or other events, before this change it only set the selected ipp value but showed the first item.
Default ipp in this class is 50, that will be showed when nothing is selected.

If we want to have 10 as default, $ipp should be set to 10 instead of 50.

### DIFF
--- a/demos/collection/grid.php
+++ b/demos/collection/grid.php
@@ -105,4 +105,4 @@ $grid->addModalBulkAction(['Delete selected', 'icon' => 'trash'], '', static fun
 });
 
 // setting ipp with an array will add an ItemPerPageSelector to paginator
-$grid->setIpp([10, 100, 1000]);
+$grid->setIpp([10, 50, 100, 1000]);
--- a/src/Grid.php
+++ b/src/Grid.php
@@ -214,13 +214,18 @@ class Grid extends View
      *
      * @return $this
      */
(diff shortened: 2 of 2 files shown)


mvorisek:
Default of 50 is prefered as better to render more by default for faster UX.

rickyosser:
Ok,

I'm on vacation for a month, change it accordingly or I'll do it when I'm back in August.

Thanks...
Rickard

Den 3 juli 2026 10:57:09 CEST, "Michael Voříšek" ***@***.***> skrev:
>mvorisek left a comment (atk4/ui#2315)
>
>Default of 50 is prefered as better to render more by default for faster UX.
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/atk4/ui/pull/2315#issuecomment-4874522415
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #2317, opened 2026-09-16
Fix DateTime filters

DarkSide666:
Currently modifiers like `today`, `tomorrow`, `yesterday` fail because that method returns `null` instead of proper `DateTime` object.


mvorisek:
Please post a minimal repro code.

DarkSide666:
In Atk4\Ui demos folder create page with
```
declare(strict_types=1);
namespace Atk4\Ui\Demos;

$grid = \Atk4\Ui\Grid::addTo($app);

$model = new Stat($app->db);
$grid->setModel($model);
$grid->addFilterColumn();
```

Try filtering by any date/datetime field by "today", "yesterday" etc.
For example, "Is before Today", "Is after Yesterday".

